PR #4, opened 2024-04-12
Add worker, fill out CLI, + various lints

nbsp:
it compiles! and it seemingly connects to the server just fine.
however, my tests are only at the worker level (e.g. connects, sends heartbeats, exits okay)
i don't know how to test any real-world functionality without the plugins, despite the fact that i'm 100% sure that every sub-component does *not* work as expected (since i never tested any)

request for comments, tests, nits, lints, etc

### DIFF
--- a/agents/src/worker.ts
+++ b/agents/src/worker.ts
@@ -29,7 +29,7 @@ const cpuLoad = (): number =>
   (os
     .cpus()
     .reduce(
-      (acc, x) => acc + x.times.user / Object.values(x.times).reduce((acc, x) => acc + x, 0),
+      (acc, x) => acc + x.times.idle / Object.values(x.times).reduce((acc, x) => acc + x, 0),
       0,
     ) /
     os.cpus().length) *


agents/src/worker.ts
```diff
@@ -0,0 +1,357 @@
...
+
+const MAX_RECONNECT_ATTEMPTS = 10;
+const ASSIGNMENT_TIMEOUT = 15 * 1000;
+const LOAD_INTERVAL = 5 * 1000;
+
+const cpuLoad = (): number =>
+  (os
+    .cpus()
+    .reduce(
+      (acc, x) => acc + x.times.user / Object.values(x.times).reduce((acc, x) => acc + x, 0),
```
davidzhao: I know you are porting over the Python behavior, but I think we would ideally want the CPU utilization instead of load. Load tends to be slower acting, which makes it more difficult to react to load conditions.

Another question.. do you know if this is cgroups compatible?

Here's [the implementation](https://github.com/livekit/protocol/blob/main/utils/hwstats/cpu_linux.go#L98) on the Go side for reference.

nbsp: this should be as easy as replacing `x.times.user` with `x.times.idle` to get the idle CPU percentage.

as for cgroups... i think? Node [supports cgroups v1](https://github.com/nodejs/node/pull/27508), and supported v2 for a minor version before [reverting it](https://github.com/nodejs/node/pull/50036), and then [updating libuv](https://github.com/nodejs/node/pull/48618) so i *think* it's compatible now. 

Bun claims [full compatibility](https://bun.sh/docs/runtime/nodejs-apis#node-os) with `node:os` so it shouldn't introduce any problems there either.

PR #5, opened 2024-04-16
End-to-end integration fixes

### LINKED ISSUE #2: list of as-yet-unfinished components and their dependencies
here's the main dependency chain:

- [x] ~~`apipe`~~ trying to do this synchronously
  - [x] `ipc/job_main` (shim in place, full rewrite of the Python code needed)
    - [x] `ipc/job_process` (same as `ipc/job_main`)
      - [x] `worker`
        - [x] testing, in any capacity
        - [x] CLI

a few other blockers:

- [x] ~~Théo's single-threaded `AsyncPipe`~~
- [x] i don't *think* i need any of `aio`, i'm pretty sure i can replicate it with `EventEmitter`s, but i'm not 100% on that
~~*update: also utilizing `streams::{Readable, Writable, Duplex}`*~~ not doing that anymore as they're not async-friendly
- [ ] ~~i still need to look at {en,de}coding libraries~~ **out of scope for 0.1.0**
  - [ ] ~~`codecs/mp3`~~
- [x] `AbstractEventLoop`s will be the death of me
*update: i have a simple `PromiseQueue` in place, but it's been proving to be unneeded in the synchronous refactor so far*
  - [x] `http_server` and a bunch of other things
- [ ] ~~currently working on `llm/` (probably requires overhaul, i don't think TypeScript allows that kind of function generation, *e.g.* no `type` type)~~ **out of scope for 0.1.0**
  - [ ] ~~`voice_assistant`~~

nbsp:
i'm finally out of integration hell :tada: general changes to the whole control flow, making things that errored not error anymore, work from start to finish, etc.

manually tested and confirmed to be pretty much functionally equivalent to livekit/agents, from the server side at least. the logs also behave in the same way. tested using agents-playground both using LK Cloud and local livekit-server.

notable missing implementation detail is multi-processing and job isolation, which will be done shortly

closes #2

### DIFF
--- a/agents/src/cli.ts
+++ b/agents/src/cli.ts
@@ -22,6 +22,7 @@ const runWorker = async (args: CliArgs) => {
 
   process.on('SIGINT', async () => {
     await worker.close();
+    log.info('worker closed');
     process.exit(130); // SIGINT exit code
   });
 
@@ -29,6 +30,7 @@ const runWorker = async (args: CliArgs) => {
     await worker.run();
   } catch {
     log.fatal('worker failed');
+    process.exit(1);
   }
 };
 
--- a/examples/minimal.ts
+++ b/examples/minimal.ts
@@ -2,16 +2,16 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { runApp } from '../agents/src/cli'
-import { JobContext, JobRequest, WorkerOptions } from '../agents/src'
+import { runApp } from '../agents/src/cli';
+import { JobContext, JobRequest, WorkerOptions } from '../agents/src';
 
 const requestFunc = async (req: JobRequest) => {
-  console.log('received request', req)
+  console.log('received request', req);
   await req.accept(async (_: JobContext) => {
-    console.log('starting voice assistant...')
+    console.log('starting voice assistant...');
 
     // etc
-  })
-}
+  });
+};
 
-runApp(new WorkerOptions({ requestFunc }))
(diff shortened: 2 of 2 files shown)


agents/src/worker.ts
```diff
@@ -119,6 +120,15 @@ class ActiveJob {
   }
 }
 
+class PendingAssignment {
+  promise = new Promise<JobAssignment>((resolve) => {
+    this.resolve = resolve; // oh, JavaScript.
```
davidzhao: what does this do?

nbsp: this is JavaScript's way of triggering a Promise.resolve outside of a Promise. it looks wrong but that's the way you're supposed to do it
later in the code we do `PendingAssignment.promise.await`

the reason `PendingAssignment.resolve(arg: JobAssignment)` isn't empty is because of ESLint type checking, otherwise it complains about unused-variable

davidzhao: I see what you are trying to do.. perhaps we should make [this helper](https://github.com/livekit/client-sdk-js/blob/c2b763eae25daae5640f0cd95e072085d9a6881b/src/room/utils.ts#L345) a shared utils lib somehow

nbsp: hm... we could make a `@livekit/common` helper library, but i don't see real use for it unless there's a lot of shared code between the projects
the cost of importing a new library outweighs the cost of redundancy up to a certain point

nbsp: either way this can be addressed later, i'm going to merge

PR #6, opened 2024-04-19
job isolation

nbsp:
implements job isolation via child processes and simplifies IPC protocol greatly.

the way this is done:
- entry function is serialized to string, then re-evaluated using `new Function('return ' + entry)()(...)`
- instead of trying to pass a `Job` instance down to the child, we keep a copy of the raw packet from LK server and unpack it again there. redundant but it's the only way to do it and it isn't expensive
- both of them are then passed as positional arguments to the child

this works, but it's hacky. i don't know if there's a better way to do any of this though.

### DIFF
--- a/agents/src/ipc/job_main.ts
+++ b/agents/src/ipc/job_main.ts
@@ -53,6 +53,14 @@ if (process.send) {
   const start = () => {
     if (room.isConnected && !closed) {
       process.send!({ type: IPC_MESSAGE.StartJobResponse });
+
+      // JavaScript doesn't let us pass down function arguments to children /as functions/,
+      // because the [Function] type is a class, and thus is incompatible with the Web Workers
+      // structured clone algorithm, which only allows simple types and object literals.
+      // attempting to pass a function here through IPC will lead to a runtime DataCloneError.
+      //
+      // further info:
+      // https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
       new Function('return ' + process.argv[3])()(new JobContext(closeEvent, args.job!, room));
     }
   };


agents/src/ipc/job_main.ts
```diff
@@ -2,40 +2,58 @@
...
 
-      task = args.acceptData.entry;
-      context = new JobContext(event, request.job, room);
+  const conn = room.connect(args.url || process.argv[4], args.token);
 
-      task(context);
+  const start = () => {
+    if (room.isConnected && !closed) {
+      process.send!({ type: IPC_MESSAGE.StartJobResponse });
+      new Function('return ' + process.argv[3])()(new JobContext(closeEvent, args.job!, room));
```
davidzhao, line 64: I don't think we should use a function coming from the main process. IMO it should just launch a local function that drives the job itself.

the only information the child process needs is the specifics about the job.

agents/src/ipc/job_main.ts
```diff
@@ -2,40 +2,58 @@
...
+      });
+    }
+  });
+
+  // don't do anything on C-c
+  process.on('SIGINT', () => {});
 
-      task = args.acceptData.entry;
-      context = new JobContext(event, request.job, room);
+  const conn = room.connect(args.url || process.argv[4], args.token);
```
davidzhao, line 51: the parent process should not connect to the room, this should take place in the child process.

nbsp: this does take place in the child process. everything under `if (process.send)` is executed by the child; the function before is called by the parent

agents/src/ipc/job_main.ts
nbsp: this function that we're running here takes a `JobContext` that itself takes a `Job` and a `Room`, both of them are in the child
passing them up to the parent for handling is not only impractical with node's clone rules, it's also bad for security and for scalability

as a rule, the user-defined function passed into `runApp` is the weakest link in the program, since we didn't write it: it could be insecure, or buggy, or have a memory leak, or just be inefficient. we should not trust it enough to run it on the main process in my opinion

agents/src/ipc/job_main.ts
davidzhao: sorry I got confused by the diff.

agents/src/ipc/job_main.ts
davidzhao: you are right, I don't think either them should be passed back to parent. may I ask what is the reason to pass a serialized function in argv[3]?

nbsp: take [this](https://github.com/livekit/agents/blob/4ea33bc7c5087325eea5308c6ba0a122d5419bd1/livekit-agents/livekit/agents/ipc/job_main.py#L93) line from the Python Agents library:

```python
            usertask = asyncio.create_task(args.accept_data.entry(ctx))
```

- this line of code runs in a child process
- `args.accept_data.entry` is a function, of type `Callable[[JobContext], Coroutine]` (an async function that takes a `JobContext` class instance)
- we get it from `cli.run_app` right [here](https://github.com/livekit/agents/blob/4ea33bc7c5087325eea5308c6ba0a122d5419bd1/examples/minimal_worker.py#L17-L23), outside the library

now in JavaScript we can't directly pass `args.accept_data` and call `entry` since that's a function and it DataCloneErrors on runtime. instead what we do is serialize it, pass it as a string, then reeval it and execute it as the child.

it is never executed in the parent, and is treated as an opaque blob in transmission.

agents/src/ipc/job_main.ts
```diff
@@ -53,6 +53,14 @@ if (process.send) {
...
     if (room.isConnected && !closed) {
       process.send!({ type: IPC_MESSAGE.StartJobResponse });
+
+      // JavaScript doesn't let us pass down function arguments to children /as functions/,
+      // because the [Function] type is a class, and thus is incompatible with the Web Workers
+      // structured clone algorithm, which only allows simple types and object literals.
+      // attempting to pass a function here through IPC will lead to a runtime DataCloneError.
+      //
+      // further info:
+      // https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
```
nbsp, line 63: @davidzhao does this look clearer now?

davidzhao: thanks!

PR #8, opened 2024-05-17
Monorepo setup with task runners, testing lib and API consistency tools

lukasIO:
this adds a couple of things:

- [x]  turbo as a monorepo task runner
- [x] a root package.json for easy script invocation across the monorepo
- [x] vitest as a testrunner
- [x] rushstack api-extractor to enable CI checks for API consistency (once we need it, too early right now in pre-release) 
- [x] eslint+prettier combined setup taken from the components-js repo

TODO:
- [ ] add changesets CLI setup (would want to tackle this in a separate PR as this one is already a monster)


.github/workflows/build.yml
```diff
@@ -3,52 +3,33 @@
...
-          cd agents
-          pnpm build
-          cd ../plugins/elevenlabs
-          pnpm build
+      - name: Lint
+        run: pnpm lint
+      - name: Prettier
+        run: pnpm format:check
+      - name: Build
+        run: pnpm build
```
nbsp, line 35: should keep fsfe/reuse-action here, if it were to fail the licensing is broken

.gitignore
```diff
@@ -173,3 +173,10 @@ dist
 
 # Finder (MacOS) folder config
 .DS_Store
+
```
nbsp, line 176: nit: extra line?

agents/src/index.ts
nbsp, line 1: is this needed? it's more verbose but i don't know if it helps readability.
either way the REUSE header should be on top

examples/src/minimal.ts
```diff
@@ -1,13 +1,12 @@
...
 // SPDX-License-Identifier: Apache-2.0
-
-import { cli, JobContext, JobRequest, WorkerOptions } from '@livekit/agents';
+import { type JobContext, type JobRequest, WorkerOptions, cli } from '@livekit/agents';
 import { fileURLToPath } from 'url';
 
 // your entry file *has* to include an exported function [entry].
 // this file will be imported from inside the library, and this function
-// will be called.
+// will be called.#
```
nbsp: typo

examples/package.json
```diff
@@ -4,17 +4,16 @@
   "type": "module",
   "scripts": {
     "build": "tsc",
-    "lint": "eslint src"
+    "lint": "eslint -f unix \"src/**/*.ts\""
   },
   "devDependencies": {
-    "@types/bun": "latest"
   },
```
nbsp: oh also delete this (devdeps is empty)

.eslintrc
nbsp, line 1: .eslintrc is [deprecated](https://eslint.org/docs/latest/use/configure/configuration-files-deprecated). best to keep it around here for now but it should be revised on all repos sometime

.github/workflows/build.yml
lukasIO: Not sure I follow, the reuse-action runs before `build`, just like it did before?

agents/src/index.ts
lukasIO: this is just a result of the prettier + eslint config applied and running `pnpm format:write`. REUSE header is still on top, I think?

.github/workflows/build.yml
nbsp: oh, oops, yeah missed that. saw a lot of red :sweat_smile:

agents/src/index.ts
nbsp: it got a bit overzealous and put it on top of REUSE

.eslintrc
lukasIO: oh, I didn't know! thanks for pointing that out. Should I try to revert to the format you used? 
I just tried to make it simple by being able to copy over the definitions from another of our repos. 
Will go back to a `.js` file

agents/src/index.ts
lukasIO: hah, now I see it!

.eslintrc
lukasIO: actually, with this being already defined [here](https://github.com/livekit/components-js/tree/c1c3d074ba026bf8260e6975015e4db93ab494a4/tooling/eslint-config-custom) as its own lint package, maybe it makes most sense to publish the custom config on npm and import it here for ease of use across repos?

nbsp: the new configs are a new format from the ground up, i think once we port it over it should (a) be its own PR and (b) happen in all repos at once
consistency is more important i think

lukasIO: makes sense!

examples/package.json
```diff
@@ -4,17 +4,14 @@
...
-    "@types/bun": "latest"
+    "lint": "eslint -f unix \"src/**/*.ts\""
   },
   "peerDependencies": {
     "typescript": "^5.0.0"
   },
   "dependencies": {
     "@livekit/agents": "workspace:*",
-    "@livekit/elevenlabs": "workspace:*",
+    "@livekit/agents-plugin-elevenlabs": "workspace:*",
```
lukasIO, line 14: just for context: the idea behind renaming this was to have a standard way of naming agent plugins in the future, following an example of rollup plugin conventions. 
There officially maintained plugins are named `@rollup/plugin-official-plugin` and community plugins should be named `rollup-plugin-my-plugin`

PR #9, opened 2024-05-18
Simplify entry file heuristics

nbsp:
still feels unidiomatic and handwavey magic, but it's better.

Suggested-by: lukasIO <mail@lukasseiler.de>

### DIFF
--- a/agents/src/ipc/job_main.ts
+++ b/agents/src/ipc/job_main.ts
@@ -10,15 +10,15 @@ import { log } from '../log.js';
 import { IPC_MESSAGE, type JobMainArgs, type Message, type Ping } from './protocol.js';
 
 export const runJob = (args: JobMainArgs): ChildProcess => {
-  return fork(__filename, [args.raw, args.entry, args.fallbackURL]);
+  return fork(import.meta.filename, [args.raw, args.entry, args.fallbackURL]);
 };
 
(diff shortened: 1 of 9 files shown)


agents/src/generator.ts
```diff
@@ -0,0 +1,22 @@
+// SPDX-FileCopyrightText: 2024 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import type { JobContext } from './job_context.js';
+
+type entryFunction = (job: JobContext) => Promise<void>;
```
davidzhao, line 6: should it return `Promise<void | () => void>` to reflect that it could return a function?

nbsp: does it return anything? all of the examples on Python's agents return void, and regardless we don't do anything with the potential return value of this function, we just run it.
there's no reason to return anything because there's nowhere to return it to

if we really want to allow returning things for whatever reason we should replace void with `unknown` to prevent type errors

lukasIO: the confusion is my fault, I thought it'd be nice to provide some sort of resource cleanup functionality, but I thought about it just in terms of API and not in terms of what's already implemented... I guess that we should just remove the comment that mentions the return cleanup function and leave the return type as `Promise<void>`.

PR #11, opened 2024-05-21
**/package.json: typescript -> devDependencies

lukasIO:


pnpm-workspace.yaml
```diff
@@ -1,4 +1,4 @@
 packages:
-  - "agents"
-  - "plugins/*"
-  - "examples"
+  - 'agents'
```
lukasIO: looks like this might be a change coming from the new prettier formatting. Though we're not currently checking for the whole repo with the `format:check`, but only for the src folders

PR #24, opened 2024-09-05
Fix load calculation, report worker status

bcherry:
I found that node agents had some bugs around load calculation. In testing, the agent only actually connected if you booted your room within 5s of starting the agent, because it was getting dropped from rotation as soon as it made its first load report. maybe these are holdovers from an older agents framework spec, not sure.

- load calculation was inverted, it was actually an idle calculation
- raw load was being sent to the server, instead of an availability determination
- load was in integer percent (e.g. 20) instead of fractional percent (e.g. 0.2)

this PR fixes the default function, adds the threshold param, and reports worker status. it is identicial to the python implementation now , except that draining isn't yet implemented.

i also bumped the version to 0.7 which i missed in the last PR

### DIFF
--- a/.changeset/few-llamas-listen.md
+++ b/.changeset/few-llamas-listen.md
@@ -1,0 +1,9 @@
+---
+"@livekit/agents": minor
+"@livekit/agents-plugin-elevenlabs": minor
+"livekit-agents-examples": minor
+---
+
+bump underlying dependencies
+fix load calculation
+report worker status


package.json
```diff
@@ -39,5 +39,6 @@
     "typedoc": "^0.25.13",
     "typescript": "^5.4.5",
     "vitest": "^1.6.0"
-  }
+  },
+  "packageManager": "pnpm@9.6.0+sha512.38dc6fba8dba35b39340b9700112c2fe1e12f10b17134715a4aa98ccf7bb035e76fd981cf0bb384dfa98f8d6af5481c2bef2f4266a24bfa20c34eb7147ce0b5e"
```
bcherry: @nbsp is it a problem that my pnpm wants to add this?

nbsp: yeah we don't have a package manager version pinned (using different ones fucks with pnpm-lock and creates huge diffs). let's use the same one on node-sdks:
    
```json
  "packageManager": "pnpm@9.2.0"
```

and delete the `version` option from the CI build too

and then run npm install again to rewrite the pnpm-lock one last time

agents/src/worker.ts
```diff
@@ -305,15 +311,33 @@ export class Worker {
       }),
     );
 
+    var currentStatus = WorkerStatus.WS_AVAILABLE;
```
nbsp: `var` -> `let` (ECMAScript 6)

agents/src/worker.ts
```diff
@@ -305,15 +311,33 @@ export class Worker {
...
     const loadMonitor = setInterval(() => {
       if (closingWS) clearInterval(loadMonitor);
+
+      const oldStatus = currentStatus;
+      const currentLoad = this.#opts.loadFunc();
+      const isFull = currentLoad >= this.#opts.loadThreshold;
+      const currentlyAvailable = !isFull
+      currentStatus = currentlyAvailable ? WorkerStatus.WS_AVAILABLE : WorkerStatus.WS_FULL;
```
nbsp, line 323: switch the two around and save the `const currentlyAvailable` line?

examples/src/minimal.ts
nbsp, line 1: so this *works*, but i wouldn't sweat the details on the minimal examples because the recommended workflow for agents completely changed in 0.8 and we'd need to rewrite all of this anyway

instead of `WorkerOptions({ requestFunc })` ideally we'd set `WorkerOptions({ entryFile: import.meta.filename })` and the rest would be unchanged. let's talk about it later, for now this is fine yeah

agents/src/worker.ts
bcherry: yeah it's a little awkward but I did it this way to mirror the python code where currentlyAvailable also factors in `isDraining`, which will come here soon

agents/src/worker.ts
bcherry: ooh thanks I'm thinking in swift where `let` means `const`

examples/src/minimal.ts
bcherry: yeah that makes sense - thanks. this is just the code from the wiki guide actually so i'll be sure to update it there too

package.json
bcherry: thank you!

PR #25, opened 2024-09-06
overhaul job processing

nbsp:
- [x] condense `job_request`, `job_context` and `ipc/job_process` into one file, and update the implementations to be on par with main agents
- [x] `proc_pool` implementation
  - [x] `JobExecutor` base
  - [x] `ProcJobExecutor`
  - [ ] ~~`ThreadJobExecutor`~~
- [x] make `job_main` behave
  - [x] `protocol` updates
  - [x] figuring out method for function passthrough. [slack discussion](https://live-kit.slack.com/archives/C05P6DRC6MV/p1725860754858159)
- [x] make `worker` behave (finished, but depends on `proc_pool` for testing)
- [x] go over locks again (i skipped all of the implementations so far)

does not build, obviously.

### DIFF
--- a/agents/src/cli.ts
+++ b/agents/src/cli.ts
@@ -1,7 +1,6 @@
 // SPDX-FileCopyrightText: 2024 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import { ServerInfo } from '@livekit/protocol';
 import { Command, Option } from 'commander';
 import type { EventEmitter } from 'events';
 import { initializeLogger, log } from './log.js';
--- a/agents/src/ipc/message.ts
+++ b/agents/src/ipc/message.ts
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import type { RunningJobInfo } from '../job.js';
-import { LoggerOptions } from '../log.js';
+import type { LoggerOptions } from '../log.js';
 
 export type IPCMessage =
   | {
--- a/agents/src/log.ts
+++ b/agents/src/log.ts
@@ -1,7 +1,8 @@
 // SPDX-FileCopyrightText: 2024 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import { Logger, pino } from 'pino';
+import type { Logger } from 'pino';
+import { pino } from 'pino';
 
 export type LoggerOptions = {
   pretty: boolean;
--- a/agents/src/tts/stream_adapter.ts
+++ b/agents/src/tts/stream_adapter.ts
@@ -2,7 +2,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import type { SentenceStream, SentenceTokenizer } from '../tokenize.js';
-import { ChunkedStream, SynthesisEvent, SynthesisEventType, SynthesizeStream, TTS } from './tts.js';
+import type { ChunkedStream } from './tts.js';
+import { SynthesisEvent, SynthesisEventType, SynthesizeStream, TTS } from './tts.js';
 
 export class StreamAdapterWrapper extends SynthesizeStream {
   #closed: boolean;


agents/src/ipc/proc_pool.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+  procDequeue() {
+    if (this.procQueue.length > 0) {
+      const next = this.procQueue.shift();
+      if (next) {
+        next();
+      }
+    }
+  }
```
nbsp: these are supposed to be queues and semaphores. kind of long, i may have messed something up here. if i can get another pair of eyes to compare this with the python impl 🙇‍♀️

theomonnom: @lukasIO already made a nice Mutex implementation [here](https://github.com/livekit/client-sdk-js/blob/bf7fb5d1c5df264fd1ce5adfc951bbec73fbfd77/src/room/utils.ts#L455C14-L455C19)

As well as an AsyncQueue [here](https://github.com/livekit/client-sdk-js/blob/main/src/utils/AsyncQueue.ts) (may not be useful for your current impl)

theomonnom: Python is using a semaphore to limit the number of concurrent process initializations

agents/src/ipc/job_executor.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export abstract class JobExecutor {
+  PING_INTERVAL = 2.5 * 1000;
+  PING_TIMEOUT = 90 * 1000;
+  HIGH_PING_THRESHOLD = 0.5 * 1000;
+  NO_MESSAGE_TIMEOUT = 15 * 1000;
+
+  abstract get started(): boolean;
+  abstract get startArguments(): unknown;
+  abstract set startArguments(value: unknown);
```
theomonnom: Not exposed today, let's ignore

agents/src/worker.ts
```diff
@@ -41,29 +46,37 @@ export class WorkerPermissions {
...
     this.hidden = hidden;
   }
 }
 
 export class WorkerOptions {
-  requestFunc: (arg: JobRequest) => Promise<void>;
+  agent: string;
+  requestFunc: (job: JobRequest) => Promise<void>;
   loadFunc: () => number;
+  jobExecutorType: JobExecutorType;
```
theomonnom: let's not expose today

agents/src/ipc/job_main.ts
```diff
@@ -1,85 +1,138 @@
...
-      shuttingDown = true;
-      closed = true;
-      closeEvent.emit('close');
-    } else if (msg.type === IPC_MESSAGE.Ping) {
-      process.send!({
-        type: IPC_MESSAGE.Pong,
-        lastTimestamp: (msg as Ping).timestamp,
-        timestamp: Date.now(),
-      });
+  room.on('disconnected', () => {
```
bcherry: can we use `RoomEvent.Disconnected`?

agents/src/worker.ts
```diff
@@ -41,29 +45,36 @@ export class WorkerPermissions {
...
 
 export class WorkerOptions {
-  requestFunc: (arg: JobRequest) => Promise<void>;
+  agent: string;
+  requestFunc: (job: JobRequest) => Promise<void>;
   loadFunc: () => number;
   loadThreshold: number;
-  namespace: string;
+  numIdleProcesses: number;
+  shutdownProcessTimeout: number;
```
theomonnom, line 80: Can you use the same pattern as we do in client-sdk-js https://github.com/livekit/client-sdk-js/blob/260a5e8cad752d824456f911ef6b13319d497a6b/src/room/defaults.ts#L43

having an interface defining the fields with their type and another constant will all the default values.

You can then use the defaults like this: https://github.com/livekit/client-sdk-js/blob/260a5e8cad752d824456f911ef6b13319d497a6b/src/room/Room.ts#L180

Pretty neat pattern IMO

agents/src/worker.ts
```diff
@@ -120,74 +144,77 @@ export class WorkerOptions {
...
+  #closed = true;
+  #draining = false;
+  #connecting = false;
+  #tasks: Promise<void>[] = [];
+  #pending: { [id: string]: PendingAssignment } = {};
+
+  #close = () => {};
+  #closePromise = new Promise<void>((resolve) => {
+    this.#close = resolve;
+  });
```
theomonnom: I see that we store `promise` multiple times in the codebase. Would be cool to have an utility for this. So we don't need to have 2 fields each time. 
E.g in the `ProcJobExecutor` there is also the InitPromise and the JoinPromise (even 3 fields with the error)

agents/src/ipc/job_executor.ts
```diff
@@ -0,0 +1,26 @@
...
+  agent: string;
+  initializeTimeout: number;
+  closeTimeout: number;
+}
+
+export abstract class JobExecutor {
+  PING_INTERVAL = 2.5 * 1000;
+  PING_TIMEOUT = 90 * 1000;
+  HIGH_PING_THRESHOLD = 0.5 * 1000;
+  NO_MESSAGE_TIMEOUT = 15 * 1000;
```
theomonnom: This isn't yet implemented, I'm going to add a ticket for it. Feel free to ignore on this PR

agents/src/worker.ts
nbsp: this doesn't work as well, because we have a mix of optional and non-optional attributes here. i think the way it works now is more than fine (and it also lets you do `WorkerOptions({ ... })` instead of just `{ ... }`)

theomonnom: Ah right, yes I think it's ok

PR #27, opened 2024-09-11
Realtime API voice assistant - initial spike

bcherry:
This is an initial spike to get voice-in/voice-out working with the model. It works!

### DIFF
--- a/examples/src/minimal_assistant.ts
+++ b/examples/src/minimal_assistant.ts
@@ -8,10 +8,10 @@ export default defineAgent({
   entry: async (ctx: JobContext) => {
     await ctx.connect();
 
+    console.log('starting assistant example agent');
+
     // FIXME: for some reason the remoteParticipants are not being populated at connection time nor calling onParticipantConnected
     setTimeout(() => {
-      console.log('starting assistant example agent');
-
       const assistant = new VoiceAssistant({
         ...defaultInferenceConfig,
         system_message: 'You talk unprompted.',


agents/package.json
```diff
@@ -18,7 +18,7 @@
     "typescript": "^5.0.0"
   },
   "dependencies": {
-    "@livekit/rtc-node": "^0.7.0",
+    "@livekit/rtc-node": "file:../../node-sdks/packages/livekit-rtc",
```
bcherry: currently depends on https://github.com/livekit/node-sdks/pull/264, we need to merge and update that before this one is ready

agents/src/audio.ts
```diff
@@ -0,0 +1,58 @@
+import { AudioFrame } from '@livekit/rtc-node';
```
nbsp, line 4:
```suggestion
// SPDX-FileCopyrightText: 2024 LiveKit, Inc.
//
// SPDX-License-Identifier: Apache-2.0
import { AudioFrame } from '@livekit/rtc-node';
```

plugins/openai/src/voice_assistant/proto.ts
```diff
@@ -0,0 +1,63 @@
+export enum Voice {
```
nbsp, line 4:
```suggestion
// SPDX-FileCopyrightText: 2024 LiveKit, Inc.
//
// SPDX-License-Identifier: Apache-2.0
export enum Voice {
```

bcherry: thanks - is there a way to automate this?

pnpm-lock.yaml
nbsp, line 1: this file shouldn't exist (probably a remnant of the manual linking to rtc-node)

plugins/openai/src/voice_assistant/index.ts
```diff
@@ -0,0 +1,280 @@
...
+        ...command,
+        ...truncatedDataPartial,
+      });
+    }
+    this.ws.send(JSON.stringify(command));
+  }
+
+  private handleServerEvent(event: Record<string, unknown>): void {
+    const truncatedDataPartial = event['data']
+      ? { data: (event['data'] as string).slice(0, 30) + '…' }
```
nbsp, line 155: i don't remember but there might have been a discussion about what the trunc character should be and whether it should be configurable. is this ellipsis the same as in python agents or do we use three dots there?

plugins/openai/src/voice_assistant/proto.ts
nbsp, line 1: let's make these enums `UPPER_SNAKE_CASE`? consistent with most other enums in the code

pnpm-lock.yaml
lukasIO: I think the file itself is needed?

nbsp: ah oops, thought this was a file on `agents/`

plugins/openai/src/voice_assistant/index.ts
bcherry: i don't think we have the same logging between them - we should bring them together but i was just adding it ad-hoc for now

PR #30, opened 2024-09-12
cli: add connect command

nbsp:
driveby fix: logs were broken on child_process.

~~not ready, breaks if `participant-identity` is set. figuring this out now~~


agents/src/log.ts
```diff
@@ -12,7 +18,8 @@ export const log = () => {
 };
 export default log;
 
-export const setLog = ({ pretty, level }: { pretty: boolean; level: string }) => {
+export const setLog = ({ pretty, level }: LoggerOptions) => {
```
theomonnom: Maybe this could be renamed to initializeLogger since it'll only be used once when we start the worker

PR #32, opened 2024-09-12
add pid/jobid to child and fix Future

nbsp:


agents/src/utils.ts
```diff
@@ -106,14 +106,15 @@ export class Queue<T> {
...
-      this.resolve = resolve;
-      this.reject = reject;
-    });
+export class Future {
+  #await = new Promise<void>((resolve, reject: (_: Error) => void) => {
+    this.resolve = resolve;
+    this.reject = reject;
+  });
+
+  get await() {
```
theomonnom, line 115: nit: can we avoid the double await, we could rename this to Promise 
not strong opinion

nbsp: that was actually the previous implementation and it didn't work for some reason when i attempted to `extend Promise`

PR #37, opened 2024-09-12
cli: drain on exit, and add force-exit option

nbsp:
confirmed that the force-exit works, but draining is extremely fast so i couldn't make it work unless i added an artificial delay

### DIFF
--- a/agents/src/cli.ts
+++ b/agents/src/cli.ts
@@ -33,7 +33,9 @@ const runWorker = async (args: CliArgs) => {
       log().info('worker closed forcefully');
       process.exit(130); // SIGINT exit code
     });
-    await worker.drain();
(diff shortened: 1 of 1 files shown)


agents/src/cli.ts
```diff
@@ -27,7 +27,13 @@ const runWorker = async (args: CliArgs) => {
...
   }
 
-  process.on('SIGINT', async () => {
+  process.once('SIGINT', async () => {
+    // allow C-c C-c for force interrupt
+    process.once('SIGINT', () => {
+      log().info('worker closed forcefully');
+      process.exit(130); // SIGINT exit code
+    });
+    await worker.drain();
```
theomonnom: we shouldn't drain in devmode

PR #40, opened 2024-09-13
Send agent transcript progressively and handle interruptions

bcherry:
This implements a simplified version of TTSForwarder that doesn't do tokenization for now. Instead it uses a fixed estimate of characters-per-second (tuning found that 16 is a decent value) and then does a simple computation once the final audio and text lengths are known.  Since it's so much more basic than what's in Python, I kept it in the OpenAI plugin and didn't give it the same name.

Interruption is done same as in python, no issues there.

There's room for improvement - this works great on small texts but on long ones it can get pretty far out of sync. A proper hyphenator with punctuation awareness would probably help a lot.

### DIFF
--- a/plugins/openai/src/voice_assistant/index.ts
+++ b/plugins/openai/src/voice_assistant/index.ts
@@ -12,7 +12,6 @@ import type {
   Room,
 } from '@livekit/rtc-node';
 import {
-  AudioFrame,
   AudioSource,
   AudioStream,
   AudioStreamEvent,


plugins/openai/src/voice_assistant/proto.ts
```diff
@@ -114,25 +118,20 @@ export type ClientEvent =
...
+            | {
                 type: 'audio';
-                audio: 'string'; // base64 encoded buffer
-              };
-            }
-        ))
+                audio: string; // base64 encoded buffer
+              },
+          ];
+        }
```
nbsp: according to spec, The client cannot add audio to `system` or `assistant` messages.

PR #41, opened 2024-09-13
llm,openai: implement function calling

nbsp:
the openai.start() constructor is a bit janky with the new union, but it's the cleanest way to do it.

### DIFF
--- a/plugins/openai/src/voice_assistant/index.ts
+++ b/plugins/openai/src/voice_assistant/index.ts
@@ -50,20 +50,20 @@ export class VoiceAssistant {
   subscribedTrack: RemoteAudioTrack | null = null;
   readMicroTask: { promise: Promise<void>; cancel: () => void } | null = null;
 
-  constructor(
-    inferenceConfig: proto.InferenceConfig & {
-      functions?: llm.FunctionContext;
-    } = defaultInferenceConfig,
-    apiKey: string = process.env.OPENAI_API_KEY || '',
-  ) {
+  constructor({
+    inferenceConfig = defaultInferenceConfig,
+    functions = {},
+    apiKey = process.env.OPENAI_API_KEY || '',
+  }: {
+    inferenceConfig?: proto.InferenceConfig;
+    functions?: llm.FunctionContext;
+    apiKey?: string;
+  }) {
     if (!apiKey) {
       throw new Error('OpenAI API key is required, whether as an argument or as $OPENAI_API_KEY');
     }
 
-    const functions = inferenceConfig.functions || {};
-    delete inferenceConfig.functions;
     inferenceConfig.tools = tools(functions);
-
     this.options = {
       apiKey,
       inferenceConfig,
@@ -276,7 +276,7 @@ export class VoiceAssistant {
         });
         break;
       }
-      default: {
+      case 'message': {
         const itemId = event.id as string;
         if (itemId && this.pendingMessages.has(itemId)) {
           const text = this.pendingMessages.get(itemId) || '';
--- a/examples/src/minimal_assistant.ts
+++ b/examples/src/minimal_assistant.ts
@@ -12,8 +12,10 @@ export default defineAgent({
     console.log('starting assistant example agent');
 
     const assistant = new VoiceAssistant({
-      ...defaultInferenceConfig,
-      system_message: 'You talk unprompted.',
+      inferenceConfig: {
+        ...defaultInferenceConfig,
+        system_message: 'You talk unprompted.',
+      },
       functions: {
         weather: {
           description: 'Get the weather in a location',


plugins/openai/src/voice_assistant/index.ts
```diff
@@ -249,17 +259,38 @@ export class VoiceAssistant {
...
+            tool_call_id: event.tool_call_id as string,
+            content: JSON.stringify(content),
+          });
+          this.sendClientCommand({
+            event: proto.ClientEvent.CLIENT_TURN_FINISHED,
+          });
+        });
+        break;
+      }
+      default: {
```
bcherry: use an explicit item type here and leave default for a warning about unhandled types

plugins/openai/src/voice_assistant/index.ts
```diff
@@ -398,3 +429,13 @@ export class VoiceAssistant {
     });
   }
 }
+
+const tools = (ctx: llm.FunctionContext): proto.Tool[] =>
```
bcherry, line 433: consider exporting this method, or one like it, in the plugin's util package and instead of accepting a `functions` on VoiceAssistant parameter instead just have the caller convert their zod blob to `Tool[]` first using this function and include it on inference config.

nbsp: you have to pass the zod schema into the VoiceAssistant anyway, i think it should handle the conversion by itself. the interface could be improved, though.
does this work?

```ts
const assistant = new VoiceAssistant(
  inferenceConfig: {
    ...defaultInferenceConfig,
    system_message: 'You talk unprompted.',
  },
  functions: {
    weather: {
    description: 'Get the weather in a location',
    parameters: z.object({
      location: z.string().describe('The location to get the weather for'),
    }),
    execute: async ({ location }) => ({
      location,
      temperature: await fetch(`https://wttr.in/${location}?format=%C+%t`).then((data) =>
        data.text(),
      ),
      source: 'wttr.in',
    }),
  },
});

```

plugins/openai/src/voice_assistant/index.ts
nbsp: according to the spec there isn't any other type that goes here so we don't need default

plugins/openai/src/voice_assistant/index.ts
nbsp: we could even apply `...defaultInferenceConfig` at the constructor, so `inferenceConfig` only has to include the changes to the default

plugins/openai/src/voice_assistant/index.ts
```diff
@@ -249,17 +259,38 @@ export class VoiceAssistant {
...
-      const trackSid = this.getLocalTrackSid();
-      if (participantIdentity && trackSid) {
-        this.publishTranscription(participantIdentity, trackSid, text, true, itemId);
-      } else {
-        this.logger.error('Participant or track not set');
+    switch (event.type) {
+      case 'tool_call': {
+        const name = event.name as string;
+        const args = event.arguments as object;
+        this.options.functions[name].execute(args).then((content) => {
```
theomonnom, line 266: Is the event.arguments validated by Zod when using inferParameters?

nbsp: yes:

```ts
export type inferParameters<P extends z.ZodTypeAny> = z.infer<P>;

export interface CallableFunction<P extends z.ZodTypeAny = any, R = any> {
  description: string;
  parameters: P;
  execute: (args: inferParameters<P>) => PromiseLike<R>;
}
```

args has to be of the same type as P, and it's being explicitly verified against the constraints of the zod type

plugins/openai/src/voice_assistant/index.ts
bcherry: no other type _today_

plugins/openai/src/voice_assistant/index.ts
bcherry: i think its odd that you can pass tools in one spot but they're ignored, and this makes zod an explicit dependency to initialize a voice assistant. if you made a utility method imported separately then voice assistant could accept a "plain" object and if you want to use zod to make it, you can.

bcherry: (not a blocker to merge - we've got time to get internal feedback and iterate anyways)

nbsp: this only makes zod an explicit dependency if you use function calling. there's no reason for you to pass functions to openai `tools` without also passing them to `functions`, because otherwise you can't execute it

a thing we could support down the line is JSON schema instead of zod? which would cut the zod dependency, but the API would stay the same: you have to pass VoiceAssistant the function *somewhere*

PR #42, opened 2024-09-14
explicitly define types for {Server,Client}Event

nbsp:
this is forwards-compatible as long as the server is backwards-compatible; `ClientEvent` should always be accepted on their side unless there's a breaking API change, and for `ServerEvent` TypeScript doesn't do any runtime checking, so as long as the values we expect are still there it'll run great.

curiously, this is slightly inconsistent with the API as defined in the PDF; item_added.message_id is actually item_added.id in the impl.

### DIFF
--- a/plugins/openai/src/voice_assistant/proto.ts
+++ b/plugins/openai/src/voice_assistant/proto.ts
@@ -221,3 +221,10 @@ export type InferenceConfig = {
   tools: Tool[];
   tool_choice: ToolChoice;
 };
+
+export enum State {
+  INITIALIZING = 'initializing',
+  LISTENING = 'listening',
+  THINKING = 'thinking',
+  SPEAKING = 'speaking',
+}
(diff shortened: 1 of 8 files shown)


plugins/openai/src/voice_assistant/proto.ts
```diff
@@ -31,7 +31,66 @@ export enum ServerEvent {
...
+      | {
+          type: 'message';
+        }
+      | { type: 'tool_call'; tool_call_id: string; name: string }
+    ))
+  | {
+      event: ServerEventType.ADD_CONTENT;
+      item_id: string;
+      type: 'text' | 'audio' | 'tool_call_arguments';
+      data: string; // text or base64 audio or JSON stringified object
```
bcherry, line 60: i wonder if you could break these out into three different subtypes, too. if type == text then data always == string, etc.

bcherry: but i guess it's always a string anyways here so nvm

PR #44, opened 2024-09-14
add voice_assistant.state support

nbsp:
tested, compatible with the new components.

there's only one issue and it's that it goes back to `LISTENING` the moment the turn is marked as done, which could be way before the agent actually stops talking. i couldn't find a good heuristic for "done publishing data after turn marked as done".

### DIFF
--- a/plugins/openai/src/voice_assistant/index.ts
+++ b/plugins/openai/src/voice_assistant/index.ts
@@ -81,6 +81,10 @@ export class VoiceAssistant {
   private pendingMessages: Map<string, string> = new Map();
   private logger = log();
 
+  private speechLeftMs: number | null = null;
+  private speechStarted: number = 0;
+  private speechTimeout: ReturnType<typeof setTimeout> | undefined = undefined;
+
   start(room: Room, participant: RemoteParticipant | string | null = null): Promise<void> {
     return new Promise(async (resolve, reject) => {
       if (this.ws !== null) {
@@ -207,7 +211,6 @@ export class VoiceAssistant {
         this.handleItemAdded(event);
         break;
       case proto.ServerEvent.TURN_FINISHED:
-        this.setState(proto.State.LISTENING);
         break;
       case proto.ServerEvent.VAD_SPEECH_STARTED:
         this.handleVadSpeechStarted(event);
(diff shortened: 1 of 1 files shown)


agents/src/worker.ts
```diff
@@ -456,6 +456,7 @@ export class Worker {
               canPublish: this.#opts.permissions.canPublish,
               canSubscribe: this.#opts.permissions.canSubscribe,
               canPublishData: this.#opts.permissions.canPublishData,
+              canUpdateMetadata: this.#opts.permissions.canUpdateMetadata,
```
davidzhao, line 459: just to check, does this default to true?

nbsp: yes, `WorkerPermissions` defaulted this to true, but it never sent it to the server

plugins/openai/src/voice_assistant/index.ts
```diff
@@ -220,6 +237,18 @@ export class VoiceAssistant {
...
+          this.speechLeftMs -= Date.now() - this.speechStarted;
+          this.speechLeftMs += (serverFrame.data.length / proto.SAMPLE_RATE) * 1000;
+        } else {
+          this.speechLeftMs = (serverFrame.data.length / proto.SAMPLE_RATE) * 1000;
+        }
+        this.speechStarted = Date.now();
+        this.speechTimeout = setTimeout(() => {
+          this.setState(proto.State.LISTENING);
+        }, this.speechLeftMs);
+
```
nbsp, line 251: this is the method i came up with: keep track of how many samples are being added and set a timeout to trigger once they're all done. it works perfectly but it might be a tad hacky

lukasIO: I think the same issue also exists in python (ran into this as well when testing the react components). 
If this turns out to be a viable solution, we should consider adding it to python, too

plugins/openai/src/voice_assistant/index.ts
```diff
@@ -264,6 +299,7 @@ export class VoiceAssistant {
         const name = event.name as string;
         const args = event.arguments as object;
         this.options.functions[name].execute(args).then((content) => {
+          this.thinking = false;
```
bcherry, line 302: this might run into issues if multiple tool calls are ongoing at once. maybe better to keep a registry of active toolcalls and make an updateStateIfNeeded method that computes the correct state based on that, the playout state, etc and produces the right value, instead of all the individual call-sits making their own determination of what they think state should be?

nbsp: we could keep an array of promises resolved by each of these calls and then Promise.all that array

plugins/openai/src/voice_assistant/index.ts
bcherry: this is actually pretty easy with AgentPlayout and PlayoutHandle, so should be easy to adapt in python as well https://github.com/livekit/agents-js/pull/40/commits/468e17531dd77ca9d3e39355f4f4dddfdcf02fb2

PR #46, opened 2024-09-15
close dangling non-running jobs on drain

nbsp:
also run drain on all jobs asynchronously

### DIFF
--- a/agents/src/worker.ts
+++ b/agents/src/worker.ts
@@ -349,11 +349,14 @@ export class Worker {
     );
 
     const joinJobs = async () => {
-      for (const proc of this.#procPool.processes) {
-        if (proc.runningJob) {
-          await proc.join();
-        }
-      }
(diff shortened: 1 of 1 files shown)


agents/src/worker.ts
```diff
@@ -349,11 +349,14 @@ export class Worker {
...
     const joinJobs = async () => {
-      for (const proc of this.#procPool.processes) {
-        if (proc.runningJob) {
-          await proc.join();
-        }
-      }
+      return Promise.all(
+        this.#procPool.processes.map((proc) => {
+          if (!proc.runningJob) {
+            proc.close();
```
theomonnom, line 355: Kind of confusing that the joinJobs method is also closing some processes

nbsp: it's only closing the processes that are waiting for a job, otherwise just join() waits forever

PR #47, opened 2024-09-16
openai: don't skip waiting for participant

nbsp:
it's not often i get to fix an issue by only removing or moving lines.

### DIFF
--- a/plugins/openai/src/omni_assistant/index.ts
+++ b/plugins/openai/src/omni_assistant/index.ts
@@ -129,7 +129,6 @@ export class OmniAssistant {
       this.participant = participant;
       this.setState(proto.State.INITIALIZING);
 
-
       this.localSource = new AudioSource(proto.SAMPLE_RATE, proto.NUM_CHANNELS);
       this.agentPlayout = new AgentPlayout(this.localSource);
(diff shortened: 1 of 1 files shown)


plugins/openai/src/voice_assistant/index.ts
```diff
@@ -98,31 +98,10 @@ export class VoiceAssistant {
 
         this.linkParticipant(participant.identity);
       });
-      room.on(RoomEvent.TrackPublished, () => {
-        this.subscribeToMicrophone();
-      });
```
bcherry: are you sure its safe to remove this one?

nbsp: it doesn't work without this being removed, it detects its own TrackPublished and tries to subscribe to a mic that doesn't exist

bcherry: but it works reliably to listen to the user? i had some cases where it didn't

nbsp: i didn't have issues, but maybe? maybe we should add this back _after_ the wait for subscription.

bcherry: yeah maybe - just because participant exists and is subscribed doesn't mean they have published a mic track yet

PR #48, opened 2024-09-17
Hotfix protocol for breaking changes

bcherry:
They haven't sent docs yet but it looks like:

- `model_listening` -> `generation_canceled` (this was undocumented before anyways)
- `set_inference_config` -> `update_session_config` (this eliminates the error but system prompt isn't coming through so it must have moved or maybe can only be set via messages now)
- `add_item`, `item_added`, etc -> `add_message`, `message_added` etc
- `item_id` -> `message_id`

I didn't test tool calling but I suspect it's broken too and I haven't tried to sort it out. We can wait for docs.

### DIFF
--- a/plugins/openai/src/voice_assistant/index.ts
+++ b/plugins/openai/src/voice_assistant/index.ts
@@ -255,7 +255,7 @@ export class VoiceAssistant {
         this.handleGenerationFinished(event);
         break;
       case proto.ServerEventType.GENERATION_CANCELED:
-        this.handleGenerationCanceled(event);
+        this.handleGenerationCanceled();
         break;
       case proto.ServerEventType.VAD_SPEECH_STARTED:
         this.handleVadSpeechStarted(event);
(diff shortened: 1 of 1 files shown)


plugins/openai/src/voice_assistant/index.ts
```diff
@@ -169,15 +169,17 @@ export class VoiceAssistant {
...
+      event: proto.ClientEventType.ADD_MESSAGE,
       type: 'message',
-      role: 'user',
-      content: [
-        {
-          type: 'text',
-          text: text,
-        },
-      ],
+      message: {
```
nbsp, line 174: is it now inside `message`? double-checking to make sure, lgtm otherwise

bcherry: yeah it is but i'm not sure of the types/roles/content rules. as is, this now works flawlessly with the minimal example except that the system prompt is ignored

PR #54, opened 2024-09-18
update README with information and links

nbsp:
also added readme-manager stubs, but i don't have a bootstrapped environment to run it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ First, a few concepts:
 - **Plugin**: A library class that performs a specific task, *e.g.* speech-to-text, from a specific
   provider. An agent can compose multiple plugins together to perform more complex tasks.
 
-Your index file for an agent is built of two parts:
+Your main file for an agent is built of two parts:
 
 - The boilerplate code that runs when you run this file, creating a new worker to orchestrate jobs
 - The code that is exported when this file is imported into Agents, to be ran on all jobs (which


README.md
```diff
@@ -4,26 +4,133 @@ SPDX-FileCopyrightText: 2024 LiveKit, Inc.
...
-This is a Node port of the [LiveKit Agents framework](https://livekit.io/agents), originally written in Python.
+# LiveKit Agents for Node.js
+
+<!--BEGIN_DESCRIPTION-->
+
+The Agent Framework is designed for building realtime, programmable participants that run on
+servers. Use it to create conversational, multi-modal voice agents that can see, hear, and
+understand.
+
+This is a Node.js port of the [LiveKit Agents framework](https://livekit.io/agents), originally
```
bcherry: don't call it "port" - maybe "distribution" would be a better word?

README.md
```diff
@@ -4,26 +4,133 @@ SPDX-FileCopyrightText: 2024 LiveKit, Inc.
...
+First, a few concepts:
+
+- **Agent**: A function that defines the workflow of a programmable, server-side participant. This
+  is your application code.
+- **Worker**: A container process responsible for managing job queuing with LiveKit server. Each
+  worker is capable of running multiple agents simultaneously.
+- **Plugin**: A library class that performs a specific task, *e.g.* speech-to-text, from a specific
+  provider. An agent can compose multiple plugins together to perform more complex tasks.
+
+Your index file for an agent is built of two parts:
```
bcherry: none of our examples actually call it `index` so maybe we can use a different name

nbsp: don't wanna call it the agent file though because that implies it needs to be bundled into one file, and "entry" is taken by entrypoint function
main maybe?

PR #56, opened 2024-09-18
feat(elevenlabs): re-add

nbsp:
- [x] audio not encoding properly, this is probably a goof on my end with javascript typed array conversions
- [ ] ~~mp3 decoding needs to be figured out~~ disabled for now


huzaifahj:
Following as we'd love to use an ElevenLabs plugin for Node.

Wondering if you could include an example of using OpenAI Realtime with `modalities: ["text"]` and piping the text output into ElevenLabs to stream audio?

theomonnom (approved):

plugins/elevenlabs/src/tts.ts
```diff
@@ -0,0 +1,259 @@
...
+  closed: boolean;
+  #opts: TTSOptions;
+  #logger = log();
+  readonly streamURL: string;
+
+  constructor(opts: TTSOptions) {
+    super();
+    this.#opts = opts;
+    this.closed = false;
+    this.streamURL = `${this.#opts.baseURL}/text-to-speech/${this.#opts.voice.id}/stream-input?model_id=${this.#opts.modelID}&output_format=${this.#opts.encoding}&optimize_streaming_latency=${this.#opts.streamingLatency}&enable_ssml_parsing=${this.#opts.enableSsmlParsing}`;
```
bcherry: can we either `encodeURIComponent` or use `URLSearchParams` or other helpers to construct this URL in a safe way?

PR #69, opened 2024-09-25
Maximize import compatibility in example

bcherry:
It turns out that some versions of node don't have import.meta.filename but do have import.meta.url, which can be parsed to get the filename.

This change uses the wordier-but-more-compatible version in the example.

I tested it on Node 20.11.0 (which also worked before), and Node 21.1.0 (which did not). It works on both.

### DIFF
--- a/agents/src/generator.ts
+++ b/agents/src/generator.ts
@@ -12,7 +12,7 @@ export interface Agent {
 /** Helper to check if an object is an agent before running it.
  *
  * @internal
-*/
+ */
 export function isAgent(obj: unknown): obj is Agent {
   return (
     typeof obj === 'object' &&


agents/src/worker.ts
```diff
@@ -152,12 +152,15 @@ export class WorkerOptions {
...
+    self_import = undefined,
   }: {
     /**
      * Path to a file that has {@link Agent} as a default export, dynamically imported later for
      * entrypoint and prewarm functions
+     * 
+     * If using the current file, pass `self_import: import.meta` instead.
      */
-    agent: string;
+    agent_file?: string;
```
nbsp: camelCase

PR #72, opened 2024-09-26
Plug RealtimeModel+RealtimeSession into OmniAssistant, get it talking

bcherry:
This handles getting it talking through the new path. Transcripts are still stubbed and functions are broken and some other mess is about.

### DIFF
--- a/plugins/openai/src/realtime/api_proto.ts
+++ b/plugins/openai/src/realtime/api_proto.ts
@@ -173,7 +173,7 @@ export interface SessionResource {
   tools: Tool[];
   tool_choice: ToolChoice; // default: "auto"
   temperature: number; // default: 0.8
-  // max_output_tokens: number | null; // FIXME: currently rejected by OpenAI and fails the whole update
+  max_response_output_tokens: number | null;
 }
 
 // Conversation Resource
@@ -249,7 +249,7 @@ export interface SessionUpdateEvent extends BaseClientEvent {
     tools: Tool[];
     tool_choice: ToolChoice;
     temperature: number;
-    max_output_tokens: number;
+    max_response_output_tokens: number;
   }>;
 }
 
@@ -313,7 +313,7 @@ export interface ResponseCreateEvent extends BaseClientEvent {
     tools?: Tool[];
     tool_choice: ToolChoice;
     temperature: number;
-    max_output_tokens: number;
+    max_response_output_tokens: number;
   }>;
 }
 
--- a/plugins/openai/src/realtime/realtime_model.ts
+++ b/plugins/openai/src/realtime/realtime_model.ts
@@ -43,7 +43,7 @@ interface ModelOptions {
       }
     | 'none';
   temperature: number;
-  maxOutputTokens: number;
+  maxResponseOutputTokens: number;
   apiKey: string;
   baseURL: string;
 }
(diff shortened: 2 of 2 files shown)


plugins/openai/src/index.ts
```diff
@@ -2,4 +2,6 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-export * from './omni_assistant/index.js';
+export * from './omni_assistant.js';
+export * from './agent_playout.js';
+export * from './realtime/index.js';
```
nbsp, line 7: do we want to have this be a namespace? 

```ts
export * as realtime from './realtime/index.js';
```

plugins/openai/src/omni_assistant.ts
```diff
@@ -0,0 +1,398 @@
...
+
+  private started: boolean = false;
+  private participant: RemoteParticipant | string | null = null;
+  private agentPublication: LocalTrackPublication | null = null;
+  private localTrackSid: string | null = null;
+  private localSource: AudioSource | null = null;
+  private agentPlayout: AgentPlayout | null = null;
+  private playingHandle: PlayoutHandle | null = null;
+  private logger = log();
+  private session: RealtimeSession | null = null;
```
nbsp: this is maybe a bigger question for later but we should align on using `private` vs `#` for private attributes and functions. currently other node SDKs are self-consistent at least. the difference is that `#` does compile-time obfuscation so you can't use the private attrs even if you really wanted, while `private` basically just asks nicely.

plugins/openai/src/realtime/api_proto.ts
```diff
@@ -21,6 +21,29 @@ export enum AudioFormat {
...
+}
+
+export enum InputTranscriptionModel {
+  WHISPER_1 = 'whisper-1',
+}
+
+export enum Modality {
+  TEXT = 'text',
+  AUDIO = 'audio',
+}
```
nbsp, line 45: let's edit the protos below and the interfaces in realtime_model to use the enum values mentioned here instead of str literals

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -0,0 +1,858 @@
...
+        threshold?: number;
+        prefix_padding_ms?: number;
+        silence_duration_ms?: number;
+      }
+    | 'none';
+  temperature: number;
+  maxOutputTokens: number;
+  apiKey: string;
+  baseURL: string;
+}
```
nbsp: this is a thing i added that i meant to comment on, i don't like the redundancy of this interface and how it's almost exactly the same as session update options but missing two attrs and with a different two attrs on top. doesn't look very robust if anything were to change.

plugins/openai/src/omni_assistant.ts
bcherry: yeah we should use # - i'm updating some of these but i've been a little sloppy with it. it's hard when porting from python because `_` is used for both true private and internal-only so it's not obvious how to declare in ts

plugins/openai/src/realtime/realtime_model.ts
bcherry: agree

plugins/openai/src/omni_assistant.ts
nbsp: we can set everything to public at first and then have a pass for privating / `@internal`-ing when it's done

PR #73, opened 2024-09-27
feat(job): add waitForParticipant

nbsp:


agents/src/job.ts
```diff
@@ -100,6 +100,22 @@ export class JobContext {
     this.shutdownCallbacks.push(callback);
   }
 
+  async waitForParticipant(identity?: string): Promise<RemoteParticipant> {
```
lukasIO, line 103: this should also throw if the connection transitions to disconnected while waiting

bcherry: https://github.com/livekit/agents-js/pull/75

PR #75, opened 2024-09-28
Fix waitForParticipant

bcherry:
Two bugs:

1. it hangs indefinitely if the participant was already in the room but you didn't pass a specific identity
2. it doesn't apply the identity check in the event callback when waiting is required

### DIFF
--- a/agents/src/job.ts
+++ b/agents/src/job.ts
@@ -105,8 +105,6 @@ export class JobContext {
       throw new Error('room is not connected');
     }
 
-    console.log(this.#room.remoteParticipants.values());
-
     for (const p of this.#room.remoteParticipants.values()) {
       if ((!identity || p.identity === identity) && p.info.kind != ParticipantKind.AGENT) {
         return p;


agents/src/job.ts
```diff
@@ -105,18 +105,34 @@ export class JobContext {
...
 
-    return new Promise((resolve) => {
-      this.#room.once(RoomEvent.ParticipantConnected, () => {
-        resolve(this.#room.remoteParticipants.values().next().value);
-      });
+    return new Promise((resolve, reject) => {
+      const onParticipantConnected = (participant: RemoteParticipant) => {
+        if (
+          (!identity || participant.identity === identity) &&
+          participant.info.kind != ParticipantKind.AGENT
```
bcherry, line 118: we might consider making this say `identity OR not-agent` instead of `identity AND not-agent`, so you could waitForParticipant another agent in a multi-agent scenario.  but python works the way this is written so i'll leave it alone.

agents/src/job.ts
```diff
@@ -105,18 +105,34 @@ export class JobContext {
       throw new Error('room is not connected');
     }
 
-    console.log(this.#room.remoteParticipants.values());
-
     for (const p of this.#room.remoteParticipants.values()) {
       if ((!identity || p.identity === identity) && p.info.kind != ParticipantKind.AGENT) {
```
lukasIO, line 109: We should probably also check against `EGRESS` here (and below) to make sure it doesn't resolve for recorders?

lukasIO: nevermind, they should be `hidden` anyway.

PR #81, opened 2024-09-30
fix(realtime): another round of API changes

nbsp:
doesn't include any real handling of `expires_at`, only sets the default to fifteen minutes.

### DIFF
--- a/plugins/openai/src/realtime/realtime_model.ts
+++ b/plugins/openai/src/realtime/realtime_model.ts
@@ -14,8 +14,8 @@ interface ModelOptions {
   voice: api_proto.Voice;
   inputAudioFormat: api_proto.AudioFormat;
   outputAudioFormat: api_proto.AudioFormat;
-  inputAudioTranscription?: api_proto.InputAudioTranscription;
-  turnDetection?: api_proto.TurnDetectionType;
+  inputAudioTranscription: api_proto.InputAudioTranscription | null;
+  turnDetection: api_proto.TurnDetectionType | null;
   temperature: number;
   maxResponseOutputTokens: number;
   model: api_proto.Model;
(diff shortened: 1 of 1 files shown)


plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -382,6 +389,7 @@ export class RealtimeSession extends multimodal.RealtimeSession {
     turnDetection = this.#opts.turnDetection,
     temperature = this.#opts.temperature,
     maxResponseOutputTokens = this.#opts.maxResponseOutputTokens,
+    expiresAt = this.#opts.expiresAt,
```
bcherry: this one can't be set, it's only present on response. we should consider making session start  async and send  open the socket, send an initial the initial update, and await the acknowledgement instead of sending that stuff unawaited in the constructor. the multimodalagent.start could then await full session and init, and when it returns the session it can already have expiresAt populated for the user. cc @theomonnom

nbsp: agent and session start are both already async fwiw

bcherry: oh that's true - well maybe we need to emit an event when the session gets updated, and def store and expose expiresAt on the session

plugins/openai/src/realtime/api_proto.ts
```diff
@@ -238,12 +235,13 @@ export interface SessionUpdateEvent extends BaseClientEvent {
...
-    input_audio_transcription?: InputAudioTranscription;
-    turn_detection: TurnDetectionType;
+    input_audio_transcription: InputAudioTranscription | null;
+    turn_detection: TurnDetectionType | null;
     tools: Tool[];
     tool_choice: ToolChoice;
     temperature: number;
-    max_response_output_tokens: number;
+    max_response_output_tokens: number | 'inf';
+    expires_at: number;
```
bcherry: i think you need to take this one out too

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -10,14 +10,14 @@ import * as api_proto from './api_proto.js';
...
 interface ModelOptions {
   modalities: ['text', 'audio'] | ['text'];
-  instructions?: string;
+  instructions: string;
   voice: api_proto.Voice;
   inputAudioFormat: api_proto.AudioFormat;
   outputAudioFormat: api_proto.AudioFormat;
   inputAudioTranscription?: api_proto.InputAudioTranscription;
-  turnDetection: api_proto.TurnDetectionType;
+  turnDetection?: api_proto.TurnDetectionType;
```
bcherry: these need to accept `| null` on these params too so you can null them. I think maybe the type itself should include the `| null`, it's part of the type. `undefined` is part of our framework allowing you to omit them

nbsp: oh yeah those should be nullable oops. the Partial already allows them to be undefined

PR #92, opened 2024-10-03
feat(vad): add Silero VAD, overhaul VAD class

nbsp:
- runs on same thread as we don't have `ThreadedJobExecutor` in node
- times and durations may be off due to second/millisecond discrepancies in JavaScript/Python
- plugs in stubs for `AudioResampler` which doesn't exist yet
- `silero_vad.onnx` is licensed under CC-BY-NC-SA-4.0. the license has been added to the repository and the file has been tagged as such using REUSE


nbsp:
Reference-to: https://github.com/livekit/node-sdks/pull/286

agents/src/utils.ts
```diff
@@ -260,39 +260,80 @@ export async function gracefullyCancel<T>(promise: CancellablePromise<T>): Promi
...
+    return this;
+  }
+}
+
+export class ExpFilter {
+  #alpha: number;
+  #max?: number;
+  #filtered?: number = undefined;
+
+  constructor(alpha: number, max: number | undefined = undefined) {
```
bcherry: does this do anything different than `max?: number` would?

agents/src/utils.ts
```diff
@@ -260,39 +260,80 @@ export async function gracefullyCancel<T>(promise: CancellablePromise<T>): Promi
 }
 
 /** @internal */
-export class AsyncIterableQueue<T> implements AsyncIterable<T> {
-  private queue: Queue<T | typeof AsyncIterableQueue.QUEUE_END_MARKER>;
-  private closed = false;
-  private static readonly QUEUE_END_MARKER = Symbol('QUEUE_END_MARKER');
+export class AsyncIterableQueue<T> implements AsyncIterableIterator<T> {
+  protected static readonly QUEUE_END_MARKER = Symbol('QUEUE_END_MARKER');
+  #queue = new Queue<T | typeof AsyncIterableQueue.QUEUE_END_MARKER>();
```
bcherry: reading your vad queue below made me remember the word `sentinel` that i couldn't recall when I wrote this - you should consider renaming it to `CLOSE_SENTINEL`

bcherry (approved):
lgtm minus the build errors on AudioResampler

agents/src/utils.ts
nbsp: nope, just forgot this works because `?:` is picky in typescript sometimes

nbsp:
doesn't run yet, depends on https://github.com/livekit/node-sdks/pull/291. when that's fully sorted, it runs, with a couple major issues:

- `max_buffered_speech` is reached repeatedly, flooding the logs
- probability never goes above 0.1 and hovers around 0.03 to 0.07. this isn't an exponent issue, it's about on par with the numbers we get from silero itself

nbsp:
it's ready now, only annoying thing is how much it floods the logs even though it works great

plugins/silero/src/onnx_model.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export const newInferenceSession = (forceCPU: boolean) => {
+  return InferenceSession.create(new URL('silero_vad.onnx', import.meta.url).pathname, {
+    interOpNumThreads: 1,
+    intraOpNumThreads: 1,
+    executionMode: 'sequential',
+    executionProviders: forceCPU ? [{ name: 'cpu' }] : undefined,
+    extra: {
+      session: {
+        inter_op: { allow_spinning: '0' },
```
theomonnom: Is this valid? Do you have any link around that? I'm not sure to find it

plugins/silero/src/vad.ts
```diff
@@ -0,0 +1,321 @@
...
+          pubPrefixPaddingSamples = Math.ceil(this.#opts.prefixPaddingDuration * pubSampleRate);
+
+          speechBuffer = new Int16Array(
+            (this.#opts.maxBufferedSpeech + this.#opts.prefixPaddingDuration) * pubCurrentSample,
+          );
+
+          if (this.#opts.sampleRate !== pubSampleRate) {
+            // resampling needed: the input sample rate isn't the same as the model's
+            // sample rate used for inference
+            resampler = new AudioResampler(
```
theomonnom, line 148: 🔥

plugins/silero/src/vad.ts
```diff
@@ -0,0 +1,321 @@
...
+          this.queue.put({
+            type: VADEventType.INFERENCE_DONE,
+            samplesIndex: pubCurrentSample,
+            timestamp: pubTimestamp,
+            silenceDuration: pubSilenceDuration,
+            speechDuration: pubSpeechDuration,
+            probability: p,
+            inferenceDuration,
+            frames: [
+              new AudioFrame(inputFrame.data.subarray(0, toCopyInt), pubSampleRate, 1, toCopyInt),
```
theomonnom: Is subarray returning a view? If so, we need to copy instead

nbsp: ah yeah subarray returns a slice of the same array.

plugins/silero/src/onnx_model.ts
nbsp: https://onnxruntime.ai/docs/api/js/interfaces/InferenceSession.SessionOptions.html#extra

theomonnom: Thanks, I'm seeing this tho: 
> This setting is available only in WebAssembly backend. Will support Node.js binding and react-native later

nbsp: didn't notice that. i don't think we can set allow_spinning any other way, though, but we should still remove this

theomonnom (approved):
lgtm, do you have any stats related to the CPU usage?
since allow_spinning can't be disabled the CPU may be at 100%

nbsp:
> lgtm, do you have any stats related to the CPU usage? since allow_spinning can't be disabled the CPU may be at 100%

CPU usage seems to be minimal, no more than 6% even after a while of using it

PR #93, opened 2024-10-04
feat: Make tools inventory overridable per session update

tillkolter:
For better control over the selection of tools for the current segment of the session there should be programmatic control over changing the set of selected functions.


nbsp (approved):

nbsp:
thanks!

PR #99, opened 2024-10-07
Support chat transcription forwarding

lukasIO:


agents/src/transcription.ts
```diff
@@ -118,12 +126,72 @@ export class BasicTranscriptionForwarder implements TranscriptionForwarder {
...
+        await new Promise((resolve) => setTimeout(resolve, sleepInterval * 1000));
+      }
+    }
+
+    if (this.isRunning) {
+      this.close(false);
+    }
+  }
+
+  protected async sendChatMessage() {
```
lukasIO, line 185: one issue with this approach is that we'll eventually run into datapacket size limit restrictions with very long messages. 
should we segment them in the sdk or use a tokenizer of sorts after a certain length?

nbsp:
can you mark the changeset as minor and anchor this to v0.4.0 instead of main?

agents/src/transcription.ts
bcherry: IMO it would be fine to simply truncate at 15KB or up to 63KB and log a warning, for now.  If we do decide to split into multiple messages things get messy, esp with edits. and tokenization-aware splitting may be tricky too in UTF-8 where characters have variable bytesizes...

agents/src/transcription.ts
```diff
@@ -118,12 +126,72 @@ export class BasicTranscriptionForwarder implements TranscriptionForwarder {
...
+      this.currentCharacterIndex = Math.floor(this.currentPlayoutTime * this.charsPerSecond);
+      isComplete = this.textIsComplete && this.currentCharacterIndex >= this.currentText.length;
+      switch (this.type) {
+        case TranscriptionType.CHAT:
+        case TranscriptionType.CHAT_AND_TRANSCRIPTION:
+          await this.sendChatMessage();
+        case TranscriptionType.TRANSCRIPTION:
+        case TranscriptionType.CHAT_AND_TRANSCRIPTION:
+          await this.publishTranscription(false);
+        default:
```
bcherry, line 171: this switch looks wrong to me - won't chat only still fall through to transcription?  a switch doesn't seem like the right conditional type to use for a-b-or-both behavior

agents/src/transcription.ts
```diff
@@ -118,12 +126,72 @@ export class BasicTranscriptionForwarder implements TranscriptionForwarder {
...
     // Publish whatever we had as final
     if (!interrupt) {
-      this.currentCharacterIndex = this.#currentText.length;
+      this.currentCharacterIndex = this.currentText.length;
+    }
+    await this.publishTranscription(true);
+  }
+}
+
+export class ChatAndTranscriptionForwarder extends BasicTranscriptionForwarder {
```
bcherry, line 139: implementing this as a subclass feels unnecessarily brittle and hard to follow - it's already _slightly_ tricky code.

agents/src/transcription.ts
```diff
@@ -14,101 +15,108 @@ export interface TranscriptionForwarder {
   currentCharacterIndex: number;
 }
 
+export enum TranscriptionType {
```
bcherry, line 18: `ForwardingType` would be a clearer name. I'd also consider making the options `Transcription` and `Chat` and the setting an array, rather than a third enum value. then you can simplify the forwarding checks  in the loop to

```
if (forwarding_types.contains('transcription')) {
...
}
if (forwarding_types.contains('chat')) {
...
}
```

bcherry (changes requested):

PR #100, opened 2024-10-07
Include waitForParticipant in minimal example

bcherry:
It's not always necessary but improves reliability (e.g. SIP) and seems worth having in the default case.


nbsp:
this should happen automatically inside the assistant imo, it does on PipelineVoiceAssistant on python iirc

bcherry:
@nbsp it doesn't look like it does - all of the python examples use `participant = await ctx.wait_for_participant()` and then `assistant.start(ctx.room, participant)`

nbsp (approved):
ah, sorry, i think it waits for the track internally, not the participant. lg then

PR #101, opened 2024-10-07
add API validation to workflow

nbsp:
this doesn't use api-extractor, because i've still got no clue how that works. instead it runs TypeDoc validator (with some settings tweaked for sanity).

this PR marked as draft because currently none of openai is documented, and neither is STT/TTS. will come back to this PR once everything else is finalized-ish in terms of API, to minimize merge conflicts later down the line.


bcherry (commented):
awesome thank you

nbsp:
@bcherry i unexported some things and set some others to be internal, please have a look and make sure everything is as it should be (notably unexported api_proto on openai)

bcherry (approved):
love it

nbsp:
lukas is working on a better implementation using api-extractor, i'm waiting for that instead

PR #108, opened 2024-10-14
Fix subscription timing

davidzhao:
Agent fails to subscribe to user track if it was connect to room before agent.

### DIFF
--- a/agents/src/multimodal/multimodal_agent.ts
+++ b/agents/src/multimodal/multimodal_agent.ts
@@ -5,6 +5,8 @@ import type {
   LocalTrackPublication,
   RemoteAudioTrack,
   RemoteParticipant,
+  RemoteTrack,
+  RemoteTrackPublication,
   Room,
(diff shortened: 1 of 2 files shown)


theomonnom (approved):
lg

agents/src/multimodal/multimodal_agent.ts
```diff
@@ -297,14 +304,50 @@ export class MultimodalAgent extends EventEmitter {
...
-        this.#subscribeToMicrophone();
-      });
+    }
+
+    // also check if already subscribed
+    for (const publication of this.linkedParticipant.trackPublications.values()) {
+      if (publication.source === TrackSource.SOURCE_MICROPHONE && publication.track) {
+        this.#handleTrackSubscription(publication.track, publication, this.linkedParticipant);
+        break;
+      }
```
theomonnom, line 314: This part doesn't subscribe if not subscribed. So you may miss an event?

PR #112, opened 2024-10-15
Use shared mutex helper lib

lukasIO:


agents/src/ipc/proc_pool.ts
```diff
@@ -22,12 +23,12 @@ export class ProcPool {
 
   constructor(
     agent: string,
-    numIdleProcesses: number,
+    _numIdleProcesses: number,
```
nbsp, line 26: can you keep this as the same name and just add an eslint-disable to this line? i don't want to break the API here since i'm going to fix this later

nbsp (approved):

agents/src/ipc/proc_pool.ts
lukasIO: This is not API breaking as it's not defined as an object. TS doesn't care about parameter names

PR #116, opened 2024-10-24
Copy README images when generating docs

bcherry:
This should fix the broken image at the top of https://docs.livekit.io/agents-js/


kath0la (approved):
Thanks for fixing this!!

PR #120, opened 2024-10-31
feat(llm): add ChatContext 

nbsp:
incomplete implementation, notably missing tool call hooks (see comment
for more info).

### DIFF
--- a/agents/src/llm/chat_context.ts
+++ b/agents/src/llm/chat_context.ts
@@ -25,6 +25,12 @@ export interface ChatAudio {
 
 export type ChatContent = string | ChatImage | ChatAudio;
 
+const defaultCreateChatMessage = {
+  text: '',
+  images: [],
+  role: ChatRole.SYSTEM,
+};
+
 export class ChatMessage {
   readonly role: ChatRole;
   readonly id?: string;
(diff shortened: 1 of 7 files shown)


agents/src/llm/chat_context.ts
```diff
@@ -0,0 +1,105 @@
...
+  ASSISTANT,
+  TOOL,
+}
+
+export interface ChatImage {
+  image: string | AudioFrame;
+  inferenceWidth?: number;
+  inferenceHeight?: number;
+  /** Used by LLM implementations to store a processed version of the image for later use. */
+  cache: { [id: string | number | symbol]: any };
```
theomonnom, line 22: any way to make sure "cache" is private?

nbsp: to what degree? LLM needs direct access to it, so we can't make it private or readonly. i can mark is as `@internal` so it's clear only llms should use it

theomonnom: internal sounds good

theomonnom (approved):

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -129,12 +138,127 @@ class ConversationItem {
...
   }
 
-  create(item: api_proto.ConversationItemCreateContent, previousItemId?: string): void {
-    this.#session.queueMsg({
-      type: 'conversation.item.create',
-      item,
-      previous_item_id: previousItemId,
-    });
+  // create(item: api_proto.ConversationItemCreateContent, previousItemId?: string): void {
+  create(message: llm.ChatMessage, previousItemId?: string): void {
```
theomonnom, line 142: 🔥

PR #126, opened 2024-11-01
support arrays in function call schema

jb17q:
this PR
* updates `oaiParams` to include the `items` property in arrays to match  openAI's function call schema requirements (fixes issue #124)
* updates  `oaiParams`  to omit optional fields from the `required` section of the generated openAI function call schema
* adds unit tests for `oaiParams`

### DIFF
--- a/.changeset/breezy-pandas-compare.md
+++ b/.changeset/breezy-pandas-compare.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Subscribe to published mic track for linked participant only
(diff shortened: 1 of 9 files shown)


nbsp:
hi! thanks for the fix, especially for the unit tests, we really need to add a full suite of those sooner rather than later ^^;
i've enabled CI tests for you, please check that it all passes fine, and if you could revert the package.json and CHANGELOG changes -- we have a bot handle our versioning for us.

jb17q:
> hi! thanks for the fix, especially for the unit tests, we really need to add a full suite of those sooner rather than later ^^; i've enabled CI tests for you, please check that it all passes fine, and if you could revert the package.json and CHANGELOG changes -- we have a bot handle our versioning for us. additionally since this involves minor changes could you please point this to `v0.4.0` instead of `main`?

Yes, I'll revert the package.json and changelog changes. I'll also point to the v0.4.0 branch instead

PR #130, opened 2024-11-02
feat(llm): add LLM, LLMStream

nbsp:
deferred function calling depended on adding an additional type to function_context.ts. i called it `DeferredFunction`, but maybe @theomonnom has a better name for it?

### DIFF
--- a/agents/src/llm/llm.ts
+++ b/agents/src/llm/llm.ts
@@ -1,0 +1,106 @@
+// SPDX-FileCopyrightText: 2024 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { AsyncIterableQueue } from '../utils.js';
+import type { ChatContext, ChatRole } from './chat_context.js';
+import type {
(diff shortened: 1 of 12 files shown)


agents/src/llm/llm.ts
```diff
@@ -0,0 +1,106 @@
...
+import type {
+  CallableFunctionResult,
+  DeferredFunction,
+  FunctionContext,
+} from './function_context.js';
+
+export interface ChoiceDelta {
+  role: ChatRole;
+  content?: string;
+  toolCalls?: FunctionContext;
```
theomonnom, line 15: This confuses me, we don't have the ctx here on python-agents. We have a list of FunctionInfo

theomonnom (approved):

agents/src/llm/llm.ts
nbsp: this is pretty much the same. note that FunctionContext here isn't our actual context, it's just a key-value record of functions and their names

PR #131, opened 2024-11-03
feat(openai): add LLM

nbsp:
work in progress: grep for TODO(nbsp). might require some more changes to ChatContext. also need to think about how to resize VideoFrames.


plugins/openai/src/llm.ts
```diff
@@ -0,0 +1,659 @@
...
+    this.functionCalls.push(functionInfo);
+
+    return {
+      requestId: id,
+      choices: [
+        {
+          delta: {
+            content: choice.delta.content || undefined,
+            role: llm.ChatRole.ASSISTANT,
+            toolCalls: this.fncCtx,
```
theomonnom: This seems wrong? we should add the function call

plugins/openai/src/llm.ts
```diff
@@ -0,0 +1,659 @@
...
+  const map = new WeakMap();
+
+  return (object: any) => {
+    if (!map.has(object)) {
+      map.set(object, ++currentId);
+    }
+
+    return map.get(object);
+  };
+})();
```
theomonnom: No need to be exactly like Python, feel free to generate an ID inside the constructor

agents/src/llm/llm.ts
```diff
@@ -3,16 +3,12 @@
...
-  DeferredFunction,
-  FunctionContext,
-} from './function_context.js';
+import type { DeferredFunction, FunctionContext } from './function_context.js';
 
 export interface ChoiceDelta {
   role: ChatRole;
   content?: string;
-  toolCalls?: FunctionContext;
+  toolCalls?: DeferredFunction[];
```
theomonnom: We should call it `FunctionCallInfo` to match with Python?

plugins/openai/src/llm.ts
```diff
@@ -0,0 +1,669 @@
...
+      oaiMsg.role = 'user';
+      break;
+    case llm.ChatRole.ASSISTANT:
+      oaiMsg.role = 'assistant';
+      break;
+    case llm.ChatRole.TOOL:
+      oaiMsg.role = 'tool';
+      if (oaiMsg.role === 'tool') {
+        oaiMsg.tool_call_id = msg.toolCallId;
+      }
```
theomonnom, line 593:
```suggestion
      oaiMsg.tool_call_id = msg.toolCallId;
```

plugins/openai/src/llm.ts
```diff
@@ -0,0 +1,669 @@
...
+              completionTokens: usage.completion_tokens,
+              promptTokens: usage.prompt_tokens,
+              totalTokens: usage.total_tokens,
+            },
+          });
+        }
+      }
+    }
+
+    this.queue.close();
```
theomonnom: we should put the queue close inside a try.. finally. 
Otherwise it will never close the AsyncIterator if an error occured

theomonnom (approved):
🚀

agents/src/llm/llm.ts
nbsp: see slack message, i think it might be better to name it differently because it's so fundamentally different than the python impl and it would be more confusing to name it similarly

plugins/openai/src/llm.ts
nbsp: this is for typescript validations, it doesn't work without this type enforcement

theomonnom: Ah lol ok, weird that it isn't smart enough since the previous line literally set it to tool

agents/src/llm/llm.ts
theomonnom: But what happens if I want to simulate a conversation? 
For example I want to load a serialized ChatContext from a database and it contains a CallInfo

nbsp: hmm yeah, that's fair. i'll rename it to functioncallinfo

PR #139, opened 2024-11-07
chore(agents): explicitly unsupport CommonJS

nbsp:
there's a weird edge-case with `tsx` where if `module` is unset in package.json but the typescript file still uses module imports, it will still run, but as CommonJS, without checking for inconsistencies. this doesn't immediately break agents, instead erroring on `job_main.js`, without throwing a useful error.

we do not support and have not supported CommonJS, this PR just throws an error for this specific usecase which has thrown a few folks for a loop.

for reference, compiling to JavaScript and then running with those settings throws an explicit error on Node's side before ever reaching index.js.

### DIFF
--- a/agents/src/index.ts
+++ b/agents/src/index.ts
@@ -15,6 +15,19 @@ import * as multimodal from './multimodal/index.js';
 import * as stt from './stt/index.js';
 import * as tts from './tts/index.js';
 
+const isCommonJS = (): boolean => {
+  try {
+    return !!require;
+  } catch {
+    return false;
+  }
+};
+if (isCommonJS()) {
+  throw new ReferenceError(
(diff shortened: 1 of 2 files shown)


agents/src/index.ts
```diff
@@ -15,6 +15,19 @@ import * as multimodal from './multimodal/index.js';
...
+const isCommonJS = (): boolean => {
+  try {
+    return !!require;
+  } catch {
+    return false;
+  }
+};
+if (isCommonJS()) {
+  throw new ReferenceError(
+    '@livekit/agents cannot be used in a CommonJS environment. Please set your project to use ES modules.',
```
lukasIO: we could explicitly tell them to set the module field i. package.json, that might be clearer

lukasIO (approved):
lgtm!

PR #150, opened 2024-11-11
chore(ci): enable LFS in publish script

nbsp:
silero vad is broken currently on npm because the LFS file wasn't pulled


davidzhao (approved):
lg. this should work

PR #160, opened 2024-11-15
feat(openai): emit Realtime API response text delta and done

danielmahon:
This adds simple support for reading the responses of a text-only Realtime API Model. I found no way currently to get the responses as transcriptions don't seem to be created when only the "text" modality is chosen. 

In our case, we patched these changes and forwarded the text to the agent's publication track so our client handles audio and text transcriptions the same way.


nbsp (approved):

nbsp:
thanks!

PR #165, opened 2024-11-17
fix(elevenlabs): override with defaults if opts are provided

gching:
Providing any `opts` to `elevenlabs.TTS` without the `apiKey` will result in an error since we are only merging the `opts` with the defaults after the check
ex.
```ts
// Will throw an error since an apiKey isn't provided and the defaults with the env variable isn't provided
elevenlabs.TTS({}) 
```

### Changes
* Set and merge `#opts` with defaults before `apiKey` check
* Remove defaulting `opts` argument to `defaultTTSOptions` since its technically not needed

### DIFF
--- a/.changeset/fifty-apes-whisper.md
+++ b/.changeset/fifty-apes-whisper.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents-plugin-elevenlabs": patch
+---
+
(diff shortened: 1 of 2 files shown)


nbsp (approved):
thanks!

PR #167, opened 2024-11-17
fix(agent): ensure token stream flushes

gching:
Was noticing the last word for TTS was not being outputted (Websocket, ElevenLabs).
The input stream was not being flushed, and therefore, the remaining token was not being sent upstream.

### Changes
* Ensure `BufferedTokenStream` ends the input and is flushed before closing
* Implement missing wrapping methods for `BufferedWordStream` so the underlying `WordStream` implementations isn't used

### Reproduction
* Use voice agent pipeline with ElevenLabs
* Every spoken prompt would be missing the last word

### Notes
* `WordStream` can probably be converted to an interface since the underlying implementation is technically not needed

### DIFF
--- a/.changeset/itchy-sheep-smash.md
+++ b/.changeset/itchy-sheep-smash.md
@@ -1,0 +1,6 @@
+---
+"@livekit/agents": patch
+"@livekit/agents-plugin-elevenlabs": patch
+---
+
+Ensure token stream flushes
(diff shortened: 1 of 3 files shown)


nbsp (approved):
thanks a ton! caught this with SentenceStream, didn't think to add it here as well.

nbsp:
ack re: WordStream as an interface, might address that in a future PR

PR #168, opened 2024-11-18
chore(tsconfig): enable `noUncheckedIndexedAccess`

nbsp:
used `!` for places where we either check earlier, or we know for absolutely certain that there's gonna be a value.

### DIFF
--- a/agents/src/ipc/job_main.ts
+++ b/agents/src/ipc/job_main.ts
@@ -110,10 +110,11 @@ if (process.send) {
   process.on('SIGINT', () => {});
 
   await once(process, 'message').then(([msg]: IPCMessage[]) => {
-    if (msg!.case !== 'initializeRequest') {
+    msg = msg!;
+    if (msg.case !== 'initializeRequest') {
       throw new Error('first message must be InitializeRequest');
     }
-    initializeLogger(msg!.value.loggerOptions);
+    initializeLogger(msg.value.loggerOptions);
   });
   const proc = new JobProcess();
   let logger = log().child({ pid: proc.pid });
--- a/agents/src/ipc/job_main.ts
+++ b/agents/src/ipc/job_main.ts
@@ -94,7 +94,7 @@ if (process.send) {
   //   [1] import.meta.filename
   //   [2] import.meta.filename of function containing entry file
   const moduleFile = process.argv[2];
-  const agent: Agent = await import(pathToFileURL(moduleFile).href).then((module) => {
+  const agent: Agent = await import(pathToFileURL(moduleFile!).href).then((module) => {
     const agent = module.default;
     if (agent === undefined || !isAgent(agent)) {
       throw new Error(`Unable to load agent: Missing or invalid default export in ${moduleFile}`);
@@ -110,10 +110,10 @@ if (process.send) {
   process.on('SIGINT', () => {});
 
   await once(process, 'message').then(([msg]: IPCMessage[]) => {
-    if (msg.case !== 'initializeRequest') {
+    if (msg!.case !== 'initializeRequest') {
       throw new Error('first message must be InitializeRequest');
     }
-    initializeLogger(msg.value.loggerOptions);
+    initializeLogger(msg!.value.loggerOptions);
   });
   const proc = new JobProcess();
   let logger = log().child({ pid: proc.pid });
--- a/agents/src/ipc/proc_job_executor.ts
+++ b/agents/src/ipc/proc_job_executor.ts
@@ -122,7 +122,7 @@ export class ProcJobExecutor extends JobExecutor {
     this.#proc!.send({ case: 'initializeRequest', value: { loggerOptions } });
     await once(this.#proc!, 'message').then(([msg]: IPCMessage[]) => {
       clearTimeout(timer);
-      if (msg.case !== 'initializeResponse') {
+      if (msg!.case !== 'initializeResponse') {
         throw new Error('first message must be InitializeResponse');
       }
     });
--- a/agents/src/job.ts
+++ b/agents/src/job.ts
@@ -190,7 +190,7 @@ export class JobContext {
   onParticipantConnected(p: RemoteParticipant) {
     for (const callback of this.#participantEntrypoints) {
       if (
-        p.identity in this.#participantTasks &&
+        this.#participantTasks[p.identity] &&
         this.#participantTasks[p.identity]!.callback == callback
       ) {
         this.#logger.warn(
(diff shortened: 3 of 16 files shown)


agents/src/ipc/job_main.ts
```diff
@@ -110,10 +110,10 @@ if (process.send) {
   process.on('SIGINT', () => {});
 
   await once(process, 'message').then(([msg]: IPCMessage[]) => {
-    if (msg.case !== 'initializeRequest') {
+    if (msg!.case !== 'initializeRequest') {
       throw new Error('first message must be InitializeRequest');
     }
-    initializeLogger(msg.value.loggerOptions);
+    initializeLogger(msg!.value.loggerOptions);
```
lukasIO: the double `!` here and two lines above could be reduced to a single check of msg

agents/src/llm/function_context.ts
```diff
@@ -133,7 +133,7 @@ export const oaiBuildFunctionInfo = (
 
   return {
     name: fncName,
-    func: fncCtx[fncName],
+    func: fncCtx[fncName]!,
```
lukasIO: a temp var here could also make sense to use it both for the checkin in L130 and here

agents/src/tokenize/basic/hyphenator.ts
```diff
@@ -51,7 +51,7 @@ class Hyphenator {
     // if the word is an exception, get the stored points
     let points: number[];
     if (this.exceptions[word.toLowerCase()]) {
-      points = this.exceptions[word.toLowerCase()];
+      points = this.exceptions[word.toLowerCase()]!;
```
lukasIO: maybe initialize `points` with this and execute the `else` clause if points is undefined?

agents/src/worker.ts
```diff
@@ -610,17 +610,19 @@ export class Worker {
...
         return asgn;
       });
 
-      await this.#procPool.launchJob({
-        acceptArguments: args,
-        job: msg.job!,
-        url: asgn.url || this.#opts.wsURL,
-        token: asgn.token,
-      });
+      if (asgn) {
```
lukasIO, line 618: do we want a warning if `asgn` is not defined?

examples/src/stt.ts
```diff
@@ -25,7 +25,7 @@ export default defineAgent({
       const recvTask = async () => {
         for await (const event of sttStream) {
           if (event.type === stt.SpeechEventType.FINAL_TRANSCRIPT) {
-            console.log(event.alternatives[0].text);
+            console.log(event.alternatives[0]!.text);
```
lukasIO: is it safe to assume that final transcript will have `event.alternatives` populated under all circumstances with at least one element?

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -1006,7 +1006,7 @@ export class RealtimeSession extends multimodal.RealtimeSession {
         `[Function Call ${item.call_id}] Executing ${item.name} with arguments ${parsedArgs}`,
       );
 
-      this.#fncCtx[item.name].execute(parsedArgs).then(
+      this.#fncCtx[item.name]?.execute(parsedArgs).then(
```
lukasIO: warn if the function did not get executed as it's not present?

examples/src/stt.ts
nbsp: yes because we check for it inside the STT class. this is a one-or-more array

plugins/openai/src/realtime/realtime_model.ts
nbsp: i'm not sure if this is info that would be useful to the user per se but sure

examples/src/stt.ts
lukasIO: we could define it as a one-or-more array on the typescript side then with `alternatives: [elementType, ...elementType[]]`

nbsp: well actually not quite. it's an empty array in case of `{START,END}_OF_SPEECH` because it's unused and it's easier to make it empty than have some sort of complicated typescript type depending on enum value for this. but for final transcript it is guaranteed. maybe i can make alternatives `|undefined` and then the `!` would be on it instead for better readability

agents/src/ipc/proc_job_executor.ts
```diff
@@ -122,7 +122,7 @@ export class ProcJobExecutor extends JobExecutor {
     this.#proc!.send({ case: 'initializeRequest', value: { loggerOptions } });
     await once(this.#proc!, 'message').then(([msg]: IPCMessage[]) => {
       clearTimeout(timer);
-      if (msg.case !== 'initializeResponse') {
+      if (msg!.case !== 'initializeResponse') {
```
gching, line 125:
```suggestion
      if (!msg || msg.case !== 'initializeResponse') {
```
Not sure if doing `!` is temporary to unblock this PR and we are tackling this later, but I do prefer to just do safety checks instead of depending on `!` to guarantee not having runtime issues

In this case, I know `msg` should always be defined, but in the case `msg` is nullish anyways, we should probably throw an error?

nbsp: i don't think we should recover in any way from this. this would signify a serious breakage that should probably throw ugly javascript errors rather than quietly/silently continuing on

nbsp: there's not much to add here with our own error that JavaScript wouldn't complain about anyway, regardless

gching: > there's not much to add here with our own error that JavaScript wouldn't complain about anyway, regardless

Oh definitely agree, makes sense that the runtime error would accomplish the same thing 😄 - sounds good this is nice!

agents/src/job.ts
```diff
@@ -190,8 +190,8 @@ export class JobContext {
   onParticipantConnected(p: RemoteParticipant) {
     for (const callback of this.#participantEntrypoints) {
       if (
-        p.identity in this.#participantTasks &&
-        this.#participantTasks[p.identity].callback == callback
+        this.#participantTasks[p.identity] &&
+        this.#participantTasks[p.identity]!.callback == callback
```
gching:
```suggestion
        this.#participantTasks[p.identity]?.callback == callback
```
Nit: Could simplify by coalescing since if its nullish, it would not be equal

examples/src/stt.ts
lukasIO: sounds good!

nbsp: already pushed!

PR #169, opened 2024-11-18
feat(multimodal): add speech committed events

### LINKED ISSUE #164: Multimodel Agent to support `user_speech_committed` and `agent_speech_committed` events
Currently they are supported on Python but not Node yet.

nbsp:
this changes the API for transcription forwarder a tiny bit as a stopgap in order to make this work. won't be necessary when the better transcription comes along, and eventually transcription v2.
fixes #164

### DIFF
--- a/agents/src/multimodal/multimodal_agent.ts
+++ b/agents/src/multimodal/multimodal_agent.ts
@@ -19,7 +19,7 @@ import {
 } from '@livekit/rtc-node';
 import { EventEmitter } from 'node:events';
 import { AudioByteStream } from '../audio.js';
-import type * as llm from '../llm/index.js';
+import * as llm from '../llm/index.js';
 import { log } from '../log.js';
 import { BasicTranscriptionForwarder } from '../transcription.js';
 import { findMicroTrackId } from '../utils.js';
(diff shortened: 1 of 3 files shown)


agents/src/transcription.ts
```diff
@@ -35,6 +36,10 @@ export class BasicTranscriptionForwarder implements TranscriptionForwarder {
     this.#messageId = messageId;
   }
 
+  get text(): string {
+    return this.#currentText;
```
bcherry, line 40: `#currentText` may have been a poor choice of name because I think (unless I misunderstand this detail of this PR) that you expect it to be only the currently cursored chunk of text? it gets filled in by the realtime api as text is generated, but this happens quickly. then it has a second cursor that we maintain that syncs to playout.

I think you intend to use `.text` here as meaning "the text that has played". if so, then you should use `this.#currentText.slice(0, this.currentCharacterIndex)`

agents/src/multimodal/multimodal_agent.ts
```diff
@@ -178,10 +178,26 @@ export class MultimodalAgent extends EventEmitter {
...
       };
 
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
       const onPlayoutStopped = (interrupted: boolean) => {
         this.emit('agent_stopped_speaking');
         this.#speaking = false;
+        if (this.#playingHandle) {
+          let text = this.#playingHandle.transcriptionFwd.text;
+          if (interrupted) {
+            text += '…';
```
bcherry, line 187: see below - i think you need to change the definition of `text` to correctly get text that's truncated at the point of interruption

nbsp: in general i couldn't get it to emit interrupted events; interruptions work but it just swallows the attempted event and starts a new one without throwing interrupt

agents/src/transcription.ts
nbsp: that actually breaks things: sometimes the cursor is too early and the end of the transcription gets cut out. `this.#currentText` works great without it

theomonnom (approved):

agents/src/multimodal/multimodal_agent.ts
nbsp: i'll tackle this in a later PR, for now non-interrupt works great and i'd rather work on this later

PR #170, opened 2024-11-19
Use peer dependencies for @livekit/rtc-node and @livekit/agents

lukasIO:


plugins/deepgram/package.json
```diff
@@ -20,8 +20,10 @@
     "typescript": "^5.0.0"
   },
   "dependencies": {
-    "@livekit/agents": "workspace:*",
-    "@livekit/rtc-node": "^0.11.1",
     "ws": "^8.16.0"
+  },
+  "peerDependencies": {
+    "@livekit/agents": "workspace:*",
```
nbsp:
```suggestion
    "@livekit/agents": "workspace:^",
```

so that it doesn't request the exact version and error if not found
this will match 0.x.y to the next minor version

lukasIO: good point, but should we instead maybe use `~` so that it doesn't require the one set in the workspace as a minimum?

nbsp: it would still require it as a minimum — in 0.x they are equivalent, ~ does it until next minor and ^ until next major
if we want to just accept any value shipped with agents (which should be fine unless you're manually upgrading only one for some reason), we can do simply `*` without `workspace:`

lukasIO: oh, you're right, I always had a understanding of _any_ patch version for tilde.

lukasIO:
had to mirror the peerDeps in devDeps because of https://github.com/vercel/turborepo/issues/1601

nbsp (approved):

PR #172, opened 2024-11-19
Add missing package info

lukasIO:


nbsp (approved):
lukewarm on having changesets for this, it's not really a source thing and we don't include changesets for CI either. other than that lg

lukasIO:
yeah, not a functionality change, but it requires a new release in order to take effect, that's why I included it.

Issue #177, opened 2024-11-22
"interruptions are not allowed"


hamza-ghufran:
seems like there is an issue with `pipeline_agent.ts` regarding interruptions

1. In `validateReplyIfPossible()`, the condition for checking interruption allowance is reversed:

https://github.com/livekit/agents-js/blob/4a77de81eb1dc35485f6acc6ad2c45cfe29e4790/agents/src/pipeline/pipeline_agent.ts#L774

```
// Current 
if (!speech.allowInterruptions) speech.interrupt();

// should be the reverse, correct ? 
```

2. The first condition in `validateReplyIfPossible()`
https://github.com/livekit/agents-js/blob/4a77de81eb1dc35485f6acc6ad2c45cfe29e4790/agents/src/pipeline/pipeline_agent.ts#L750

```
// Should be:
if (this.#playingSpeech && !this.#playingSpeech.allowInterruptions)
```

facing this issue (see screenshot)
![Screenshot 2024-11-22 at 2 04 35 PM](https://github.com/user-attachments/assets/5f09bddb-a091-4d3b-9f49-3fa731450606)

with current code as:

```
  const agent = new pipeline.VoicePipelineAgent(
    vad,
    new deepgram.STT({ fillerWords: true, model: 'nova-2-general' }),
    new openai.LLM({ model: 'gpt-4o-mini', temperature: 0.0 }),
    new openai.TTS({ voice: 'shimmer' }),
    {
      chatCtx: initialContext,
      allowInterruptions: false,
      minEndpointingDelay: 2.5,
    }
  );
  ```

PR #178, opened 2024-11-22
fix(agent): correct interruption logic in VoicePipelineAgent

hamza-ghufran:
fixes: https://github.com/livekit/agents-js/issues/177


nbsp (approved):
great, thank you!

PR #184, opened 2024-11-25
Fix a couple of bugs

posobin:


plugins/deepgram/src/stt.ts
```diff
@@ -261,12 +261,12 @@ export class SpeechStream extends stt.SpeechStream {
                   if (isFinal) {
                     this.queue.put({
                       type: stt.SpeechEventType.FINAL_TRANSCRIPT,
-                      alternatives: [alternatives[0], ...alternatives.splice(0)],
```
posobin, line 264: This just duplicates the first item in the array. Not sure what the intention was, perhaps to shallowly copy the array?

agents/src/pipeline/pipeline_agent.ts
```diff
@@ -693,31 +693,31 @@ export class VoicePipelineAgent extends (EventEmitter as new () => TypedEmitter<
...
 
-      if (handle.addToChatCtx && (!userQuestion || handle.userCommitted)) {
-        this.chatCtx.messages.push(...extraToolsMessages);
-        if (interrupted) {
-          collectedText + '…';
-        }
-
-        const msg = ChatMessage.create({ text: collectedText, role: ChatRole.ASSISTANT });
-        this.chatCtx.messages.push(msg);
+    if (handle.addToChatCtx && (!userQuestion || handle.userCommitted)) {
```
posobin, line 698: This was nested in an if when I think it wasn't supposed to be so AGENT_SPEECH_INTERRUPTED/COMMITTED were never firing

theomonnom (approved):
Thanks!

plugins/deepgram/src/stt.ts
```diff
@@ -261,12 +261,14 @@ export class SpeechStream extends stt.SpeechStream {
...
+                      // Ensures that TS knows that the array is not empty
```
nbsp: can you delete these comments? these are helpful at review time but not so much as code

nbsp (approved):
looks good, thanks! just one nitpick and i'll merge it.

plugins/deepgram/src/stt.ts
posobin: I don't agree (would have been useful to me when reading this code), but sure, removed

PR #190, opened 2024-11-29
feat: add metrics

nbsp:


agents/src/llm/llm.ts
```diff
@@ -48,16 +59,56 @@ export abstract class LLM {
 }
 
 export abstract class LLMStream implements AsyncIterableIterator<ChatChunk> {
+  protected output = new AsyncIterableQueue<ChatChunk>();
   protected queue = new AsyncIterableQueue<ChatChunk>();
```
theomonnom, line 63: The two queues are kind of confusing. If I'm a user implementing a plugin, I would not be sure which one to use

theomonnom (approved):

agents/src/llm/llm.ts
nbsp: i'm currently looking into replacing AsyncIterableQueue with web streams across the whole framework, which have a tee function

PR #194, opened 2024-12-02
chore: move to node:stream/web from homegrown API

nbsp:
this PR gets rid of our homegrown `AsyncIterableQueue` implementation in favour of WHATWG's Streams API.

- [ ] go over plugins
- [ ] implement locking
- [ ] figure out how to check whether a writer is closed


nbsp:
still doesn't work, there's a rejected promise somewhere

PR #200, opened 2024-12-07
test(tokenizer): add tests + fix tokenizer

nbsp:
while writing this patch, i found that using `bun test` instead of vitest is much _much_ (up to 100x) faster. i know we don't explicitly target Bun but it might be worth using it in our CI for testing instead of vitest, especially with later integration tests.

### DIFF
--- a/.changeset/few-schools-run.md
+++ b/.changeset/few-schools-run.md
@@ -2,4 +2,4 @@
 "@livekit/agents": patch
 ---
 
-add tests + fix tokenizer
+fix tokenizer
--- a/.changeset/few-schools-run.md
+++ b/.changeset/few-schools-run.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+add tests + fix tokenizer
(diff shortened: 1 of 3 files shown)


agents/src/tokenize/basic/sentence.ts
```diff
@@ -67,3 +76,7 @@ export const splitSentences = (text: string, minLength = 20): [string, number, n
 
   return sentences;
 };
+
+const rawRegex = (r: RegExp): string => {
```
lukasIO: wouldn't this be the same as `r.source` ?

nbsp: i knew there was a thing for this but i couldn't find it for the life of me :woman_facepalming:

.changeset/few-schools-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+add tests + fix tokenizer
```
lukasIO: should the tests bit be part of the changelog?

lukasIO (approved):

.changeset/few-schools-run.md
nbsp: nope, fixing

PR #212, opened 2024-12-15
test: run conditional tests only for changed files

nbsp:
run all tests if tests or agents are changed, otherwise run only relevant plugin tests.


lukasIO (approved):
would be nice if we could replace some of the custom logic with already existing actions

PR #225, opened 2024-12-22
feat: add turn detector

nbsp:
- [x] inference_executor
- [x] inference_proc_executor
- [x] inference_proc_lazy_main
- [x] job_executor
- [x] job_proc_executor
- [ ] job_proc_lazy_main
- [ ] proc_client
- [x] proc_pool
- [ ] supervised_proc (sort of)
- [x] message


nbsp:
ready for review, only missing two things that i know of:
- the inference process doesn't do the right thing with shutdown events, so it always exits as orphaned
- i will deal with the merge conflicts Some Other Time (there's a lot of them)

agents/src/worker.ts
```diff
@@ -180,6 +184,8 @@ export class WorkerOptions {
     port = undefined,
     logLevel = 'info',
     production = false,
+    jobMemoryWarnMB = 300,
```
theomonnom, line 187: Do we check the memory somewhere?

theomonnom (approved):

agents/src/worker.ts
```diff
@@ -284,11 +295,27 @@ export class Worker {
...
+    if (Object.entries(InferenceRunner.registeredRunners).length) {
+      this.#inferenceExecutor = new InferenceProcExecutor(
+        InferenceRunner.registeredRunners,
+        30000,
+        5000,
+        2000,
+        0,
+        5000,
+        60000,
+        2500,
```
theomonnom: We should clean this, hard to know which number is for what

agents/src/worker.ts
nbsp: nope, it's currently stubbed, i have a comment for it somewhere that i forgot to go look at

PR #238, opened 2025-01-06
fix(worker): error can be a non-string

nbsp:


agents/src/worker.ts
```diff
@@ -331,6 +331,11 @@ export class Worker {
           this.#runWS(this.#session);
           return;
         } catch (e) {
```
lukasIO: let's type this as `unknown`

agents/src/worker.ts
```diff
@@ -331,6 +331,11 @@ export class Worker {
...
+          if (typeof e !== 'string') {
+            // it's probably an Error type
```
lukasIO: instead of assuming this another type checking for error instanceof Error would be a lot cleaner

lukasIO (approved):

PR #239, opened 2025-01-08
feat(openai): Add whisper prompt support to STT

FlorDonnaSanders:
Hey there!

OpenAI's whisper model supports the option of supplying a [system prompt](https://platform.openai.com/docs/api-reference/audio/createTranscription#audio-createtranscription-prompt) when calling the transcription API.  
We make use of this feature heavily and adding support for it should be as simple as adding the two proposed line changes.

Let me know what needs to happen before this can be officially supported in Agents JS.

### DIFF
--- a/.changeset/happy-coins-drop.md
+++ b/.changeset/happy-coins-drop.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-openai': minor
+---
+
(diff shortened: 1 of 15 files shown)


nbsp:
can you rebase your PR onto `livekit/agents-js:next` instead?

nbsp (approved):

PR #243, opened 2025-01-13
fix(multimodal): re-request audio response

nbsp:
request audio response when a text response is given instead.

this pulls from https://github.com/livekit/agents/pull/1015 and https://github.com/livekit/agents/pull/1121 but is not a full port of the former; it just fixes this specific issue to unblock folks while i work on the bigger PR.

i still have not managed to reproduce the error, so, consider this untested.


longcw:
lgtm!

davidzhao:
@longcw wanna approve as well? you'll have more context on this PR than either theo or I

longcw (approved):

PR #244, opened 2025-01-13
WIP data stream transcription forwarder

lukasIO:
A simplified version of a transcription forwarder using data streams. 

Instead of a publishing loop (which will cause unhandled exceptions in user land if something goes wrong) it queues text updates as they come in and delays the execution for the next text update for a bit to mimic the previous forwarding delays on char index basis.


nbsp:
superseded by #301

lukasIO:
@nbsp the PR you linked does not handle data stream forwarding though, right?

nbsp:
ahh, that's true, the forwarding part is actually still done with the old method. still, this PR is pretty outdated and uses old APIs, so let's keep it closed

Issue #250, opened 2025-01-14
Additional Deepgram Plugin config


NikoLandgraf:
Some parameters are missing from the Deepgram config. 

Most importantly the possibility to opt out of their model improvement program using the `mip_opt_out`:
https://developers.deepgram.com/docs/the-deepgram-model-improvement-partnership-program#want-to-opt-out

There is a similar (pull) request in the python plugin:
https://github.com/livekit/agents/pull/1020

PR #264, opened 2025-01-17
feat: update stt options

s-hamdananwar:


nbsp:
potentially also worth incorporating https://github.com/livekit/agents-js/issues/250 here

s-hamdananwar:
> potentially also worth incorporating #250 here

what do you think about creating a new PR for this? this PR is already pretty loaded with changes I believe

PR #271, opened 2025-01-23
fix(proc): clearer errors on crash

nbsp:
we see a lot of bug reports focusing on ERR_IPC_CHANNEL_CLOSED, because it is the *big scary exception*, when in reality the actual error gets logged a few lines up. this gets rid of the useless log and should hopefully lead to more useful bug reports

### DIFF
--- a/agents/src/ipc/proc_job_executor.ts
+++ b/agents/src/ipc/proc_job_executor.ts
@@ -94,9 +94,9 @@ export class ProcJobExecutor extends JobExecutor {
       }
     };
     this.#proc!.on('message', listener);
-    this.#proc!.on('error', (err) => {
+    this.#proc!.on('error', () => {
       if (this.#closing) return;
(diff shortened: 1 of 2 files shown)


agents/src/ipc/proc_job_executor.ts
```diff
@@ -94,9 +94,9 @@ export class ProcJobExecutor extends JobExecutor {
       }
     };
     this.#proc!.on('message', listener);
-    this.#proc!.on('error', (err) => {
+    this.#proc!.on('error', () => {
       if (this.#closing) return;
-      this.#logger.child({ err }).warn('job process exited unexpectedly');
+      this.#logger.warn('job process exited unexpectedly');
```
lukasIO: I think it would be good to at least keep the error message for debugging purposes though in cases where it might not be that exact error that's causing it? 

We can include "look for the actual error further up in the logs" or something like that

lukasIO (approved):

Issue #268, opened 2025-01-22
[feature] Update options on Silero VAD plugin


FlorDonnaSanders:
Hi!

For testing and development purposes it would be nice to update the VAD settings in a live session.  
Taking a look at the Python agents reference, it seems updating the options is possible [there](https://docs.livekit.io/python/livekit/plugins/silero/index.html#livekit.plugins.silero.VAD.update_options).

It would be nice to have this same functionality in the JS framework.
I'll see if I can make some time to contribute to this myself.

nbsp:
we're working on adding `updateOptions(...)` methods to all relevant plugins by 0.7 - contributions welcome!

PR #274, opened 2025-01-23
feat(silero): update options for Silero VAD

FlorDonnaSanders:
Implementation of `updateOptions` function on Silero VAD (Issue #268).

I Took heavy inspiration for this from the [python implementation](https://github.com/livekit/agents/blob/695f7b567d0b7fffa6a18f14c74e50421434a0aa/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py#L217), though I did not look too deep at the differences between `#task` and `_main_task`, so there may be implications I am not aware of.

### DIFF
--- a/plugins/silero/src/vad.ts
+++ b/plugins/silero/src/vad.ts
@@ -377,9 +377,9 @@ export class VADStream extends baseStream {
   /**
    * Update the VAD options
    *
-   * This method allows you to update the VAD options after the VAD object has been created
-   *
    * @param opts - Partial options object containing the values to update
+   * @remarks
+   * This method allows you to update the VAD options after the VAD object has been created
    */
   updateOptions(opts: Partial<VADOptions>) {
     const oldMaxBufferedSpeech = this.#opts.maxBufferedSpeech;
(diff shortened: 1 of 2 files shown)


FlorDonnaSanders:
@nbsp converted to draft since I'm not sure if 100% functional (see comment above).
Feel free to make changes as needed.

plugins/silero/src/vad.ts
```diff
@@ -340,4 +373,39 @@ export class VADStream extends baseStream {
       }
     });
   }
+
+  /**
+   * Update the VAD options
+   *
+   * This method allows you to update the VAD options after the VAD object has been created
```
nbsp: put this under `@remarks`

plugins/silero/src/vad.ts
```diff
@@ -46,12 +46,29 @@ const defaultVADOptions: VADOptions = {
 export class VAD extends baseVAD {
   #session: InferenceSession;
   #opts: VADOptions;
+  #streams: VADStream[];
```
nbsp, line 49: i'm concerned about the lack of a cleanup function here, in case one of the VAD streams needs to be closed (though we don't have a close function, either). potentially out of scope for this PR, though

nbsp (approved):
looks good!

plugins/silero/src/vad.ts
ghost: As there is currently no close function on the VAD streams, I do indeed propose to tackle this under a different PR.

PR #283, opened 2025-01-28
include name, arguments, result & error (as applicable) in function_call_XXX events so they are much more useful

arthurblake:
These events are perfect for observing when a function is called, completes or throws an error so that additional steps can be taken by the agent that cannot normally run directly within the function context (such as taking some additional action before hanging up or transferring a call). But the events as coded unfortunately did not included any context about the function in the event payload. Adding that!

### DIFF
--- a/plugins/openai/src/realtime/realtime_model.ts
+++ b/plugins/openai/src/realtime/realtime_model.ts
@@ -1101,11 +1101,13 @@ export class RealtimeSession extends multimodal.RealtimeSession {
         return;
       }
 
-      this.emit('function_call_started', {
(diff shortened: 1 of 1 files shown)


arthurblake:
Link to slack discussion about the need for this change: https://livekit-users.slack.com/archives/C07FRP66XGE/p1736962087704659

arthurblake:
Hello. Is there anything I can do to get this PR merged?
@nbsp

nbsp (approved):

nbsp:
not merging for now as this would be a minor bump, but lgtm.

PR #300, opened 2025-02-13
Use #opts.apiVersion

brianbscho:
It's using `2024-10-01-preview` hardcoded value without chance to change apiVersion.

In [Azure document](https://learn.microsoft.com/en-us/azure/ai-services/openai/realtime-audio-quickstart?tabs=keyless%2Clinux&pivots=programming-language-typescript), it is stated that: "Use the latest `2024-12-17` model version".


theomonnom (approved):

aaronwangj:
@theomonnom @s-hamdananwar There are a bunch of Livekit users using realtime API and we're blocked without this. Can we get this into main?

PR #301, opened 2025-02-13
replace transcription forwarder with synchronizer

nbsp:
- [x] there's a slight delay, causing it to exit early somtimes
- [x] it only works on function calls (??)
- [x] it stops (breaks?) after the first sentence
- [x] isn't implemented yet on pipeline
  - [x] breaks on interrupt or function call
  - [x] figure out AGENT_SPEECH_INTERRUPTED

### DIFF
--- a/agents/src/transcription.ts
+++ b/agents/src/transcription.ts
@@ -121,7 +121,8 @@ export class TextAudioSynchronizer extends (EventEmitter as new () => TypedEmitt
     }
 
     this.#textData!.done = true;
-    this.#textData!.sentenceStream.close()
+    this.#textData?.sentenceStream.flush()
+    this.#textData?.sentenceStream.close()
     this.#textData = undefined;
   }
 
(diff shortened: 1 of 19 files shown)


agents/src/transcription.ts
```diff
@@ -1,134 +1,308 @@
...
   }
 
-  #computeSleepInterval(): number {
-    return Math.min(Math.max(1 / this.#charsPerSecond, 0.0625), 0.5);
+  async close(interrupt: boolean) {
+    if (this.#closed) {
+      return;
+    }
+    this.#closed = true;
+    this.#interrupted = interrupt;
```
theomonnom, line 156: Do we need to store the interrupted state inside the TextSynchronizer? This seems like it isn't its responsability vs the VPA (The boolean inside close caught my eye because it was unclear what it was for)

nbsp: this is used in the sentence synchronizer, because the sync is sometimes a bit too slow, and the text doesn't show up fully. the functionality now --
- if interrupted, it will stop the text
- if not interrupted, it will concatenate the rest of the unwritten text to the end

theomonnom (approved):
not fan of the boolean inside close, otherwise lgtm.

PR #317, opened 2025-03-06
fix(multimodal): never feed a negative audio end

nbsp:
this is a patchy fix to get stuff to not break as fast as possible, but we should look into why the length can be negative in the first place.

### DIFF
--- a/agents/src/multimodal/multimodal_agent.ts
+++ b/agents/src/multimodal/multimodal_agent.ts
@@ -335,7 +335,7 @@ export class MultimodalAgent extends EventEmitter {
           this.#session!.conversation.item.truncate(
             this.#playingHandle.itemId,
             this.#playingHandle.contentIndex,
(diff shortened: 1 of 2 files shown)


theomonnom (approved):

agents/src/multimodal/multimodal_agent.ts
```diff
@@ -335,7 +335,7 @@ export class MultimodalAgent extends EventEmitter {
           this.#session!.conversation.item.truncate(
             this.#playingHandle.itemId,
             this.#playingHandle.contentIndex,
-            Math.floor((this.#playingHandle.audioSamples / 24000) * 1000),
+            Math.max(0, Math.floor((this.#playingHandle.audioSamples / 24000) * 1000)),
```
lukasIO, line 338: opened #319 which I think addresses the underlying issue. Still not pretty, but a bit closer to the root cause

PR #332, opened 2025-03-17
fix(worker): reconnect on severed websocket conn

nbsp:


agents/src/worker.ts
```diff
@@ -483,10 +482,14 @@ export class Worker {
     };
     this.event.on('worker_msg', send);
 
-    ws.addEventListener('close', () => {
-      closingWS = true;
-      this.#logger.error('worker connection closed unexpectedly');
-      this.close();
+    const close = new Promise<void>((resolve) => {
```
lukasIO, line 485: additionally an error listener that logs the ws error would be nice!

wmurphyrd:
Could we have this released, please? It's taking down entire clusters of production agent servers about once week for us

jflemingframe:
`@livekit/agents-plugin-silero` did not receive an update, but the other packages we use did. 
Is the fix not relevant to that?

PR #335, opened 2025-03-18
Keyterm fix for nova-3-general

tomc98:
The `keywords` isn't supported by nova-3-general, it has been changed to `keyterm` which is quite similar, it's just missing the multiplier at on the words. I believe it's best to abstract this away and make it work as I lost hours trying to upgrade to nova-3-general. A lot of people use `keywords` and the error messages aren't good when it fails from deepgram.


nbsp:
superseded by #337, to match how the Python version does it.

PR #338, opened 2025-03-18
Multimodal: Don't mark segment end if cancelled

pettersamuelsen:
Fixes https://github.com/livekit/agents-js/issues/311#issuecomment-2734518910

PR #342, opened 2025-03-20
fix: resolve race condition in TextAudioSynchronizer closure

xavierroma:
## Description

The race condition occurs because the `TextAudioSynchronizer` is closed before the readText task is cancelled. When the readText task continues running and calls methods on the already-closed synchronizer, it triggers the "TextAudioSynchronizer is closed" error. This fix ensures all tasks are properly cancelled before closing the synchronizer.

### DIFF
--- a/.changeset/fix-synchronizer-race.md
+++ b/.changeset/fix-synchronizer-race.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+fix: fix race condition in TextAudioSynchronizer causing "TextAudioSynchronizer is closed" errors in AgentPlayout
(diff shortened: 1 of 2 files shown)


nbsp:
i believe this was already fixed in https://github.com/livekit/agents-js/pull/338, let me know if you still get the error on the `next` branch (as-yet-unreleased)

xavierroma:
You are right that fixed it, hadn't seen that I started from main. However, imo this PR is still relevant as it fixes the root cause: shutdown sequence, rather than just fixing one symptom. @nbsp

nbsp (approved):

PR #351, opened 2025-04-04
feat(openai): add new TTS model and voices, include optional instruct…

tomc98:
The new openAI voices and model are now available, they are fully compatible with the currently used endpoint + implementation. See here: https://platform.openai.com/docs/api-reference/audio/createSpeech
This change is to allow instructions on the new tts model.

### DIFF
--- a/.changeset/selfish-ducks-compare.md
+++ b/.changeset/selfish-ducks-compare.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents-plugin-openai": patch
+---
+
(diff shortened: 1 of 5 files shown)


davidzhao (approved):
lgtm!

tomc98:
We need to update the openai package to latest

davidzhao:
could you update it in this PR before merge?

tomc98:
I'm trying to get it sorted now :)

tomc98:
Ayyooo it builds @davidzhao

PR #358, opened 2025-04-11
Add noise cancellation support to VoicePipelineAgent and MultimodalAgent

typester:
- For VoicePipelineAgent: Added to VPAOptions interface
  - For MultimodalAgent: Added as constructor parameter

Regarding the implementation for MultimodalAgent, I've currently added an extra parameter to the constructor, but I believe this isn't the best approach. Similar to how we added `AudioStreamOptions` to `AudioStream`, I think it would be better to create a parameter that consolidates options.
However, since I lack context on this project, I'd appreciate your opinion on how to best structure this.

### DIFF
--- a/.changeset/noise-cancellation.md
+++ b/.changeset/noise-cancellation.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+add inbound noise cancellation support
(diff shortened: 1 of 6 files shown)


bcherry (approved):
from my perspective this seems fine as is. I don't think it's a breaking change like this and we'll be working on adopting 1.0 interface soon anyways so I'd suggest we keep our focus there but def up to the agents team!

Shubhrakanti (approved):

lukasIO (approved):

theomonnom (approved):

PR #370, opened 2025-04-23
refactor(job): show job request attrs on print

### LINKED ISSUE #346: Agent request callback always contains empty job
### Describe the bug

The `job` argument passed to `requestFunc` will always be an empty object so there seems to be no way to have an agent decide whether it should join a room without explicit dispatch. Which is not what I would expect given the documentation:

https://docs.livekit.io/agents/build/anatomy/#request-handler

### Relevant log output

_No response_

### Describe your environment

node v22

"@livekit/agents": "^0.7.1",
"livekit-server-sdk": "^2.10.2",

```ts
cli.runApp(
	new WorkerOptions({
		production: isProduction,
		agent: fileURLToPath(import.meta.url),
		host: '0.0.0.0',
		port: Number.parseInt(process.env.PORT ?? '8088'),
		requestFunc: async (job) => {
			console.log('Job Requested...');
			console.log(job); // always returns {}
			await job.accept();
		},
	}),
);
```

### Additional information

_No response_

nbsp:
this is technically not a bug, just a weird thing that JavaScript does, where getters aren't shown on `console.log()`, so it looks like the object is empty. this PR just sets them statically as readonly.

closes #346

### DIFF
--- a/.changeset/tall-apes-hug.md
+++ b/.changeset/tall-apes-hug.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+refactor(job): show job request attrs on print
--- a/agents/src/job.ts
+++ b/agents/src/job.ts
@@ -255,9 +255,19 @@ export class JobProcess {
  * requests should fill idle processes and which should be outright rejected.
  */
 export class JobRequest {
-  #job: proto.Job;
   #onReject: () => Promise<void>;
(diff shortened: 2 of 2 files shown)


Shubhrakanti (approved):
Clutch. Can you add to the PR description what printing the job should look like? 

Also let's make sure we let the person that filed this know it's been updates.

PR #371, opened 2025-04-23
Fix frame size from TTS output and bug where we don't queue frames

Shubhrakanti:
Allow the agent playout to be buffered. Previously we were waiting every Frame to be acknowledged by the FFiclient before we queued the next one.

### DIFF
--- a/agents/src/pipeline/agent_playout.ts
+++ b/agents/src/pipeline/agent_playout.ts
@@ -167,7 +167,6 @@ export class AgentPlayout extends (EventEmitter as new () => TypedEmitter<AgentP
           }
           handle.pushedDuration += (frame.samplesPerChannel / frame.sampleRate) * 1000;
           handle.synchronizer.pushAudio(frame);
-          console.log('captureFrame with frame duration (ms)', frame.samplesPerChannel / frame.sampleRate * 1000);
           await this.#audioSource.captureFrame(frame);
         }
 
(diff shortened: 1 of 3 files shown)


agents/src/pipeline/agent_playout.ts
```diff
@@ -168,15 +168,10 @@ export class AgentPlayout extends (EventEmitter as new () => TypedEmitter<AgentP
...
           await this.#audioSource.captureFrame(frame);
-          await this.#audioSource.waitForPlayout();
         }
 
-        // XXX(nbsp): line 161 waits instead of this. this is not the case on python agents,
-        //            but for some reason too many TTS frames can gunk up the buffer and lead to
-        //            FFI errors. this works 🤷‍♀️
-        // if (this.#audioSource.queuedDuration > 0) {
-        //   await this.#audioSource.waitForPlayout();
-        // }
```
Shubhrakanti, line 179: Was never able to reproduce this.

agents/src/pipeline/agent_playout.ts
```diff
@@ -168,15 +168,10 @@ export class AgentPlayout extends (EventEmitter as new () => TypedEmitter<AgentP
...
+        await this.#audioSource.waitForPlayout();
```
Shubhrakanti, line 173: I still think we need to wait for playout once all the frames have been submitted to the FFI handle (similar to https://github.com/livekit/agents/blob/9a122d896254a1b8401f94b56b3609cf0f266e9c/livekit-agents/livekit/agents/voice/agent_activity.py#L648-L649)

I wasn't however able to observe any behavior change in the agent when I removed this line.

PR #374, opened 2025-04-24
feat(deepgram): add updateOptions

nbsp:


ben-achilleshr (approved):

plugins/deepgram/src/stt.ts
```diff
@@ -238,7 +256,7 @@ export class SpeechStream extends stt.SpeechStream {
     );
 
     const listenTask = async () => {
-      while (!this.closed) {
+      while (!this.closed && !closing) {
```
Shubhrakanti, line 259: JC: was this causing an issue? If so - how did you notice it?

plugins/deepgram/src/stt.ts
```diff
@@ -312,7 +330,9 @@ export class SpeechStream extends stt.SpeechStream {
       }
     };
 
-    await Promise.all([sendTask(), listenTask(), wsMonitor]);
+    await Promise.race([this.#resetWS.await, Promise.all([sendTask(), listenTask(), wsMonitor])]);
```
Shubhrakanti, line 333: Do we propagate exceptions from the underlying tasks? In python we have something like 

```python
for task in done:
          if task != wait_reconnect_task:
                  task.result()
```

plugins/deepgram/src/stt.ts
nbsp: this wasn't causing an issue, i just added this so that when the old websocket closes the loop breaks as opposed to listening indefinitely to new outputs from a closed websocket

Shubhrakanti (approved):
Nice!

PR #375, opened 2025-04-24
feat(cartesia): add updateOptions

nbsp:


Shubhrakanti (approved):
Nice! 

Any ideas how we can test the `update_options` flow?

PR #376, opened 2025-04-25
Fix inference job not shutting down on SIGINT

Shubhrakanti:


agents/src/ipc/supervised_proc.ts
```diff
@@ -185,15 +189,11 @@ export abstract class SupervisedProc {
     }
     this.#closing = true;
 
-    if (!this.#runningJob) {
-      this.proc!.kill();
-      this.#join.resolve();
-    }
```
Shubhrakanti, line 191: No point killing it _before_ we send the shut down request

agents/src/ipc/supervised_proc.ts
```diff
@@ -142,6 +141,11 @@ export abstract class SupervisedProc {
       this.#join.resolve();
     });
 
+    this.proc!.on('exit', (code, signal) => {
+      this.#logger.child({ code, signal }).debug('job process exited');
+      this.#join.resolve();
+    });
```
Shubhrakanti: wait for the subprocess to actually shut down before we do the shutdown in the main process.

agents/src/ipc/inference_proc_lazy_main.ts
```diff
@@ -84,8 +82,11 @@ const ORPHANED_TIMEOUT = 15 * 1000;
           });
           break;
         case 'shutdownRequest':
-          closeEvent.emit('close');
-          break;
+          logger.info('inference process received shutdown request');
+          process.send!({ case: 'done' });
+          clearTimeout(orphanedTimeout);
+          process.off('message', handleInferenceRequest);
+          process.exit();
```
theomonnom: We should let the process cleanups gracefully? process.exit() seems like a force kill?

Shubhrakanti: Oh I see. Yeah we do this in job proc too. We should just exit from the main loop and have it finish like in python. Let me look into it.

theomonnom (approved):

agents/src/ipc/inference_proc_lazy_main.ts
Shubhrakanti: @theomonnom this didn't workout. See #418 for details.

PR #383, opened 2025-05-05
Use WHATWG streams to forward AudioFrames from Livekit room into VAD

Shubhrakanti:
**Please read this [slack message](https://live-kit.slack.com/archives/C07TWVC0W4A/p1746659738184179) before reviewing this PR**

Sketch out how we would use [WHATWG Streams](https://streams.spec.whatwg.org) to power the agents 1.0 in JS.

### Functionality
Run `examples/src/basic_agent.ts` in `dev` mode. 

Waits for **one** participant to join the room. Streams audio from the Livekit agents playground and prints `Start of speech` and `End of speech` based on voice activity.

### Main files to review

- `room_io.ts` handles setting up `ReadableStream` that forwards audio once the source is available.
- `audio_recognition.ts` handles forwarding the input stream to the VAD and forwarding the VAD events to predefined hooks

### Non functional edits
- `agent.ts` Just sketches out what an agent class would look like. Doesn't affect the functionality.

### DIFF
--- a/agents/src/deferred_stream.ts
+++ b/agents/src/deferred_stream.ts
@@ -1,0 +1,61 @@
+import { ReadableStream } from 'node:stream/web';
+import { Future } from './utils.js';
+
+export class DeferredReadableStream<T> {
+  private _sourceFuture: Future<ReadableStream<T>>;
+
+  private _reader?: ReadableStreamDefaultReader<T>;
+
+  public readonly stream: ReadableStream<T>;
+
+  constructor() {
+    this._sourceFuture = new Future<ReadableStream<T>>();
+
+    this.stream = new ReadableStream<T>({
+      start: async (controller) => {
+        try {
+          const source = await this._sourceFuture.await;
+
+          this._reader = source.getReader();
+
+          const pump = async () => {
+            try {
+              while (true) {
+                const { done, value } = await this._reader!.read();
+                if (done) break;
+                controller.enqueue(value);
+              }
+              controller.close();
+            } catch (err) {
+              controller.error(err);
+            }
+          };
+
+          pump();
+        } catch (err) {
+          controller.error(err);
+        }
+      },
+      cancel: async (reason) => {
+        await this.cancel(reason);
+      },
+    });
+  }
+
+  /**
+   * Call once the actual source is ready.
+   */
+  setSource(source: ReadableStream<T>) {
+    if (this._sourceFuture.done) {
+      return;
+    }
+    this._sourceFuture.resolve(source);
+  }
+
+  async cancel(reason?: any) {
+    if (!this._sourceFuture.done) {
+      this._sourceFuture.reject(reason);
+    }
+    await this._reader?.cancel();
+  }
+}
--- a/agents/src/deferred_stream.ts
+++ b/agents/src/deferred_stream.ts
@@ -55,10 +55,9 @@ export class DeferredReadableStream<T> {
     this._sourceFuture.resolve(source);
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  async cancel(reason?: any) {
+  async cancel(reason?: Error) {
     if (!this._sourceFuture.done) {
-      this._sourceFuture.reject(reason);
+      this._sourceFuture.reject(reason ?? new Error('Stream cancelled without reason'));
     }
     await this._reader?.cancel(reason);
   }
--- a/agents/src/deferred_stream.ts
+++ b/agents/src/deferred_stream.ts
@@ -1,3 +1,6 @@
+// SPDX-FileCopyrightText: 2025 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
 import { ReadableStream } from 'node:stream/web';
 import { Future } from './utils.js';
 
--- a/agents/src/deferred_stream.ts
+++ b/agents/src/deferred_stream.ts
@@ -52,10 +52,11 @@ export class DeferredReadableStream<T> {
     this._sourceFuture.resolve(source);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   async cancel(reason?: any) {
     if (!this._sourceFuture.done) {
       this._sourceFuture.reject(reason);
     }
-    await this._reader?.cancel();
+    await this._reader?.cancel(reason);
   }
 }
--- a/agents/src/vad.ts
+++ b/agents/src/vad.ts
@@ -1,9 +1,10 @@
 // SPDX-FileCopyrightText: 2024 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import type { AudioFrame } from '@livekit/rtc-node';
+import type { AudioFrame, AudioStream } from '@livekit/rtc-node';
 import type { TypedEventEmitter as TypedEmitter } from '@livekit/typed-emitter';
 import { EventEmitter } from 'node:events';
+import { log } from './log.js';
 import type { VADMetrics } from './metrics/base.js';
 import { AsyncIterableQueue } from './utils.js';
 
@@ -83,10 +84,28 @@ export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
   protected closed = false;
   #vad: VAD;
   #lastActivityTime = BigInt(0);
+  private logger = log();
+  private inputAudioStream: Promise<AudioStream>;
+  private inputAudioStreamResolver: (value: AudioStream) => void = () => {};
 
   constructor(vad: VAD) {
     this.#vad = vad;
+    this.inputAudioStream = new Promise((resolve) => {
+      this.inputAudioStreamResolver = resolve;
+    });
     this.monitorMetrics();
+    this.main_task();
+  }
+
+  protected async main_task() {
+    try {
+      const inputStream = await this.inputAudioStream;
+      for await (const frame of inputStream) {
+        this.pushFrame(frame);
+      }
+    } catch (error) {
+      this.logger.error('Error in VADStream main_task:', error);
+    }
   }
 
   protected async monitorMetrics() {
@@ -120,6 +139,10 @@ export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
       }
     }
     this.output.close();
+  }
+
+  updateInputStream(audioStream: AudioStream) {
+    this.inputAudioStreamResolver(audioStream);
   }
 
   pushFrame(frame: AudioFrame) {
--- a/agents/src/vad.ts
+++ b/agents/src/vad.ts
@@ -94,10 +94,10 @@ export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
       this.inputAudioStreamResolver = resolve;
     });
     this.monitorMetrics();
-    this.main_task();
+    this.mainTask();
   }
 
-  protected async main_task() {
+  protected async mainTask() {
     // This is just a placeholder since VAD isn't implemented with the streams API yet.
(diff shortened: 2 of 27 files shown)


agents/src/vad.ts
```diff
@@ -83,10 +84,28 @@ export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
   protected closed = false;
   #vad: VAD;
   #lastActivityTime = BigInt(0);
+  private logger = log();
+  private inputAudioStream: Promise<AudioStream>;
+  private inputAudioStreamResolver: (value: AudioStream) => void = () => {};
```
Shubhrakanti: @lukasIO wonder if there is a better way to do this.

agents/src/voice/agent_session.ts
```diff
@@ -0,0 +1,153 @@
...
+  private _agent?: Agent;
+  private _activity?: AgentActivity;
+  private _nextActivity?: AgentActivity;
+  private started = false;
+
+  private roomIO: RoomIO | null = null;
+  private logger = log();
+
+  /** @internal */
+  audioInput: ParticipantAudioInputStream | null = null;
```
Shubhrakanti: We need to be able to close these streams (ex. python channels [can be closed](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/utils/aio/channel.py#L132-L146)). Wondering how we can do that with an abort controller.

agents/src/voice/agent_session.ts
```diff
@@ -0,0 +1,153 @@
...
+export class AgentSession<T = any> extends EventEmitter {
+  vad?: VAD;
+
+  private _updateActivityTask: Promise<void> | null = null;
+  private _agent?: Agent;
+  private _activity?: AgentActivity;
+  private _nextActivity?: AgentActivity;
+  private started = false;
+
+  private roomIO: RoomIO | null = null;
```
Shubhrakanti: these should be undefined correct?

agents/src/voice/audio_recognition.ts
```diff
@@ -0,0 +1,65 @@
...
+        case VADEventType.INFERENCE_DONE:
+          this.hooks.onVADInferenceDone(ev);
+          break;
+      }
+    }
+  }
+
+  setInputAudioStream(audioStream: ReadableStream<AudioFrame>) {
+    this.inputAudioStreamResolver(audioStream);
+  }
```
Shubhrakanti: I'm wondering if it's just better to pass the input into the constructor.

Shubhrakanti: I think that's cleaner than doing `updateInputStream` but what happens when we don't have the input stream in the constructor?

agents/src/voice/room_io.ts
```diff
@@ -0,0 +1,147 @@
...
+ * iterator for ReadableStream, because Chrome has not implemented it yet.
+ *
+ * See https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/65542#discussioncomment-6071004
+ * for more details.
+ *
+ * Since we are only running this in node, we can use the any type to get around this.
+ */
+function asCompatibleStream<T>(stream: ReadableStream<T>): any {
+  return stream;
+}
```
Shubhrakanti: @lukasIO not sure what's a better way to do this. This is only an issue with types, since this is a node only sdk.

context: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/65542#discussioncomment-6071004

agents/src/voice/room_io.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (!this.audioStream || !this.controller || this.isCancelled) {
+      this.logger.warn('Audio stream not available');
+      return;
+    }
+
+    try {
+      for await (const frame of asCompatibleStream(this.audioStream)) {
+        if (this.isCancelled) {
+          break;
+        }
```
Shubhrakanti: @lukasIO @theomonnom  What do you think of cancelling this with an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)? Something like 

```typescript
private async readAudioFrames(abortController: AbortController): Promise<void> {
  ...
  for await (const frame of asCompatibleStream(this.audioStream)) {
    if (abortController.signal.aborted) {
      break;
    }
    ...
  }
}
```

although this feels the exact same as what we just did. 

We can also do this with `Promise.race` and the cancel resolver being set when we close the stream. Any recommendations on what's the best option?

agents/src/voice/room_io.ts
lukasIO: I don't think that's the reason you're seeing an issue. We definitely don't want to cast to `any`.

due to the fact that node has two different `ReadableStream` implementations you want to make sure that you import `import type { ReadableStream } from "node:stream/web"` in every file that makes use of the streams. Otherwise you'll get incompatibilities as TS will interpret it as the node (non-web) stream.

agents/src/voice/room_io.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  private async readAudioFrames(): Promise<void> {
+    if (!this.audioStream || !this.controller || this.isCancelled) {
+      this.logger.warn('Audio stream not available');
+      return;
+    }
+
+    try {
+      for await (const frame of asCompatibleStream(this.audioStream)) {
```
lukasIO: due to the unclear spec around `for await of` with tee'd streams that you found, let's not use this pattern internally and instead do the iteration with `read()`

agents/src/voice/room_io.ts
```diff
@@ -0,0 +1,145 @@
...
+ * See https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/65542#discussioncomment-6071004
+ * for more details.
+ *
+ * Since we are only running this in node, we can use the any type to get around this.
+ */
+function asCompatibleStream<T>(stream: ReadableStream<T>): any {
+  return stream;
+}
+
+class ParticipantAudioSource implements UnderlyingSource<AudioFrame> {
```
lukasIO: I know we've talked about potential advantages of this deferred stream being a source itself, but thinking more now, I feel like there might be more upsides to create something like a `DeferredStream` helper class for internal usage and only expose the raw `ReadableStream<AudioFrame>` to users. 

Usage pattern could look like 

```ts
const deferred = new DeferredStream<AudioFrame>();

// give to users immediately
const participantInputStream = deferred.stream;

// later on track subscribed
deferred.setSourceStream(new AudioStream(track));
```

agents/src/voice/room_io.ts
```diff
@@ -0,0 +1,145 @@
...
+  private cleanup(): void {
+    this.isCancelled = true;
+    this.logger.debug('Participant audio source cancelled');
+
+    if (this.audioStream) {
+      this.audioStream.cancel();
+      this.audioStream = undefined;
+    }
+
+    this.room.off(RoomEvent.TrackSubscribed, this.onTrackAvailable.bind(this));
```
lukasIO: two nits: 

- for consistency we usually default to arrow functions declared as properties on the class instead of the `bind(this)` pattern. I find the `bind` pattern harder to reason about also.
- if we only care about one subscribed track, then unregistering the listener can already happen as soon as the first track is successfully set up

agents/src/voice/room_io.ts
lukasIO: the only condition to cancel is when the source is cancelled, right? I think this would become easier to handle with the `read()` and deferred stream pattern as mentioned above. We could call `controller.close` and `controller.error` respectively on the deferred stream

lukasIO: I think this would become easier to handle with the `DeferredStream` helper, we could simply forward the cancel request to the source stream

agents/src/voice/agent_session.ts
lukasIO: we can simply `cancel` readable streams, right?

agents/src/voice/room_io.ts
Shubhrakanti: Ahh I see. thanks!

agents/src/voice/room_io.ts
Shubhrakanti: > if we only care about one subscribed track

This is just temporary to get some boiler plate working. In the future we we will need something like https://github.com/livekit/agents/blob/992f3d1f7854b16e806915360d67e200a12b2920/livekit-agents/livekit/agents/voice/room_io/_input.py#L19-L23.

Shubhrakanti: > unregistering the listener can already happen as soon as the first track is successfully set up

Not sure what you mean by this? do you mean we can call `this.room.off(RoomEvent.TrackSubscribed, this.onTrackAvailable.bind(this));` immediately after `this._deferredAudioInputStream.setSource(new AudioStream(track));`?

davidzhao: `this.onTrackAvailable.bind(this)` probably doesn't work as expected.. when you create a new bind, it's technically a new function

Shubhrakanti: Re: "this.onTrackAvailable.bind(this)"

Yeah I removed this. mistake on my end.

Shubhrakanti: This is also not recommended by [google style guide](https://google.github.io/styleguide/tsguide.html#rebinding-this)

agents/src/deferred_stream.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  setSource(source: ReadableStream<T>) {
+    if (this._sourceFuture.done) {
+      return;
+    }
+    this._sourceFuture.resolve(source);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async cancel(reason?: any) {
```
Shubhrakanti: not sure if there's a better way to handle this. should we explicitly parse the type in 
```typescript
cancel: async (reason) => {
         await this.cancel(reason);
       },
``` ? Or is this okay

File: agents/src/voice/agent.ts [hunk 11]
```diff
@@ -0,0 +1,93 @@
+// SPDX-FileCopyrightText: 2025 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Shubhrakanti, line 7: Will enable later. This class is just skeleton for now. https://linear.app/livekit/issue/AJS-29/enable-es-list-on-agent-class

agents/src/deferred_stream.ts
lukasIO: pushed an update for this, sorry I didn't catch this

agents/src/voice/room_io.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(agentSession: AgentSession, room: Room) {
+    this.agentSession = agentSession;
+    this.room = room;
+    this.participantAudioInputStream = this._deferredAudioInputStream.stream;
+
+    this.setupEventListeners();
+  }
+
+  private setupEventListeners() {
+    this.room.on(RoomEvent.TrackSubscribed, this.onTrackSubscribed);
```
lukasIO, line 29: for the happy path this looks fine, but when things go wrong, we'd likely want to cancel the deferred stream for situations where e.g. the agent disconnects from the room unexpectedly

Shubhrakanti: Totally agree. Right now this is just a place holder. We will handle bring the room_io to the 1.0 standards in separate PRs. want to keep this PR just boiler plate enough to get started.

lukasIO (approved):

PR #385, opened 2025-05-08
Shubhra/add stt to audio recognition

Shubhrakanti:
Add STT node to Agents 1.0. End of turn prediction will be added in a separate PR.

PR #390, opened 2025-05-14
Shubhra/ajs 31 refactor vad with streams

Shubhrakanti:
Refactor the VAD component to use WHATWG streams.

We use two [Identity TransformStreams](https://streams.spec.whatwg.org/#identity-transform-stream) to represent input and output.

Some parts of the `VAD` component are deprecated because of this refactor. Will remove once we do away with `VoicePipelineAgent`. Keeping it around for now because we're still developing.


Shubhrakanti:
I tried to get this to work by implementing a `VadSrouce` (similar to how we did Audio/VideoStream in `node-sdk`) but didn't seem very natural. 

You can see my progress on this commit. https://github.com/livekit/agents-js/pull/390/commits/dd1436262c3df8847130de3fe6c11412ac5bd08f

agents/src/vad.ts
```diff
@@ -80,46 +84,73 @@ export abstract class VAD extends (EventEmitter as new () => TypedEmitter<VADCal
 
 export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
```
Shubhrakanti, line 85: I don't think this should implement `AsyncIterableIterator`. I think it would be better to have a method

```typescript

get stream(): ReadableStream<VADEvent> {
  return this.output.readable
}

```

agents/src/voice/agent_activity.ts
```diff
@@ -47,8 +47,9 @@ export class AgentActivity implements RecognitionHooks {
     this.logger.info('End of speech', ev);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   onVADInferenceDone(ev: VADEvent): void {
-    this.logger.info('VAD inference done', ev);
+    // this.logger.info('VAD inference done', ev);
```
Shubhrakanti: ignore this. Forgot to pick this up form another PR's changes. Will resolve with merge conflict

agents/src/voice/audio_recognition.ts
```diff
@@ -37,7 +37,9 @@ export class AudioRecognition {
 
   start() {
     this.vadStreamProcessor = this.vadTask().catch((err) => {
-      this.logger.error('Error in VAD task', err);
+      // raise the error
+      throw err;
+      this.logger.error(`Error in VAD task: ${err}`);
     });
```
Shubhrakanti: What's the proper way to handle errors from these async functions we run in node? I'll do some digging on this. Ideally we want to raise them rather than swallow them.

agents/src/vad.ts
lukasIO: can you elaborate on why (out of curiousity)?

agents/src/voice/audio_recognition.ts
```diff
@@ -37,7 +37,9 @@ export class AudioRecognition {
 
   start() {
     this.vadStreamProcessor = this.vadTask().catch((err) => {
-      this.logger.error('Error in VAD task', err);
+      // raise the error
+      throw err;
```
lukasIO: this will always throw as an unhandled exception, which we'd want to avoid. 

You're trying to handle an error in an async task from within a non-async method. I think `catch` is the best you can do there

agents/src/vad.ts
Shubhrakanti: Yeah this isn't a huge issue for the VAD, but the other components (LLM/TTS/STT) expect the output to be a `ReadableStream` for their default implementations (see [this comment](https://github.com/livekit/agents-js/pull/385/files#r2087921711)). This way the default usage of the components can just be 

```typescript
static default = {
     async [stt/llm/tts]Node(inputStream: ReadableStream<...>): Promise<ReadableStream<...>> {
          ...
          stream = [stt/llm/tts].stream()
          stream.updateInputStream(inputStream)
          return stream.stream()
     }
}
```

[Reference python implementation ](https://github.com/livekit/agents/blob/e57f6a0a4cf07190ca4b60aff6acdadc2cebce90/livekit-agents/livekit/agents/voice/agent.py#L411-L443)

I just want to keep all the `[STT/LLM/TTS/VAD]Stream` classes similar.

Shubhrakanti: Although in this case it makes more sense for `VADStream` to be called `VADStreamManager` or something.

lukasIO: I see, that makes sense to me!

Shubhrakanti:
Injectable stream.

agents/src/vad.ts
Shubhrakanti: I'll do this in a separate PR. There are some tricky things with the legacy implementation not worth wrestling right now since we will delete that anyway. https://linear.app/livekit/issue/AJS-48/remove-iterator-from-vadsream

PR #392, opened 2025-05-19
Simplify deferred stream helper

lukasIO:


agents/src/stream/identity_transform.ts
```diff
@@ -0,0 +1,12 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+import { TransformStream } from 'node:stream/web';
+
+export class IdentityTransform<T> extends TransformStream<T, T> {
+  constructor() {
+    super({
+      transform: (chunk, controller) => controller.enqueue(chunk),
+    });
+  }
```
Shubhrakanti, line 11: Why have a separate slack for this? Instead of a `pump` function in `DeferredReadableStream`.

Shubhrakanti (approved):

agents/src/stream/identity_transform.ts
lukasIO: The PR's intention was to avoid the `pump()` pattern we had previously. 

The reason for this one is just convenience. I think we'll have more scenarios where an IdentityTransformStream will come in handy.

Shubhrakanti: Yeah that makes sense to me.

lukasIO: Arguably we could even get rid of the `DeferredStream` as the only thing it adds on top of the identity stream is the `setSource` thing, wdyt?

Shubhrakanti: > Arguably we could even get rid of the DeferredStream as the only thing it adds on top of the identity stream is the setSource thing, wdyt?

Yeah I'm down. Less LOC. The only thing `DeferredStream` adds is a little better readability, but we can just make the variable name of the input transform stream `deferredInputStream`.

lukasIO: I'll merge this for now as is and we can iterate over it if we realise `DeferredStream` as a construct seems too convoluted

PR #394, opened 2025-05-19
Shubhra/ajs 30 end of turn detection in audio recogntion

Shubhrakanti:


agents/src/voice/audio_recognition.ts
```diff
@@ -99,7 +120,70 @@ export class AudioRecognition {
...
+      }
+    };
+
+    if (this.bounceEOUAbortController) {
+      this.bounceEOUAbortController.abort();
+    }
+
+    this.bounceEOUAbortController = new AbortController();
+
+    bounceEOUTask(this.lastSpeakingTime, this.bounceEOUAbortController.signal);
```
Shubhrakanti: This is the best I could come up with cancelling the promise. By checking for the abort signal right before the sleep, we fix https://github.com/livekit/agents/blob/a12e89512097dd15cfafed74c04802c8c66026a8/livekit-agents/livekit/agents/voice/audio_recognition.py#L329

Let me know if you think there's a better way to do this @lukasIO

agents/src/voice/audio_recognition.ts
```diff
@@ -56,8 +72,17 @@ export class AudioRecognition {
   }
 
   private async onSTTEvent(ev: SpeechEvent) {
-    // TODO(AJS-30) ignore stt event if user turn already committed and EOU task is done
-    // or it's an interim transcript
+    if (
+      this.manualTurnDetection &&
+      this.userTurnCommitted &&
+      (this.bounceEOUAbortController === undefined ||
+        this.eouTaskDone ||
```
Shubhrakanti: Here we need to detect if the `eouTask` has been started or if it's been aborted. I don't like the fact that we're doing this so implicitly. [[Python source](https://github.com/livekit/agents/blob/a12e89512097dd15cfafed74c04802c8c66026a8/livekit-agents/livekit/agents/voice/audio_recognition.py#L206-L207)]

Python's asyncio's task has a `done()` method, but there is no such equivalent for a `Promise` in JS

agents/src/voice/audio_recognition.ts
```diff
@@ -99,7 +129,73 @@ export class AudioRecognition {
...
+        this.audioTranscript = '';
+      }
+      if (!abortSignal.aborted) {
+        this.eouTaskDone = true;
+      }
+    };
+
+    if (this.bounceEOUAbortController) {
+      this.bounceEOUAbortController.abort();
+    }
```
Shubhrakanti, line 183: This doesn't feel super clean to me, but I couldn't find a better way to do it with our in house [CancellablePromise](https://github.com/livekit/agents-js/blob/be56527c9a4dd86eab3de579b8fd256acb7490e1/agents/src/utils.ts#L153-L221)

agents/src/voice/audio_recognition.ts
Shubhrakanti: Couldn't really figure out how to use [CancellablePromise](https://github.com/livekit/agents-js/blob/be56527c9a4dd86eab3de579b8fd256acb7490e1/agents/src/utils.ts#L153-L221) here.

agents/src/voice/agent_activity.ts
```diff
@@ -73,7 +76,12 @@ export class AgentActivity implements RecognitionHooks {
...
   }
 
-  onEndOfTurn(ev: EndOfTurnInfo): void {
-    this.logger.info('End of turn', ev);
+  async onEndOfTurn(ev: EndOfTurnInfo): Promise<boolean> {
+    this.logger.info(ev, 'End of turn');
+    return true;
+  }
+
+  retrieveChatCtx(): ChatContext {
```
lukasIO, line 84: will this stay sync or should this return a promise for future proofing?

agents/src/voice/agent_session.ts
```diff
@@ -3,13 +3,34 @@
...
+
+const defaultVoiceOptions: VoiceOptions = {
+  allow_interruptions: true,
+  discard_audio_if_uninterruptible: true,
+  min_interruption_duration: 0.5,
+  min_interruption_words: 0,
+  min_endpointing_delay: 0.5,
+  max_endpointing_delay: 6.0,
+  max_tool_steps: 3,
+};
```
lukasIO: if you add `as const` here you get some additional TS benefits for static value checking with those defaults

agents/src/voice/agent_session.ts
```diff
@@ -21,13 +42,17 @@ export class AgentSession {
...
   private roomIO?: RoomIO;
   private logger = log();
-
+  private _chatCtx: ChatContext;
+  private _options: VoiceOptions;
   /** @internal */
   audioInput?: ReadableStream<AudioFrame>;
 
-  constructor(vad: VAD, stt: STT) {
+  constructor(vad: VAD, stt: STT, options: Partial<VoiceOptions> = defaultVoiceOptions) {
```
lukasIO: nit: it's a bit convoluted to set the defaults here and then also spread them a couple of lines under

```suggestion
  constructor(vad: VAD, stt: STT, options?: Partial<VoiceOptions> ) {
```

agents/src/voice/agent_session.ts
```diff
@@ -64,4 +89,13 @@ export class AgentSession {
       await this.activity.start();
     }
   }
+
+  get chatCtx(): ChatContext {
+    // TODO(shubhra): Return a readonly object
+    return this._chatCtx;
+  }
+
+  get options(): VoiceOptions {
```
lukasIO: should this also be readonly or do we allow mutating these ?

agents/src/voice/audio_recognition.ts
```diff
@@ -21,26 +22,41 @@ export interface RecognitionHooks {
   onVADInferenceDone: (ev: VADEvent) => void;
   onInterimTranscript: (ev: SpeechEvent) => void;
   onFinalTranscript: (ev: SpeechEvent) => void;
-  onEndOfTurn: (info: EndOfTurnInfo) => void;
+  onEndOfTurn: (info: EndOfTurnInfo) => Promise<boolean>;
+
+  retrieveChatCtx: () => ChatContext;
```
lukasIO, line 29: same async question as above

agents/src/voice/audio_recognition.ts
lukasIO: I think the `Future` helper has a `done` property

agents/src/voice/audio_recognition.ts
```diff
@@ -67,30 +92,35 @@ export class AudioRecognition {
 
         if (!transcript) return;
 
-        this.logger.debug('received user transcript', {
-          user_transcript: transcript,
-          language: this.lastLanguage,
-        });
+        this.logger.debug(
+          {
```
lukasIO, line 98: are we changing the logger signature to get context first and then the message? 👀

agents/src/voice/audio_recognition.ts
```diff
@@ -99,7 +129,73 @@ export class AudioRecognition {
...
+            endpointingDelay = this.maxEndpointingDelay;
+          }
+        }
+      }
+
+      const extraSleep = lastSpeakingTime + endpointingDelay - Date.now();
+      if (abortSignal.aborted) {
+        return;
+      }
+      await new Promise((resolve) => setTimeout(resolve, Math.max(extraSleep, 0)));
```
lukasIO: there should be a `sleep` helper function already, if there's none then it's probably worth adding it

agents/src/voice/audio_recognition.ts
lukasIO: what would you like to improve in particular?

agents/src/voice/audio_recognition.ts
lukasIO: additionally to checking for it manually, we'd want to `addEventListener('abort'` to ensure it can be cancelled at any time. 

I'd also be inclined to reimplement our current cancealble promise with an abort controller based one. that would probably feel cleaner.

lukasIO: we can also look at deno's std libs once again for this: https://jsr.io/@std/async/doc/~/abortable

agents/src/voice/agent_activity.ts
Shubhrakanti: It's [sync in python](https://github.com/livekit/agents/blob/ca575896820ac56d3434a84004c7897b3887a02e/livekit-agents/livekit/agents/voice/agent_activity.py#L993-L994)

Although the naming is quite poor, should be `getChatCtx` or something. Retrieve implies like a network imo.

agents/src/voice/agent_session.ts
Shubhrakanti: yeah this is better.

agents/src/voice/agent_session.ts
Shubhrakanti: In the [python code it is mutable](https://github.com/livekit/agents/blob/ca575896820ac56d3434a84004c7897b3887a02e/livekit-agents/livekit/agents/voice/agent_session.py#L271-L273).

(`@propety` in python returns a reference - see https://stackoverflow.com/questions/51114984/how-to-manage-access-to-a-mutable-attribute-in-python)

Edit: I think technically `@propety` is intended to be readonly. So I think we should make this read-only?

cc: @theomonnom

Shubhrakanti: @theomonnom is this intentional?

agents/src/voice/agent_activity.ts
Shubhrakanti: @theomonnom let me know what you think.

agents/src/voice/audio_recognition.ts
Shubhrakanti: yeah looks like it's sync on the python code. Might be worth renaming tho.

agents/src/voice/audio_recognition.ts
Shubhrakanti: lol I was using the logger wrong the whole time. Pino takes in the object first then message (unless I'm understating the docs wrong). I find this really confusing.

https://getpino.io/#/docs/api?id=logger-instance

> Each logging method has the following signature: ([mergingObject], [message], [...interpolationValues]).

agents/src/voice/audio_recognition.ts
Shubhrakanti: Yeah I think I can clean it up so this check just depends on one `Future` object which represents the task state.

agents/src/voice/audio_recognition.ts
Shubhrakanti: I guess I don't like the fact that the `bounceEOUAbortController`, `bounceEOUTask`, and `eouTaskDone: Future` are so tightly coupled implicitly. I guess this is kind of inevitable because we don't have the `done` method of an asynchronous function which python gives us.

agents/src/voice/audio_recognition.ts
lukasIO: what's your take on the `abortable` I mentioned [here](https://github.com/livekit/agents-js/pull/394#discussion_r2097913400)

agents/src/voice/audio_recognition.ts
lukasIO: maybe we can extend our Future to be [abortable](https://jsr.io/@std/async/doc/~/abortable) - that would still be tightly coupled, but you would only have the single abortable future to worry about

agents/src/voice/audio_recognition.ts
Shubhrakanti: oh I missed this comment. Looking into it now.

agents/src/voice/agent_session.ts
Shubhrakanti: made it readonly for now.

agents/src/voice/audio_recognition.ts
Shubhrakanti: Not sure if that would help simply for this specific case because Deno's `abortable` uses a [`Promise.race`](https://github.com/denoland/std/blob/9195573293bffe4c5ced44fa5bda3dd6db488c0d/async/abortable.ts#L134) which doesn't actually halt execution of the function. 

See this [slack thread](https://live-kit.slack.com/archives/C07TWVC0W4A/p1747941121372549?thread_ts=1747936302.834019&cid=C07TWVC0W4A) for context.

Shubhrakanti: But I used Deno's delay method instead of our sleep method. And I do see a use case for using `abortable` for iterators.

agents/src/voice/audio_recognition.ts
```diff
@@ -99,7 +131,72 @@ export class AudioRecognition {
...
+          const endOfTurnProbability = await turnDetector.predictEndOfTurn(chatCtx);
+          const unlikelyThreshold = turnDetector.unlikelyThreshold(this.lastLanguage);
+          if (unlikelyThreshold && endOfTurnProbability < unlikelyThreshold) {
+            endpointingDelay = this.maxEndpointingDelay;
+          }
+        }
+      }
+
+      const extraSleep = lastSpeakingTime + endpointingDelay - Date.now();
+      if (abortSignal.aborted) {
```
Shubhrakanti: @theomonnom

agents/src/voice/audio_recognition.ts
```diff
@@ -99,7 +131,79 @@ export class AudioRecognition {
...
+          const endOfTurnProbability = await turnDetector.predictEndOfTurn(chatCtx);
+          const unlikelyThreshold = turnDetector.unlikelyThreshold(this.lastLanguage);
+          if (unlikelyThreshold && endOfTurnProbability < unlikelyThreshold) {
+            endpointingDelay = this.maxEndpointingDelay;
+          }
+        }
+      }
+
+      const extraSleep = lastSpeakingTime + endpointingDelay - Date.now();
+      await delay(extraSleep, { signal: abortSignal });
```
Shubhrakanti, line 163: This is the best solution for this use case IMO. Can't really use `ts-debounce` since the debounce delay is dynamically calculated based on the `endOfTurnProbability`

PR #395, opened 2025-05-21
Generate reply audio after turn detected

Shubhrakanti:


agents/src/voice/room_io.ts
```diff
@@ -58,16 +62,17 @@ export class RoomIO {
...
     this.publication = await this.room.localParticipant?.publishTrack(
       track,
       new TrackPublishOptions({ source: TrackSource.SOURCE_MICROPHONE }),
     );
   }
 
   start() {
-    this.publishTrack();
+    this.audioSource = new AudioSource(this.sampleRate, this.numChannels);
+    this.publishTrack(this.audioSource);
```
Shubhrakanti, line 75: set up should be done on start method. This is a follow up for #396

agents/src/voice/io.ts
```diff
@@ -3,9 +3,21 @@
...
 import type { SpeechEvent } from '../stt/stt.js';
 
 export type STTNode = (
   audio: ReadableStream<AudioFrame>,
   modelSettings: any, // TODO(shubhra): add type
 ) => Promise<ReadableStream<SpeechEvent | string> | null>;
+
+export type LLMNode = (
+  chatCtx: ChatContext,
+  modelSettings: any, // TODO(shubhra): add type
```
Shubhrakanti: add a ticket for this

agents/src/voice/generation.ts
```diff
@@ -0,0 +1,156 @@
...
+  const out = new _TextOutput('', new Future());
+  return [forwardText(textOutput, source, out), out];
+}
+
+export class _AudioOutput {
+  constructor(
+    public readonly audio: Array<AudioFrame>,
+    public readonly firstFrameFut: Future,
+  ) {}
+}
```
Shubhrakanti: I think this is better as an interface or type. Less LOC.

agents/src/voice/generation.ts
```diff
@@ -0,0 +1,156 @@
...
+          if (!content) continue;
+          data.generatedText += content;
+          writer.write(content);
+        } else {
+          //TODO(shubhra): get access to logger instance
+          console.error('Unexpected chunk type:', chunk);
+        }
+      }
+    } finally {
+      console.log('+++++++++++++ Closing writer in performLLMInference');
```
Shubhrakanti: remove

agents/src/voice/generation.ts
```diff
@@ -0,0 +1,156 @@
...
+          writer.write(chunk);
+          // TODO(shubhra): better way to check??
+        } else if ('choices' in chunk) {
+          const content = chunk.choices[0]?.delta.content;
+          if (!content) continue;
+          data.generatedText += content;
+          writer.write(content);
+        } else {
+          //TODO(shubhra): get access to logger instance
+          console.error('Unexpected chunk type:', chunk);
```
Shubhrakanti: Right now we initiate the logger within a class ex. `private logger = log()`. Is there an issue with doing this in the beginning of a file so we don't have to pass it in to functions? This is what we do in python. Let me know what you think @lukasIO

agents/src/tts/tts.ts
```diff
@@ -119,8 +122,33 @@ export abstract class SynthesizeStream
   #metricsText = '';
   #monitorMetricsTask?: Promise<void>;
 
+  private deferredInputStream: DeferredReadableStream<
+    string | typeof SynthesizeStream.FLUSH_SENTINEL
```
Shubhrakanti, line 126: Should make a ticket to see if we need the `FLUSH_SENTINEL`s anymore now that we have `stream.close()`

agents/src/tts/tts.ts
```diff
@@ -119,8 +122,33 @@ export abstract class SynthesizeStream
...
+  >;
+  private logger = log();
+
   constructor(tts: TTS) {
     this.#tts = tts;
+    this.deferredInputStream = new DeferredReadableStream();
+    this.mainTask();
+  }
+
+  protected async mainTask() {
```
Shubhrakanti, line 137: Add todo to remove later

agents/src/voice/agent.ts
```diff
@@ -11,13 +11,16 @@ import type { ChatChunk, ChatMessage, LLM } from '../llm/index.js';
 import { ChatContext } from '../llm/index.js';
 import { StreamAdapter as STTStreamAdapter } from '../stt/index.js';
 import type { STT, SpeechEvent } from '../stt/index.js';
+import { SentenceTokenizer as BasicSentenceTokenizer } from '../tokenize/basic/index.js';
 import type { TTS } from '../tts/index.js';
+import { SynthesizeStream, StreamAdapter as TTSStreamAdapter } from '../tts/index.js';
 import type { VAD } from '../vad.js';
 import type { AgentActivity } from './agent_activity.js';
 
+export class StopResponse extends Error {}
```
Shubhrakanti: @lukasIO is this proper or is there a better way to do this?

agents/src/voice/agent.ts
```diff
@@ -125,5 +137,59 @@ export class Agent {
...
+      modelSettings: any, // TODO(shubhra): add type
+    ): Promise<ReadableStream<ChatChunk | string> | null> {
+      const activity = agent.getActivityOrThrow();
+      const stream = activity.llm.chat({ chatCtx });
+      return new ReadableStream({
+        async start(controller) {
+          for await (const chunk of stream) {
+            controller.enqueue(chunk);
+          }
+          console.log('+++++++++++++ LLM stream ended in llmNode');
```
Shubhrakanti: Remove

agents/src/voice/agent_activity.ts
```diff
@@ -76,12 +106,182 @@ export class AgentActivity implements RecognitionHooks {
...
+      0,
+      this.currentSpeech,
+    );
+
+    if (instructions) {
+      instructions = `${this.agent.instructions}\n${instructions}`;
+    }
+
+    this.pipelineReplyTask(handle, chatCtx || this.agent.chatCtx, userMessage?.copy()); // add instructions
+    // this.scheduleSpeech(handle, SpeechHandle.SPEECH_PRIORITY_NORMAL);
```
Shubhrakanti: Remove the schedule speech for now. There should be a separate ticket to handle this.

agents/src/voice/generation.ts
lukasIO: there was a reason why were doing it this way, but I don't remember the details

agents/src/voice/agent.ts
lukasIO: we should probably give custom errors a custom `name` to make it distinguishable from general errors

agents/src/tts/tts.ts
Shubhrakanti: yeah we will need this. https://github.com/livekit/agents/blob/124cf26beadfeafe4c6c948bab72e25250071b92/livekit-agents/livekit/agents/tts/tts.py#L280-L281

agents/src/voice/generation.ts
Shubhrakanti: Oh it's because you have to call `initializeLogger` somehow before all the imports run. https://github.com/livekit/agents-js/blob/236d2c62cc28d9a5c19e73e1b98709f243b91a77/agents/src/log.ts#L22

mildly annoying, we can just throw an error here for now.

agents/src/pipeline/speech_handle.ts
```diff
@@ -41,6 +44,20 @@ export class SpeechHandle {
     this.#userQuestion = userQuestion;
     this.#fncNestedDepth = fncNestedDepth;
     this.#fncExtraToolsMesages = extraToolsMessages;
+    this.parent = parent;
+  }
+
+  static create(allowInterruptions: boolean = false, stepIndex: number = 0, parent?: SpeechHandle) {
```
toubatbrian, line 50: for factory method, would be nicer if we make the arguments "orderless", i.e. `create(opts: SpeechHandleOptions)` where SpeechHandleOptions defines an interface.

agents/src/voice/agent.ts
```diff
@@ -27,6 +35,8 @@ export class Agent {
 
   /** @internal */
   agentActivity?: AgentActivity;
```
toubatbrian, line 37: `_agentActivity`

toubatbrian, line 37: Should this be private?

toubatbrian (commented):
added some comments

agents/src/voice/agent.ts
Shubhrakanti: We use It inside the agents code base, but don't want users using this. Decided that the internal flag was the best. I have a ticket to make all the names start with `_` if they are internal.

agents/src/pipeline/speech_handle.ts
Shubhrakanti: yeah I'm going to do this for all the class constructors too. going to fix `speech_handle` later since it's going to get major changes 1.0 - but good call.

agents/src/voice/audio_recognition.ts
```diff
@@ -160,7 +160,7 @@ export class AudioRecognition {
       }
 
       const extraSleep = lastSpeakingTime + endpointingDelay - Date.now();
-      await delay(extraSleep, { signal: abortSignal });
+      await delay(Math.max(extraSleep, 0), { signal: abortSignal });
```
Shubhrakanti, line 163: small bug fix

PR #397, opened 2025-05-21
Add agent/user state 

Shubhrakanti:
Added boiler plate for agent state and user state.


agents/src/voice/events.ts
lukasIO, line 1: was this supposed to be called `types.ts`?

Shubhrakanti: No we have an events.py file where keep and the events emitted by the Agent session. Keeping the name the same.

PR #401, opened 2025-05-28
Keep job process and inference process alive even if worker shuts down

Shubhrakanti:


theomonnom:
We actually do

theomonnom:
Python is exiting here: https://github.com/livekit/agents/blob/945f56358d9da1e334e8382a840d3f1b2a54a104/livekit-agents/livekit/agents/ipc/proc_client.py#L153

theomonnom:
If something goes wrong in the worker, we're already in a pretty bad state, it will most likely always be the opposite. 
orphan processes are generally a bad practice

PR #402, opened 2025-05-29
Shubhra/ajs 37 refactor tts with streams

Shubhrakanti:
Similar to #390

### DIFF
--- a/agents/src/tts/tts.ts
+++ b/agents/src/tts/tts.ts
@@ -4,7 +4,11 @@
 import type { AudioFrame } from '@livekit/rtc-node';
 import type { TypedEventEmitter as TypedEmitter } from '@livekit/typed-emitter';
 import { EventEmitter } from 'node:events';
+import type { ReadableStream } from 'node:stream/web';
+import { log } from '../log.js';
 import type { TTSMetrics } from '../metrics/base.js';
+import { DeferredReadableStream } from '../stream/deferred_stream.js';
+import { IdentityTransform } from '../stream/identity_transform.js';
 import { AsyncIterableQueue, mergeFrames } from '../utils.js';
 
 /** SynthesizedAudio is a packet of speech synthesis as returned by the TTS. */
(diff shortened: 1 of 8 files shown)


agents/src/vad.ts
```diff
@@ -84,21 +84,22 @@ export abstract class VAD extends (EventEmitter as new () => TypedEmitter<VADCal
...
   #vad: VAD;
   #lastActivityTime = BigInt(0);
   private logger = log();
   private deferredInputStream: DeferredReadableStream<AudioFrame>;
-
+  private input = new IdentityTransform<AudioFrame | typeof VADStream.FLUSH_SENTINEL>();
+  private output = new IdentityTransform<VADEvent>();
   private metricsStream: ReadableStream<VADEvent>;
+  private outputReader: ReadableStreamDefaultReader<VADEvent>;
+  private inputWriter: WritableStreamDefaultWriter<AudioFrame | typeof VADStream.FLUSH_SENTINEL>;
```
Shubhrakanti, line 101: Making access modifiers more restrictive . Forgot to do it in #390

agents/src/vad.ts
```diff
@@ -220,7 +221,9 @@ export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
   }
 
   close() {
-    this.input.writable.close();
+    if (!this.inputClosed) {
+      this.inputWriter.close();
+    }
```
Shubhrakanti, line 226: forgot to do in #390

agents/src/tts/tts.ts
```diff
@@ -105,22 +109,73 @@ export abstract class SynthesizeStream
...
+  private input = new IdentityTransform<string | typeof SynthesizeStream.FLUSH_SENTINEL>();
+  private output = new IdentityTransform<
+    SynthesizedAudio | typeof SynthesizeStream.END_OF_STREAM
+  >();
+  private inputWriter: WritableStreamDefaultWriter<string | typeof SynthesizeStream.FLUSH_SENTINEL>;
+  private outputReader: ReadableStreamDefaultReader<
+    SynthesizedAudio | typeof SynthesizeStream.END_OF_STREAM
+  >;
+  private logger = log();
+  private inputClosed = false;
```
Shubhrakanti, line 137: This is duplicative. The Readable/WritableStreamDefaultWriter internally tracks whether it's closed but doesn't expose it. The only way to know is when you try to write to or close an already-closed writer, it throws an error.

The other option would be to 
```typescript
    try {
      this.inputWriter.write(SynthesizeStream.FLUSH_SENTINEL);
    } catch (error) {
      throw new Error('Input is closed');
    }
```
everywhere, which doesn't seem much better. Let me know what you thoughts are @lukasIO @toubatbrian

PR #408, opened 2025-05-30
wip injectable stream

lukasIO:


agents/src/stream/injectable_stream.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  async cancel(reason?: any) {
+    await Promise.all([
+      this.mergedStream.cancel(reason),
+      this.identityStream.writable.abort(reason),
+    ]);
+  }
+
+  get readable() {
```
toubatbrian, line 60: Do we also consider supporting "eject" (read single value from stream) operation? Basically making stream also have a queue-like API with `put` as `inject` and `get` as `eject`

lukasIO: Sure if there's a concrete use case? 

I'd be in favor of implementing additional features (generally) only once we know we have a defined use case for it.

agents/src/stream/injectable_stream.ts
```diff
@@ -0,0 +1,48 @@
...
+    try {
+      const writer = this.identityStream.writable.getWriter();
+      await writer.write(value);
+      await writer.close();
+    } finally {
+      unlock();
+    }
+  }
+
+  async close() {
```
Shubhrakanti, line 35: @toubatbrian might be worth taking over this PR and working on the releasing the lock on the readable source once we close the stream. 

Some tests would be good to.

agents/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../tsconfig.json",
 
   "compilerOptions": {
+    "types": ["node"],
```
lukasIO, line 5: instead of defining it here, I think it would make sense to define it in the root package.json so that all packages (including plugins) get it by default

PR #409, opened 2025-05-30
Shubhra/ajs 58 add speech scheduling in Agent Activity

Shubhrakanti:
Add basic speech Q


Shubhrakanti:
Not sure why we still have ```(node:42348) TimeoutNegativeWarning: -4837 is a negative number.
Timeout duration was set to 1.
    at new Timeout (node:internal/timers:195:17)
    at setTimeout (node:timers:163:19)
    at AudioSource.captureFrame (file:///Users/shubhrakantiganguly/development/agents-js/node_modules/.pnpm/@livekit+rtc-node@0.13.13/node_modules/@livekit/rtc-node/dist/audio_source.js:86:20)
    at forwardAudio (file:///Users/shubhrakantiganguly/development/agents-js/agents/dist/voice/generation.js:93:22)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Promise.all (index 3)
    at async AgentActivity.pipelineReplyTask (file:///Users/shubhrakantiganguly/development/agents-js/agents/dist/voice/agent_activity.js:216:5)```

agents/src/utils.ts
```diff
@@ -339,3 +339,50 @@ export class AudioEnergyFilter {
...
+      this.promise = new Promise<void>((resolve) => {
+        this.resolve = resolve;
+      });
+    }
+  }
+
+  get isSet(): boolean {
+    return this._isSet;
+  }
+}
```
Shubhrakanti: Might be better to just use an event emitter? Something like 

```typescript
import { EventEmitter, once } from 'node:events';

export class AgentActivity implements RecognitionHooks {
  private queueEventEmitter = new EventEmitter();
  
  private wakeUpMainTask(): void {
    this.queueEventEmitter.emit('queue-updated');
  }

  private async mainTask(): Promise<void> {
    while (true) {
      await once(this.queueEventEmitter, 'queue-updated');
      
      // Process speech queue...
      
      if (this.draining && this.speechTasks.length === 0) {
        break;
      }
    }
  }
}
```

agents/src/pipeline/pipeline_agent.ts
```diff
@@ -778,8 +778,6 @@ export class VoicePipelineAgent extends (EventEmitter as new () => TypedEmitter<
           speechId: handle.id,
         })
         .debug('committed agent speech');
-
-      handle.setDone();
```
Shubhrakanti, line 782: We never actually used this in the legacy implementation. Nothing was "waiting" for the handle to be done.

agents/src/voice/agent_activity.ts
```diff
@@ -32,12 +34,16 @@ export class AgentActivity implements RecognitionHooks {
   private turnDetectionMode?: string;
   private _draining = false;
   private currentSpeech?: SpeechHandle;
+  private speechQueue: Heap<[number, number, SpeechHandle]>;
+  private queueEvents: EventEmitter;
```
Shubhrakanti: I like this better.

Shubhrakanti:
@toubatbrian @lukasIO let me know which implementation you prefer for the `q_updated` event. I swapped it to the `EventEmitter` logic in this commit https://github.com/livekit/agents-js/pull/409/commits/092fff5763fed8e2519c81816409ebad133dce17.

agents/package.json
```diff
@@ -43,6 +43,7 @@
     "@livekit/typed-emitter": "^3.0.0",
     "@std/async": "npm:@jsr/std__async@^1.0.13",
     "commander": "^12.0.0",
+    "heap-js": "^2.6.0",
```
Shubhrakanti, line 46: No strong preference on this library. Happy to use a different one if you guys think we should.

Shubhrakanti: https://github.com/ignlg/heap-js#readme

agents/src/voice/agent_activity.ts
Shubhrakanti: Hmm maybe we can just use a `Future`

agents/src/voice/agent_activity.ts
```diff
@@ -132,6 +142,28 @@ export class AgentActivity implements RecognitionHooks {
...
+        const heapItem = this.speechQueue.pop();
+        if (!heapItem) {
+          throw new Error('Speech queue is empty');
+        }
+        const speechHandle = heapItem[2];
+        this.currentSpeech = speechHandle;
+        speechHandle.authorizePlayout();
+        await speechHandle.waitForPlayout();
+        this.currentSpeech = undefined;
+        this.q_updated = new Future();
```
Shubhrakanti, line 157: is this hacky? In python `Event` has a `clear` function

toubatbrian: I made a Event primitive ([PR](https://github.com/livekit/agents-js/pull/413)) similar to how `asyncio.Event` in python does, maybe we can consider using that to keep the functionality aligned.

lukasIO (approved):

agents/src/voice/agent_activity.ts
Shubhrakanti: discussed offline. `new Future` is fine for this use case.

PR #412, opened 2025-05-31
[WIP] Add Manual Turn Detection

toubatbrian:


agents/src/utils.ts
```diff
@@ -362,24 +389,60 @@ export class AbortableTask<T> {
...
   }
 
+  /**
+   * Cancels the task and waits for it to complete.
+   *
+   * @param timeout - The timeout in milliseconds
+   * @returns The result of the task
+   * @throws {@link TASK_TIMEOUT_ERROR} if the task times out
+   */
+  async cancelAndWait(timeout: number = 2000) {
```
toubatbrian: @lukasIO Do you see any room of improvements for this function?

agents/src/stream/deferred_stream.ts
```diff
@@ -24,4 +24,9 @@ export class DeferredReadableStream<T> {
     }
     source.pipeTo(this.transform.writable);
   }
+
+  cancel() {
```
lukasIO: closing and cancelling differ in intent. 

is the intention to `cancel` or to `close` here ?

agents/src/utils.ts
```diff
@@ -362,24 +389,59 @@ export class AbortableTask<T> {
...
+   */
   cancel() {
     this.controller.abort();
   }
 
+  /**
+   * Cancels the task and waits for it to complete.
+   *
+   * @param timeout - The timeout in milliseconds
+   * @returns The result of the task
```
lukasIO: that sounds misleading, it's only returning `TaskResult` enum, not any return value / result of the task, right?

agents/src/utils.ts
```diff
@@ -362,24 +389,59 @@ export class AbortableTask<T> {
...
+   * @throws {@link TASK_TIMEOUT_ERROR} if the task times out
```
lukasIO: If you want to be explicit about errors being thrown, then keep in mind that you're currently throwing _all_ errors that happen in `await result` additionally to the timeout error

agents/src/utils.ts
```diff
@@ -362,24 +389,59 @@ export class AbortableTask<T> {
...
+  async cancelAndWait(timeout: number = 2000) {
+    this.cancel();
+
+    try {
+      // Race between task completion and timeout
+      const result = await Promise.race([
+        this.result
+          .then(() => TaskResult.Completed)
+          .catch((error) => {
+            if (error.name === 'AbortError') {
```
lukasIO: I believe this is only true if `controller.abort` has been called without any reason given. 

to follow up questions
- is this exclusively an internal utility that we have usage control over?
- can we wrap it so that it's possible to prevent reason being passed?

agents/src/utils.ts
toubatbrian: I see. Will change the spec, thanks for spotting it!

agents/src/stream/deferred_stream.ts
toubatbrian: This method wasn't intended to be included in this PR. I will delete this and make another PR for closing logic.

Also, could you briefly mention the difference between closing and cancelling? @lukasIO

agents/src/utils.ts
toubatbrian: > is this exclusively an internal utility that we have usage control over?

Yes

> can we wrap it so that it's possible to prevent reason being passed?

Generally, passing reason message from outside of task is impossible, since we abstract it under `.cancel()` function, inside of which it calls `this.controller.abort();`. However, it is possible for internal developer, though not recommended, to call `controller.abort()` with possibly non-empty reason message inside the task since controller is exposed as a function parameter.

I was thinking wrapping the controller parameter to be a function like
```ts
const shouldAbort = () => this.controller.signal.aborted

// and pass shouldAbort as function parameter
const task = Task.from((shouldAbort) => { .... });
```

On a second thought, I realize sometimes we might need explicit reference to `controller` or `controller.signal` object inside function, like `await delay(100, { signal: controller.signal });`

So it becomes quite tricky to create a wrapper that hide the reason message. 

I would assume this function is only used for internal codebase and as long as we do not call `controller.abort()` with a given reason inside task, we should be fine. What do you think?

lukasIO: yeah sounds good to me! thanks for thinking through the possibilities!

lukasIO: coming back to this, I think it would already help if we only passed the `signal` to the abortable task's function. 
I don't see any use case for the whole controller to be required, the signal makes the intention also clear that you're supposed to handle aborts within the function, but not triggering aborts yourself

agents/src/stream/deferred_stream.ts
lukasIO: This is mostly about semantic meaning, but `close` implies a "successful" closing while `cancel` carries a notion of  "aborting" as well.

agents/src/utils.ts
Shubhrakanti: Yeah I agree with this. We should only be passing in the signal.

.eslintrc
```diff
@@ -22,6 +22,6 @@
     "@typescript-eslint/ban-ts-comment": "warn",
     "@typescript-eslint/no-empty-interface": "warn",
     "@typescript-eslint/consistent-type-imports": "warn",
-    "@typescript-eslint/no-explicit-any": "warn",
-  },
+    "@typescript-eslint/no-explicit-any": "off"
+  }
```
Shubhrakanti: Let's of this inline

agents/src/stream/deferred_stream.ts
```diff
@@ -6,27 +6,9 @@ import type {
   ReadableStreamDefaultReader,
   WritableStreamDefaultWriter,
 } from 'node:stream/web';
+import { isStreamReaderReleaseError } from '../utils.js';
```
Shubhrakanti: let's keep it in this file? It's only relevant to `DefferedStream`

agents/src/stream/identity_transform.test.ts
```diff
@@ -4,7 +4,7 @@
 import { describe, expect, it } from 'vitest';
 import { IdentityTransform } from './identity_transform.js';
 
-describe('IdentityTransform', () => {
+describe('IdentityTransform', { timeout: 2000 }, () => {
```
Shubhrakanti: timeouts are already set globally - is there a reason for making this different?

agents/src/stream/merge_readable_streams.ts
```diff
@@ -0,0 +1,48 @@
...
+  let resolve!: (value: T | PromiseLike<T>) => void;
+  let reject!: (reason?: any) => void;
+  
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+
+  return { promise, resolve, reject };
+}
```
Shubhrakanti: Super useful - let's move this to `utils`? Also does node not have something like this/can we just directly import from `deno`?

Shubhrakanti: I bet we can refactor `Future` with this to make it cleaner.

Shubhrakanti (commented):
Great progress! We got some neat primitives from diving into this task.

Just snooped around and left some thoughts. Let me know if you need any help getting this across the finish line!

.eslintrc
```diff
@@ -22,6 +22,6 @@
     "@typescript-eslint/ban-ts-comment": "warn",
     "@typescript-eslint/no-empty-interface": "warn",
     "@typescript-eslint/consistent-type-imports": "warn",
-    "@typescript-eslint/no-explicit-any": "warn",
-  },
+    "@typescript-eslint/no-explicit-any": "warn"
+  }
```
Shubhrakanti: lint? This should be a no-op

agents/src/stream/deferred_stream.ts
```diff
@@ -14,7 +14,7 @@ import { IdentityTransform } from './identity_transform.js';
  * Invalid state: Releasing reader
  * Invalid state: The reader is not attached to a stream
  */
-function isStreamReaderReleaseError(e: unknown) {
+export function isStreamReaderReleaseError(e: unknown) {
```
Shubhrakanti, line 17: do we use this outside `deferred_stream.ts`? If not then no need for the export

agents/src/stream/identity_transform.test.ts
Shubhrakanti: bump?

agents/src/stream/merge_readable_streams.ts
```diff
@@ -0,0 +1,40 @@
+// SPDX-FileCopyrightText: 2025 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { ReadableStream } from 'node:stream/web';
+import { withResolvers } from '../utils.js';
+
+// Adapted from https://github.com/denoland/std/blob/main/streams/merge_readable_streams.ts
+// we manually adapted to make ReadableStream<T> typing compatible with our current node
+// version as well as typescript configuration
+export function mergeReadableStreams<T>(...streams: ReadableStream<T>[]): ReadableStream<T> {
```
Shubhrakanti, line 10: We're not actually using this in this PR? Generally we should try to keep the scope of PRs as small as possible.

agents/src/utils.ts
```diff
@@ -462,3 +462,15 @@ export class Task<T> {
     return this.resultFuture.done;
   }
 }
+
+export function withResolvers<T = unknown>() {
```
Shubhrakanti, line 468: Same we're not using it in this PR. We should add this in a separate PR with the actual use case

agents/src/voice/agent.ts
```diff
@@ -118,6 +119,7 @@ export class Agent {
       audio: ReadableStream<AudioFrame>,
       modelSettings: any, // TODO(AJS-59): add type
     ): Promise<ReadableStream<SpeechEvent | string> | null> {
+      const logger = log();
```
Shubhrakanti: make this a class attribute with a default constructor `private logger = log()`

agents/src/voice/agent.ts
```diff
@@ -131,17 +133,26 @@ export class Agent {
...
+            logger.debug('Agent.default.sttNode: enqueued event');
           }
+          logger.debug('Agent.default.sttNode: closing controller');
           controller.close();
+          logger.debug('Agent.default.sttNode: controller closed');
         },
         cancel() {
+          logger.debug('Agent.default.sttNode: detaching deferred input stream');
+          stream.detachInputStream();
+          logger.debug('Agent.default.sttNode: closing stt readable stream');
```
Shubhrakanti: This feels really noisy. Do we really need all these logs?

agents/src/voice/agent_activity.ts
```diff
@@ -92,10 +92,33 @@ export class AgentActivity implements RecognitionHooks {
...
   updateAudioInput(audioStream: ReadableStream<AudioFrame>): void {
     this.audioRecognition?.setInputAudioStream(audioStream);
   }
 
+  commitUserTurn() {
+    if (!this.audioRecognition) {
+      throw new Error('AudioRecognition is not initialized');
+    }
+
+    // TODO(brian): add audio_detached flag
```
Shubhrakanti, line 109: Let's make a ticket and add `TODO(AJS-#)`

agents/src/voice/agent_activity.ts
```diff
@@ -106,6 +129,12 @@ export class AgentActivity implements RecognitionHooks {
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   onVADInferenceDone(ev: VADEvent): void {
+    // skip speech handle interruption for manual and realtime model
+    if (this.turnDetection === 'manual' || this.turnDetection === 'realtime_llm') {
+      return;
+    }
+
+    // TODO(AJS-40): Implement this
```
Shubhrakanti: shouldn't be here.

agents/src/voice/agent_activity.ts
```diff
@@ -131,16 +160,18 @@ export class AgentActivity implements RecognitionHooks {
       // TODO(shubhra): should we "forward" this new turn to the next agent/activity?
       return true;
     }
+
     if (
       this.stt &&
-      this.turnDetectionMode !== 'manual' &&
+      this.turnDetection !== 'manual' &&
```
Shubhrakanti, line 165: is there a reason for the rename? I think`turnDetectionMode` is clearer?

examples/src/push_to_talk.ts
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    const endTurn = (callerIdentity: string) => {
+      // session.input.setAudioEnabled(false);
+      session.commitUserTurn();
+      console.log('end turn', callerIdentity);
+    };
+
+    const cancelTurn = (callerIdentity: string) => {
+      // session.input.setAudioEnabled(false);
```
Shubhrakanti, line 58: Can we get rid of the commented code?

package.json
```diff
@@ -46,5 +47,5 @@
     "vitest": "^3.2.2",
     "tsup": "^8.4.0"
   },
-  "packageManager": "pnpm@9.7.0"
+  "packageManager": "pnpm@10.11.0"
```
Shubhrakanti: Just curious -is there a reason for this?

agents/src/voice/audio_recognition.ts
```diff
@@ -38,52 +42,98 @@ interface _TurnDetector {
...
+    const secondAgo = (time: number) => (Date.now() - time) / 1000;
+
+    this.logger.debug('===============================================');
+    const debugObject = {
+      interimTranscript: this.audioInterimTranscript,
+      audioTranscript: this.audioTranscript,
+      lastFinalTranscriptTime: `${secondAgo(this.lastFinalTranscriptTime)}s ago`,
+      userTurnCommitted: this.userTurnCommitted,
+    };
+    this.logger.debug(debugObject, 'debugObject');
```
Shubhrakanti: Remove this. We should only use stuff like this for immediate debugging.

agents/src/voice/audio_recognition.ts
```diff
@@ -38,52 +42,98 @@ interface _TurnDetector {
...
     });
+
+    // every 1 second, print interm transcript and audio transcript and lastFinalTranscriptTime and userTurnCommitted
+    // (async () => {
+    //   while (true) {
+    //     this.printDebugObject();
+    //     await delay(500);
+    //   }
+    // })();
+  }
```
Shubhrakanti: Clean up

agents/src/voice/audio_recognition.ts
```diff
@@ -177,81 +252,214 @@ export class AudioRecognition {
...
+        return;
       }
-    }
+
+      this.logger.debug('createSttTask: create stt stream from stt node');
+      const sttStream = await this.stt(this.sttInputStream, {});
+
+      if (controller.signal.aborted) return;
+      if (sttStream === null) return;
+
```
Shubhrakanti: use `| |`

agents/src/voice/audio_recognition.ts
```diff
@@ -177,81 +252,214 @@ export class AudioRecognition {
...
+          while (true) {
+            if (controller.signal.aborted) {
+              this.logger.debug('createSttTask: abort signal detected in read loop, breaking...');
+              break;
+            }
+
+            const { done, value: ev } = await reader.read();
+
+            if (done) {
+              this.logger.debug('createSttTask: sttStream reader done, exiting loop...');
```
Shubhrakanti: Clean up these logs

Shubhrakanti (changes requested):
Nice! 

-  Would be nice to clean up all your debugging logs before flagging for review. Makes the PR very hard to review otherwise
- Only add abstractions when you actually use them. If they are quite complex use [stacked PRs](https://www.stacking.dev) to keep individual PRs small.

agents/src/stream/deferred_stream.ts
toubatbrian: Yes, inside audio_recogniztion

agents/src/stream/merge_readable_streams.ts
toubatbrian: We do use it inside `audio_recognition`: `setInputAudioStream`

agents/src/utils.ts
toubatbrian: This one been used in `mergeReadableStreams`

package.json
toubatbrian: reverted

agents/src/voice/agent_activity.ts
toubatbrian: https://linear.app/livekit/project/js-10-ec0e4d61e59e/issues

examples/src/push_to_talk.ts
toubatbrian: commented code is intended to be here in python version. Once those features are implemented, will add them back. Keeping them commented for now to track progress.

agents/src/stream/merge_readable_streams.ts
lukasIO: I still think this won't be necessary as soon as you include the `types: ['node']` in the root tsconfig 👀

examples/src/push_to_talk.ts
Shubhrakanti: Can you add a comment about this on top 
`// TODO() - commented code is intended to be here in python version. Once those features are implemented, will add them back` or something like that?

agents/src/stream/deferred_stream.test.ts
```diff
@@ -489,9 +489,10 @@ describe('DeferredReadableStream', { timeout: 2000 }, () => {
     try {
       await readPromise;
       expect.fail('readPromise should have rejected');
-    } catch (e: any) {
-      expect(e).toBeInstanceOf(Error);
-      expect(e.message).toBe('Source error');
+    } catch (e: unknown) {
+      const error = e as Error;
```
Shubhrakanti, line 493: Just curious - if we take in something as `unknown` and casting doesn't work in Runtime - how would the error show up?

agents/src/stream/merge_readable_streams.ts
Shubhrakanti: I agree we should just be able to pull these from `deno`.

agents/src/utils.test.ts
```diff
@@ -54,8 +54,8 @@ describe('AbortableTask', () => {
       // The task should reject with AbortError
       try {
         await task.result;
-      } catch (error: any) {
-        expect(error.name).toBe('AbortError');
+      } catch (error: unknown) {
+        expect((error as Error).name).toBe('AbortError');
```
Shubhrakanti, line 58: what's the default type for the `catch(error)` "error" in typescript? I wonder if there's a cleaner way to do this. cc: @lukasIO

Shubhrakanti: I guess this is the best? Man this a pain in the ass https://www.reddit.com/r/typescript/comments/rxm0q0/trycatch_in_typescript_is_such_a_gigantic_pain_in/

agents/src/utils.test.ts
```diff
@@ -224,7 +224,9 @@ describe('AbortableTask', () => {
       let child1Completed = false;
       let child2Completed = false;
 
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       let child1Task: any;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Shubhrakanti: would this be let child1Task: Task<void>? Also you don't have to add all the `// eslint-disable-next-line @typescript-eslint/no-explicit-any` in this PR. It's okay to have the warnings in the linter - it won't fail the build. We can come clean it up later. 

In fact it's probably better to throw a warning for now and clean it up later so we can carefully fix each one instead of just throwing around `// eslint-disable-next-line @typescript-eslint/no-explicit-any`.

agents/src/utils.test.ts
```diff
@@ -271,12 +273,15 @@ describe('AbortableTask', () => {
 
       // Verify all tasks were rejected with AbortError
       expect(parentResult.status).toBe('rejected');
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       expect((parentResult as any).reason.name).toBe('AbortError');
```
Shubhrakanti: Same as my point above. Is `parentResult as any` really what we want here? We can figure this out in a future PR but let's keep the warning so we know that there's an issue. We can get rid of `// eslint-disable-next-line @typescript-eslint/no-explicit-any`

agents/src/utils.test.ts
```diff
@@ -388,8 +393,8 @@ describe('AbortableTask', () => {
       try {
         await parentTask.result;
         expect.fail('Parent task should have thrown');
-      } catch (error: any) {
-        parentError = error;
+      } catch (error: unknown) {
+        parentError = error as Error;
```
Shubhrakanti, line 392: This is so annoying, but I guess there's no better solution in JS.

Shubhrakanti: I do agree that unknown is better than `any`

agents/src/voice/agent_activity.ts
Shubhrakanti: In the future let's also update the comment. In general I like to follow this [style guide for TODOs](https://google.github.io/styleguide/pyguide.html#312-todo-comments)

agents/src/voice/agent_activity.ts
Shubhrakanti: Bump?

agents/src/voice/audio_recognition.ts
```diff
@@ -177,81 +230,193 @@ export class AudioRecognition {
...
-        this.logger.error('Error in EOU detection task:', err);
+    this.bounceEOUTask.result
+      .then(() => {
+        this.logger.debug('EOU detection task completed');
+      })
+      .catch((err: any) => {
+        if (err.name === 'AbortError') {
+          this.logger.debug('EOU detection task was aborted');
+        } else {
+          this.logger.error('Error in EOU detection task:', err);
```
Shubhrakanti, line 248: I made a mistake here. This should be `this.logger.error(err, 'Error in EOU detection task:');`. We use the [pino logger](https://getpino.io/#/). The object comes first and then the message. It's really not intuitive and I don't like it - so if there's a better logger to use I'm not opposed to it. Just FYI

agents/src/utils.ts
Shubhrakanti: I'm an idiot. I didn't see that. Sorry

agents/src/worker.ts
```diff
@@ -165,7 +165,7 @@ export class WorkerOptions {
   jobMemoryWarnMB: number;
   jobMemoryLimitMB: number;
 
-  /** @param options */
+  /** @param options - Worker options */
```
Shubhrakanti, line 168: I appreciate you fixing all the linter warnings. Feel free to do this in a follow up PR next time.

agents/src/voice/io.ts
```diff
@@ -9,15 +9,18 @@ import type { SpeechEvent } from '../stt/stt.js';
 
 export type STTNode = (
   audio: ReadableStream<AudioFrame>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Shubhrakanti: Let's get rid of this. I want it to throw the warnings so I can just run `lint` and see what needs fixing.

Shubhrakanti: applies to anything with `modelSettings: any`

agents/src/voice/audio_recognition.ts
```diff
@@ -177,81 +230,193 @@ export class AudioRecognition {
...
+        } else {
+          this.logger.error('Error in user turn commit task:', err);
+        }
+      });
   }
 
-  setInputAudioStream(audioStream: ReadableStream<AudioFrame>) {
-    this.deferredInputStream.setSource(audioStream);
+  private get vadBaseTurnDetection() {
+    return this.turnDetectionMode === undefined || this.turnDetectionMode === 'vad';
```
Shubhrakanti, line 419: Random question - how should we organize methods in large classes? Maybe public methods on top then private? Wonder if there's a linter option to do this automatically or something

agents/src/voice/audio_recognition.ts
```diff
@@ -38,52 +44,75 @@ interface _TurnDetector {
...
-    private manualTurnDetection = false,
+    private turnDetectionMode?: TurnDetectionMode,
     private turnDetector?: _TurnDetector,
     private lastLanguage?: string,
   ) {
     this.deferredInputStream = new DeferredReadableStream<AudioFrame>();
+    delay;
+    const [vadInputStream, sttInputStream] = this.deferredInputStream.stream.tee();
+    this.vadInputStream = vadInputStream;
+    this.sttInputStream = sttInputStream;
```
Shubhrakanti, line 80: Just curious - why do this in the constructor instead of `start()`?

agents/src/voice/audio_recognition.ts
```diff
@@ -177,81 +230,193 @@ export class AudioRecognition {
...
 
-  private async vadTask(inputStream: ReadableStream<AudioFrame>) {
-    const vadStream = this.vad.stream();
-    vadStream.updateInputStream(inputStream);
+  setInputAudioStream(audioStream: ReadableStream<AudioFrame>) {
+    const mergedStream = mergeReadableStreams(
+      audioStream as any,
+      this.silenceAudioTransform.readable as any,
+    );
+    this.deferredInputStream.setSource(mergedStream as any);
```
Shubhrakanti, line 356: nice. this is sick

agents/src/voice/audio_recognition.ts
```diff
@@ -177,81 +230,193 @@ export class AudioRecognition {
...
+            if (typeof ev === 'string') {
+              throw new Error('STT node must yield SpeechEvent');
+            } else {
+              await this.onSTTEvent(ev);
+            }
+          }
+        } catch (e) {
+          if (isStreamReaderReleaseError(e)) {
+            return;
+          }
```
Shubhrakanti, line 296: You're 100% right we need an abstraction for this. ~~Can we add a comment here and link a ticket?~~

agents/src/stream/deferred_stream.test.ts
toubatbrian: This is to suppress the eslint warnings. If we use `any` here it will trigger warning

agents/src/utils.test.ts
toubatbrian: Let's make a separate PR to clean up this if necessary. I want this PR to be unblocked first.

agents/src/utils.test.ts
toubatbrian: This is what I originally did, and it will cause format build to fail in GHA.

agents/src/utils.test.ts
Shubhrakanti: yeah PRs approved so this is just a thought.

agents/src/voice/audio_recognition.ts
toubatbrian: https://www.npmjs.com/package/eslint-plugin-sort-class-members found this. I'll add a ticket for it.

Shubhrakanti: Awesome!

agents/src/voice/audio_recognition.ts
toubatbrian: Because this way we can define all class member variables; otherwise, we need to mark `vadInputStream` and `sttInputStream` as `ReadableStream<AudioFrame> | undefined`, which is fine but need to use `vadInputStream!` syntax everytime we use it

agents/src/voice/audio_recognition.ts
toubatbrian: what kind of ticket do you want to link to this?

agents/src/utils.test.ts
toubatbrian: SG, I just added a ticket for this

agents/src/voice/audio_recognition.ts
Shubhrakanti: On second though it's fine. I was thinking about adding a ticket to explore abstracting `read` with an abort signal. But first I need to try https://linear.app/livekit/issue/AJS-84/test-pipethrough. If that works we might not even need any of this any more.

PR #415, opened 2025-06-02
Shubhra/ajs 40 add interrupts to in agent activity

Shubhrakanti:
Add interrupt logic on agent speech. 
Follow up tasks:
 - Handle interrupt during agent playout
 - See if there's a better way to handle aborts in plugins
 - Transcript/Audio Synchronization


agents/src/voice/generation.ts
```diff
@@ -24,69 +25,107 @@ export function performLLMInference(
...
+        if (signal.aborted) {
+          console.log('++++ tts inference task aborted');
+          break;
+        }
+
         const { done, value: chunk } = await reader.read();
-        if (done) break;
-        writer.write(chunk);
+        console.log('++++ tts inference task read result:', {
+          done,
```
Shubhrakanti: this doesn't get cleaned up when there is a Long speech. We never get a `done` event, even though we call `controller.close()` in agent.ts.

agents/src/stream/identity_transform.ts
```diff
@@ -5,8 +5,12 @@ import { TransformStream } from 'node:stream/web';
...
   constructor() {
-    super({
-      transform: (chunk, controller) => controller.enqueue(chunk),
-    });
+    super(
+      {
+        transform: (chunk, controller) => controller.enqueue(chunk),
+      },
+      { highWaterMark: Number.MAX_SAFE_INTEGER },
+      { highWaterMark: Number.MAX_SAFE_INTEGER },
```
Shubhrakanti: Without this, the `ttsInferenceTask` would hang on abort because when we tried to write to the writer.

agents/src/voice/agent_activity.ts
```diff
@@ -104,7 +104,10 @@ export class AgentActivity implements RecognitionHooks {
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   onVADInferenceDone(ev: VADEvent): void {
-    // TODO(AJS-40): Implement this
+    if (this.currentSpeech && !this.currentSpeech.interrupted && this.currentSpeech.done) {
+      this.logger.info({ 'speech id': this.currentSpeech.id }, 'speech interrupted by VAD');
+      // this.currentSpeech.interrupt();
```
Shubhrakanti, line 115: will enable when all turn detection modes are in

agents/src/voice/generation.ts
```diff
@@ -5,10 +5,32 @@ import type { AudioFrame, AudioSource } from '@livekit/rtc-node';
...
+
+  return Promise.race([
+    promise,
+    new Promise<never>((_, reject) => {
+      signal.addEventListener('abort', () => {
+        reject(new DOMException('Operation was aborted', 'AbortError'));
+      });
+    }),
+  ]);
+}
```
Shubhrakanti: @lukasIO @toubatbrian  let me know what you think of this. Doesn't need to be perfect because we will refactor later with a Channel interface that takes in a abort signal in the `read` method

agents/src/stream/identity_transform.ts
toubatbrian: nice catch!

agents/src/tts/tts.ts
```diff
@@ -138,13 +139,15 @@ export abstract class SynthesizeStream
     const reader = this.deferredInputStream.stream.getReader();
     try {
       while (true) {
+        if (this.abortController.signal.aborted) {
+          break;
+        }
         const { done, value } = await reader.read();
```
toubatbrian: Will this cancel any pending read? I recommend:
```tsx
this.abortController.signal.addEventListener('abort', () => {
  this.deferredInputStream.detachSource()
});
```

agents/src/tts/tts.ts
```diff
@@ -138,13 +139,15 @@ export abstract class SynthesizeStream
...
         if (done || value === SynthesizeStream.FLUSH_SENTINEL) {
           break;
         }
         this.pushText(value);
       }
-      this.flush();
       this.endInput();
     } catch (error) {
       this.logger.error(error, 'Error reading deferred input stream');
```
toubatbrian, line 157: Also would be nice to add:
```tsx
finally {
  reader.releaseLock()
}
```

agents/src/voice/generation.ts
toubatbrian: I'm good with this as long as underling promise does not produce side-effects and can be finished properly. The reason is that if promise is a `while(true) do ...` statement, using `Promise.race` will only skip awaiting the promise but not cancel it.

agents/src/voice/agent_activity.ts
```diff
@@ -104,7 +104,10 @@ export class AgentActivity implements RecognitionHooks {
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   onVADInferenceDone(ev: VADEvent): void {
-    // TODO(AJS-40): Implement this
+    if (this.currentSpeech && !this.currentSpeech.interrupted && this.currentSpeech.done) {
```
toubatbrian: Should be `if (this.currentSpeech && !this.currentSpeech.interrupted && this.currentSpeech.allowInterruptions) {`?

agents/src/voice/agent_activity.ts
```diff
@@ -269,11 +281,12 @@ export class AgentActivity implements RecognitionHooks {
     // TODO(AJS-57): handle instructions
 
     this.agentSession._updateAgentState('thinking');
-    const tasks: Array<Promise<any>> = [];
+    const tasks: Array<Promise<void>> = [];
     const [llmTask, llmGenData] = performLLMInference(
       (...args) => this.agent.llmNode(...args),
```
toubatbrian, line 293: nit: 
```
performLLMInference(
  this.agent.llmNode,
  ...
)
```

agents/src/voice/agent_activity.ts
```diff
@@ -286,14 +299,27 @@ export class AgentActivity implements RecognitionHooks {
...
+      replyAbortController.abort();
+      await Promise.allSettled(tasks);
+      return;
+    }
 
-    const [textForwardTask, textOutput] = performTextForwarding(null, llmOutput);
+    const [textForwardTask, textOutput] = performTextForwarding(
+      null,
+      llmOutput,
+      replyAbortController.signal,
```
toubatbrian: In python I saw the `transcription_node` in this part, are we going to support that as well?

agents/src/voice/agent_activity.ts
```diff
@@ -269,11 +281,12 @@ export class AgentActivity implements RecognitionHooks {
     // TODO(AJS-57): handle instructions
 
     this.agentSession._updateAgentState('thinking');
-    const tasks: Array<Promise<any>> = [];
+    const tasks: Array<Promise<void>> = [];
     const [llmTask, llmGenData] = performLLMInference(
```
toubatbrian, line 291: nit: would be nice to use the `Task.from` primitive for cancellable promises, passing an abort controller `const llmTask = Task.from((signal) => ..., replyAbortController)` and `llmTask.cancelAndWait()`

agents/src/voice/agent_activity.ts
```diff
@@ -286,14 +299,27 @@ export class AgentActivity implements RecognitionHooks {
...
-    // TODO(AJS-40): handle interrupted authorization
-    await speechHandle.waitForAuthorization();
+    await speechHandle.waitIfNotInterrupted([speechHandle.waitForAuthorization()]);
+    if (speechHandle.interrupted) {
+      this.logger.info(
+        { speech_id: speechHandle.id },
+        'Speech interrupted after tts and llm tasks',
+      );
+      replyAbortController.abort();
+      await Promise.allSettled(tasks);
```
toubatbrian: ^

agents/src/voice/generation.ts
Shubhrakanti: yeah that's why I just kept it scoped to this file.

agents/src/voice/agent_activity.ts
```diff
@@ -313,18 +343,52 @@ export class AgentActivity implements RecognitionHooks {
...
+    }
+    if (textOutput && textOutput.text) {
+      const message = ChatMessage.create({
+        role: ChatRole.ASSISTANT,
+        text: textOutput.text,
+      });
+      chatCtx.insertItem(message);
+      this.agent._chatCtx.insertItem(message);
+      this.agentSession._conversationItemAdded(message);
+      this.logger.info(
```
toubatbrian, line 396: This part seems needed in both code path (interrupted, uninterrupted), can we lift this so we can reuse this piece of code? 

also, should we do` this.agentSession._updateAgentState('listening');` as well for un-interrupted situation?

agents/src/voice/generation.ts
```diff
@@ -24,68 +46,104 @@ export function performLLMInference(
...
+        return;
+      }
+
+      llmStreamReader = llmStream.getReader();
       while (true) {
-        const { done, value: chunk } = await reader.read();
+        if (signal?.aborted) {
+          break;
+        }
+        const { done, value: chunk } = await raceWithAbort(signal, llmStreamReader.read());
```
toubatbrian: I'm fine with this for this PR, but in the future we should make sure that we properly cancel the underlining read ops (stop up to openai SSE streaming part, so we potentially save $$$ for user as well, I guess if we cancel request earlier?).

agents/src/voice/generation.ts
```diff
@@ -24,68 +46,104 @@ export function performLLMInference(
...
-      const reader = ttsNode.getReader();
+      ttsStreamReader = ttsStream.getReader();
       while (true) {
-        const { done, value: chunk } = await reader.read();
-        if (done) break;
-        writer.write(chunk);
+        if (signal.aborted) {
+          break;
+        }
+        const { done, value: chunk } = await raceWithAbort(signal, ttsStreamReader.read());
```
toubatbrian: ^

agents/src/voice/generation.ts
```diff
@@ -133,27 +208,48 @@ async function forwardAudio(
...
-    if (!out.firstFrameFut.done) {
-      out.firstFrameFut.resolve();
+  try {
+    while (true) {
+      if (signal?.aborted) {
+        break;
+      }
+
+      const { done, value: frame } = signal
+        ? await raceWithAbort(signal, reader.read())
```
toubatbrian: ^

agents/src/voice/agent_activity.ts
Shubhrakanti: good catch. Should be same as https://github.com/livekit/agents/blob/0ede2d8b0fc0cc5b655b94b6c55f4ce0264155da/livekit-agents/livekit/agents/voice/agent_activity.py#L845-L849

agents/src/voice/agent_activity.ts
Shubhrakanti: We have to do this to preserve `this` context. Added a comment. also see https://github.com/livekit/agents-js/pull/385#discussion_r2087921955

agents/src/voice/agent_activity.ts
Shubhrakanti: Yeah I'll add tickets to do the transcription node defaults and synchronization with audio playout

agents/src/voice/generation.ts
lukasIO: agree with Brian, in order to avoid side effects here entirely, we'd probably have to clean up the pending abort promise?

toubatbrian: added a ticket for this: https://linear.app/livekit/issue/AJS-81/abortable-queuechannel-like-primitives-using-stream-api

Shubhrakanti: This doesn't handle the functionally of interrupts, but yeah we should refactor this to clean up the pending promise.

agents/src/voice/generation.ts
Shubhrakanti: Yes. We do this when we call LLMStream.close in the finally block. This cancels the underlying source defined in `llm.ts`

agents/src/voice/agent_activity.ts
Shubhrakanti: > This part seems needed in both code path (interrupted, uninterrupted), can we lift this so we can reuse this piece of code?

Absolutely. The reason I'm keeping this way is to keep it the exact same as the python implementation so it's easy to check line by line if there are differences / bug. Once we finish function calling and this function is complete - I think we should re-structure the entire pipeline reply to be more node-esque.

Shubhrakanti: We set the agent state with a callback (same as python). https://github.com/livekit/agents-js/blob/69e140b88b282a0808fe4c335cc4b9e87f457265/agents/src/voice/agent_activity.ts#L206-L210

agents/src/voice/generation.ts
Shubhrakanti: @toubatbrian @lukasIO I actually reverted this since it was causing weird bugs. I will deal with it once we're on https://linear.app/livekit/issue/AJS-81/abortable-queuechannel-like-primitives-using-stream-api

For now there are no functional issues with this so it's safe to merge.

agents/src/voice/agent_activity.ts
Shubhrakanti: ex. 

```typescript
private createPipelineTasks(
  speechHandle: SpeechHandle,
  ctx: PipelineContext,
  controller: AbortController
): PipelineTasks {
  const tasks = new Map<string, Task<any>>();
  
  // Create LLM inference task
  const [llmTask, llmGenData] = performLLMInference(
    (...args) => this.agent.llmNode(...args),
    ctx.chatCtx,
    {},
    controller,
  );
  tasks.set('llm', llmTask);
  
  // Create text streams
  const [ttsTextInput, llmOutput] = llmGenData.textStream.tee();
  
  // Create TTS task if audio output is enabled
  let ttsTask: Task<void> | null = null;
  let ttsStream: ReadableStream<AudioFrame> | null = null;
  if (ctx.audioOutput) {
    [ttsTask, ttsStream] = performTTSInference(
      (...args) => this.agent.ttsNode(...args),
      ttsTextInput,
      {},
      controller,
    );
    tasks.set('tts', ttsTask);
  }
  
  // Create forwarding tasks
  const [textForwardTask, textOutput] = performTextForwarding(
    ctx.textOutput,
    llmOutput,
    controller,
  );
  tasks.set('textForward', textForwardTask);
  
  let audioForwardTask: Task<void> | null = null;
  let audioOutput: any = null;
  if (ctx.audioOutput && ttsStream) {
    [audioForwardTask, audioOutput] = performAudioForwarding(
      ttsStream,
      ctx.audioOutput,
      controller,
    );
    tasks.set('audioForward', audioForwardTask);
  }
  
  return {
    tasks,
    llmGenData,
    textOutput,
    audioOutput,
    controller,
  };
}
```

Shubhrakanti:
```typescript
private async waitForPipelineAuthorization(
  speechHandle: SpeechHandle,
  pipelineTasks: PipelineTasks
): Promise<void> {
  // This follows the Python pattern of waiting for authorization
  // before starting the actual pipeline execution
  await speechHandle.waitIfNotInterrupted([speechHandle.waitForAuthorization()]);
  
  if (speechHandle.interrupted) {
    throw new PipelineInterruptedError('Speech interrupted before authorization');
  }
  
  this.agentSession._updateAgentState('thinking');
}

private async executePipeline(
  speechHandle: SpeechHandle,
  pipelineTasks: PipelineTasks
): Promise<void> {
  // Set up first frame callback to update agent state
  const onFirstFrame = () => {
    this.agentSession._updateAgentState('speaking');
  };
  
  if (pipelineTasks.audioOutput) {
    pipelineTasks.audioOutput.firstFrameFut.await.then(onFirstFrame);
  } else if (pipelineTasks.textOutput) {
    pipelineTasks.textOutput.firstTextFut.await.then(onFirstFrame);
  }
  
  // Wait for all tasks to complete or for interruption
  const taskPromises = Array.from(pipelineTasks.tasks.values()).map(task => task.result);
  await speechHandle.waitIfNotInterrupted(taskPromises);
  
  // Wait for audio playout if enabled
  if (pipelineTasks.audioOutput && this.agentSession.audioOutput) {
    await speechHandle.waitIfNotInterrupted([
      this.agentSession.audioOutput.waitForPlayout()
    ]);
  }
}

private async finalizePipeline(
  speechHandle: SpeechHandle,
  pipelineTasks: PipelineTasks,
  ctx: PipelineContext
): Promise<void> {
  if (speechHandle.interrupted) {
    await this.handlePipelineInterruption(speechHandle, pipelineTasks, ctx);
  } else {
    await this.handlePipelineCompletion(speechHandle, pipelineTasks, ctx);
  }
}

private async handlePipelineInterruption(
  speechHandle: SpeechHandle,
  pipelineTasks: PipelineTasks,
  ctx: PipelineContext
): Promise<void> {
  this.logger.debug(
    { speech_id: speechHandle.id },
    'Handling pipeline interruption',
  );
  
  // Cancel all tasks using the Task API
  const cancelPromises = Array.from(pipelineTasks.tasks.values()).map(
    task => task.cancelAndWait()
  );
  await Promise.allSettled(cancelPromises);
  
  // Handle interrupted message creation
  if (pipelineTasks.textOutput?.text) {
    const message = ChatMessage.create({
      role: ChatRole.ASSISTANT,
      text: pipelineTasks.textOutput.text,
    });
    ctx.chatCtx.insertItem(message);
    this.agent._chatCtx.insertItem(message);
    this.agentSession._conversationItemAdded(message);
  }
  
  // Update agent state
  if (this.agentSession.agentState === 'speaking') {
    this.agentSession._updateAgentState('listening');
  }
  
  speechHandle.markPlayoutDone();
}

private async handlePipelineCompletion(
  speechHandle: SpeechHandle,
  pipelineTasks: PipelineTasks,
  ctx: PipelineContext
): Promise<void> {
  if (pipelineTasks.textOutput?.text) {
    const message = ChatMessage.create({
      role: ChatRole.ASSISTANT,
      text: pipelineTasks.textOutput.text,
    });
    ctx.chatCtx.insertItem(message);
    this.agent._chatCtx.insertItem(message);
    this.agentSession._conversationItemAdded(message);
    
    this.logger.info(
      { speech_id: speechHandle.id, message: pipelineTasks.textOutput.text },
      'Pipeline completed successfully',
    );
  }
  
  speechHandle.markPlayoutDone();
}

private async handlePipelineError(
  speechHandle: SpeechHandle,
  controller: AbortController,
  error: any
): Promise<void> {
  this.logger.error(
    { speech_id: speechHandle.id, error },
    'Pipeline error occurred',
  );
  
  controller.abort();
  speechHandle.markPlayoutDone();
  
  if (this.agentSession.agentState === 'speaking') {
    this.agentSession._updateAgentState('listening');
  }
}
```

Shubhrakanti:
```typescript
// Add to SpeechHandle class

private pipelineTasks: Map<string, Task<any>> = new Map();

addPipelineTask(name: string, task: Task<any>): void {
  this.pipelineTasks.set(name, task);
}

async cancelAllPipelineTasks(): Promise<void> {
  const cancelPromises = Array.from(this.pipelineTasks.values()).map(
    task => task.cancelAndWait()
  );
  await Promise.allSettled(cancelPromises);
  this.pipelineTasks.clear();
}

// Enhanced interrupt method
interrupt(): SpeechHandle {
  if (!this.#allowInterruptions) {
    throw new Error('interruptions are not allowed');
  }
  
  this.interruptFut.resolve();
  
  // Cancel all associated pipeline tasks
  this.cancelAllPipelineTasks().catch(error => {
    // Log but don't throw - interruption should always succeed
    console.error('Error cancelling pipeline tasks:', error);
  });
  
  return this;
}
```

.github/workflows/test.yml
```diff
@@ -42,28 +42,29 @@ jobs:
...
-          CARTESIA_API_KEY: ${{ secrets.CARTESIA_API_KEY }}
-          NEUPHONIC_API_KEY: ${{ secrets.NEUPHONIC_API_KEY }}
-          RESEMBLE_API_KEY: ${{ secrets.RESEMBLE_API_KEY }}
-        run: |
-          plugins=$(git diff-tree --name-only --no-commit-id -r ${{ github.sha }} | grep '^plugins.*\.ts$' | cut -d/ -f2 | sort -u | tr '\n' ' ')
-          read -ra plugins <<< "$plugins"
-          for plugin in "${plugins[@]}"; do
-            pnpm test $plugin
-          done
+      # TODO (AJS-83) Re-enable once plugins are refactored with abort controllers
```
Shubhrakanti, line 45: Will do this before release

agents/src/tts/tts.ts
```diff
@@ -131,6 +132,13 @@ export abstract class SynthesizeStream
     this.#tts = tts;
     this.deferredInputStream = new DeferredReadableStream();
     this.mainTask();
+    this.abortController.signal.addEventListener('abort', () => {
+      this.deferredInputStream.detachSource();
```
toubatbrian, line 136: lg

agents/src/voice/agent_activity.ts
toubatbrian: are vad turn detection mode the only missing piece to enable this line?

Shubhrakanti: oh shoot I missed this. I think we should be able to turn this on now. I'll test it out.

agents/src/voice/agent_activity.ts
```diff
@@ -313,18 +348,52 @@ export class AgentActivity implements RecognitionHooks {
...
 
-    this.logger.info({ speech_id: speechHandle.id }, 'playout completed');
-    speechHandle.markPlayoutDone();
+    if (speechHandle.interrupted) {
+      this.logger.debug(
+        { speech_id: speechHandle.id },
+        'Aborting all pipeline reply tasks due to interruption',
+      );
+      replyAbortController.abort();
+      await Promise.allSettled(tasks.map((task) => task.cancelAndWait));
```
toubatbrian: btw `cancelAndWait`, when take no timeout parameter, will wait indefinitely if underlining async promise is hanging. Do we want to pass a default timeout threshold (5s?).

Shubhrakanti: We wait indefinitely in python as well - but yeah I think it's good to have an explicit timeout in the resource clean up. I'll add one.

Shubhrakanti: If any of the tasks fails to settle, the next turn won't take place. The agent can't speak since main loop in `AgentAcitivty` will be blocked as `speechHandle.markPlayoutDone()` will never be reached. 

I still think it's better to throw a timeout error instead of just handing so I'll add one.

agents/src/voice/generation.ts
```diff
@@ -24,72 +25,111 @@ export function performLLMInference(
...
+
     try {
-      const reader = llmStream.getReader();
+      llmStream = await node(chatCtx, modelSettings);
+      if (llmStream === null) {
+        await outputWriter.close();
+        return;
+      }
+
+      llmStreamReader = llmStream.getReader();
```
toubatbrian, line 45: nit: (totally ok to skip this, but can add a ticket)
Right now the pretty hacky way I can think of to "pause" the `await reader.read()` is to wrap the sourceStream with deferred readable stream:

```tsx
deferredLlmStream = new DeferredReadableStream();
deferredLlmStream.setSource(llmStream);

llmStreamReader = deferredLlmStream.stream().getReader();

// setup abort signal listener
signal.addEventListener("abort", deferredLlmStream.detachSource);

// inside while (true), we do
const { done, value: chunk } = await llmStreamReader.read(); // this gets unblocked when abort signal fires
```

Let me know what you think, but I'm also ok to keep the current version!

agents/src/voice/generation.ts
```diff
@@ -24,72 +25,111 @@ export function performLLMInference(
...
     } finally {
-      writer.close();
+      llmStreamReader?.releaseLock();
+      await llmStream?.cancel();
+      await outputWriter.close();
     }
   };
-  return [inferenceTask(), data];
+
+  return [Task.from((controller) => inferenceTask(controller.signal), controller), data];
```
toubatbrian, line 81: I'm seeing repeated pattern on different tasks returning a tuple like data structure. One request is can we declare it's type? e.g. type TaskOutout = [Task<T>, ...]; so when pp reading it, they know what 1st and 2nd arguments mean?

toubatbrian (approved):
Added a few comments, nothing is major concerns. LGTM! Feel free to merge it

agents/src/voice/generation.ts
Shubhrakanti: The issue is the second element for the tuple is different for each task, `[Task<void>, _AudioOutput]`, `[Task<void>, _TextOutput]`, `[Task<void>, ReadableStream<AudioFrame>]`, `[Task<void>, _LLMGenerationData]`. I think keeping it a tuple is the cleanest solution. 

I would however like to restructure the `pipelineReplyTask` to maybe even get rid of these `perform` functions. It might cleaned to have a lot of this in `SpeechHandle` and have the handle cancel all the tasks.

agents/src/voice/generation.ts
Shubhrakanti: Creating a new `DeferredReadableStream()` and calling `setSource` feels like a misuse of the `DeferredReadableStream` class. As we discussed this morning, what we likely need instead is a wrapper around `ReadableStream.read` that handles the abort signal properly.

Technically, the line `llmStreamReader.read()` could hang on cancellation (though I haven't been able to reproduce that case). That said, the timeout in `Task.cancelAndAwait` should catch it if it does happen. I’d prefer to wait until we see it occur before making any changes.

agents/src/voice/agent_activity.ts
Shubhrakanti: Added one for 5 seconds. Great idea on adding this.

PR #416, opened 2025-06-03
Add Abortable Task Primitive

toubatbrian:


agents/src/utils.ts
```diff
@@ -339,3 +339,49 @@ export class AudioEnergyFilter {
     return false;
   }
 }
+
+export class AbortableTask<T> {
+  #controller: AbortController;
```
Shubhrakanti: don't use #private

Shubhrakanti (approved):

PR #417, opened 2025-06-03
Revert "add termiante to JobRequest (#399)"

Shubhrakanti:
This reverts commit a1931398b55bb448452a487248ff6a67a8dacedb. This was causing builds to fail on main.

### DIFF
--- a/.changeset/happy-forks-march.md
+++ b/.changeset/happy-forks-march.md
@@ -1,5 +1,0 @@
----
-"@livekit/agents": patch
----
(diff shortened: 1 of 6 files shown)


Shubhrakanti:
> it looked like a bunch of modules hadn't been updated in 6 months and fixing that seemed out of scope for this change so i was planning to wait until the project got renovated to land this

Yeah we should do this with the 1.0 changes. I'll make a note.

PR #419, opened 2025-06-05
Brian/cancel deferred readable stream

toubatbrian:


agents/src/stream/deferred_stream.ts
```diff
@@ -2,26 +2,60 @@
...
    * Call once the actual source is ready.
    */
   setSource(source: ReadableStream<T>) {
-    if (this.transform.writable.locked) {
-      throw new Error('Stream is already locked');
+    if (this.pipeTask) {
+      throw new Error('Stream source already set');
     }
-    source.pipeTo(this.transform.writable);
+
```
toubatbrian, line 51: we explicitly do getReader to read from source and write to destination. So we have more control over when we release lock, close / cancelling streams

agents/tests/stream/deferred_stream.test.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    // Read should indicate end of stream
+    const result = await readPromise;
+    expect(result.done).toBe(true);
+    expect(result.value).toBeUndefined();
+
+    reader.releaseLock();
+  });
+
+  it('source set by another deferred stream after calling cancel()', async () => {
```
toubatbrian: this is the usecase we NEED for doing swapping on stt source to be piped to different deferred readable streams

agents/src/stream/deferred_stream.ts
```diff
@@ -2,26 +2,64 @@
...
+    if (this.pipeTask) {
+      throw new Error('Stream source already set');
+    }
+
+    this.pipeTask = createTask(async (controller) => {
+      try {
+        const reader = source.getReader();
+
+        while (!controller.signal.aborted) {
+          const { done, value } = await reader.read();
```
lukasIO: if you're aborting while waiting for a chunk, but the chunk never arrives, this will stall indefinitely, I think

toubatbrian: Thanks for catching that! I put the sourceReader.releaseLock() under `detachSource` right before firing abort signal. This should help unblock the `await reader.read()`.

agents/tests/stream/deferred_stream.test.ts
```diff
@@ -0,0 +1,330 @@
...
+
+            controller.enqueue('data');
+            controller.close();
+        }
+    });
+
+    deferred.setSource(source);
+    // the trick here is that we have to do both reader.cancel() and detachSource() in this order!!!
+    await reader.cancel();
+    await deferred.detachSource();
```
toubatbrian: We unfortunately have to follow a strict order of operations when dealing with a hanging source readable stream. If we don’t call `await reader.cancel()` first, the initial `readPromise` can throw an error.

Currently, users (mostly for internal devs) are expected to manually manage when to call `deferred.stream.getReader().cancel()` or `.releaseLock()`, which is prone to errors in cases like this. At the same time, we can’t safely handle the lifecycle of `deferred.stream.getReader()` ourselves, because `this.transform.readable.cancel()` won’t proceed if there’s an active external `getReader` blocking it.

cc @lukasIO wondering if you have any idea on how we can improve this?

agents/tests/stream/deferred_stream.test.ts
```diff
@@ -0,0 +1,341 @@
...
+    expect(readCompleted).toBe(false);
+
+    // Create and set the source
+    const sourceData = ['hello', 'world'];
+    const source = new ReadableStream<string>({
+      start(controller) {
+        sourceData.forEach((chunk) => controller.enqueue(chunk));
+        controller.close();
+      },
+    });
```
Shubhrakanti, line 42: helper function?

agents/src/stream/deferred_stream.ts
```diff
@@ -2,26 +2,67 @@
...
+    return this.transform.readable;
   }
 
   /**
    * Call once the actual source is ready.
    */
   setSource(source: ReadableStream<T>) {
-    if (this.transform.writable.locked) {
-      throw new Error('Stream is already locked');
+    if (this.pipeTask || this.sourceReader) {
```
Shubhrakanti: duplicate state - wonder if we can group it. But seems fine for now.

Shubhrakanti: maybe a `sourceSet` private method?

agents/tests/stream/deferred_stream.test.ts
Shubhrakanti: hmmm can we abstract this for now? maybe a helper class over `Deffered stream? or somehow detect if there are any readers active in `detatchSource` and throw an error or something?

Shubhrakanti (approved):

agents/tests/stream/deferred_stream.test.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  it("should transfer source between deferred streams while reading is ongoing", async () => {
+    console.log('starting test');
+    const deferred1 = new DeferredReadableStream<string>();
+    const deferred2 = new DeferredReadableStream<string>();
+    
+    // Create a source that slowly emits data
+    let emitCount = 0;
+    const source = new ReadableStream<string>({
+      async start(controller) {
```
toubatbrian: solved

agents/tests/stream/deferred_stream.test.ts
toubatbrian: To abstract this, we have to only expose `this.transform.readable.getReader` then. We can't expose the underlining `this.transform.readable` ReadableStream to external since the close/cancel/releaseLock may not be properly handled in right timing. The donwnside of doing so is that we completely lose the benefit of Stream API since we cannot piping the `deferredStream.stream` to other TransformStream and we only have the method to keep getting data from it (much like a AsyncIterator).

agents/tests/stream/deferred_stream.test.ts
lukasIO: what exactly is the use case here? stt source swapping due to fallback?

toubatbrian: This is for `clear_user_turn` method, where we aim to “detach” the source readable from the deferred stream that pipes into the STT stream. The goal is to ensure that the previous STT stream task receives { done: true, value: undefined } after detachment and exits gracefully.

However, current implementation is still buggy as we can’t call `releaseLock` on the reader if it’s actively awaiting `read()` in a loop. I’m putting this PR on pause for now and will revisit it once I have a more complete understanding of the various use cases.

agents/tests/stream/deferred_stream.test.ts
toubatbrian: solved

agents/tests/stream/deferred_stream.test.ts
toubatbrian: @lukasIO I figured out way to unblock this, now all tests should be passed!

agents/src/stream/deferred_stream.ts
toubatbrian: For more details, when calling reader.releaseLock(), the pending reader.read() will raise TypeError: Invalid state: Releasing reader. Now this case is properly handled as well.

PR #432, opened 2025-06-10
Improve Task primitive 

toubatbrian:


.eslintrc
```diff
@@ -22,6 +22,6 @@
     "@typescript-eslint/ban-ts-comment": "warn",
     "@typescript-eslint/no-empty-interface": "warn",
     "@typescript-eslint/consistent-type-imports": "warn",
-    "@typescript-eslint/no-explicit-any": "warn",
+    "@typescript-eslint/no-explicit-any": "off",
```
Shubhrakanti: I would prefer we do these on a case by case basis with a comment like `// @ts-ignore `

agents/src/voice/io.ts
```diff
@@ -7,17 +7,24 @@ import type { ChatContext } from '../llm/chat_context.js';
 import type { ChatChunk } from '../llm/llm.js';
 import type { SpeechEvent } from '../stt/stt.js';
 
+export type NodeOptions = {
+  signal?: AbortSignal;
+};
+
```
Shubhrakanti: why do we need this? the return type of a node is a `=> Promise<ReadableStream<...>>` so when you call `cancel` on the stream we know the consumer is done reading it. No need to pass an abort signal.

agents/src/utils.ts
```diff
@@ -340,7 +341,21 @@ export class AudioEnergyFilter {
...
 }
 
-export class AbortableTask<T> {
+export const TASK_TIMEOUT_ERROR = new Error('Task cancellation timed out');
+
+export enum TaskResult {
+  Timeout = 'timeout',
+  Completed = 'completed',
+  Aborted = 'aborted',
+}
```
Shubhrakanti, line 350: Do we have an actual use case for this? Most of these "tasks" in the agents code base are long running loops.

Shubhrakanti: If it's just for testing, then I suggest we move this outside of the public interface

agents/src/utils.ts
```diff
@@ -340,7 +341,21 @@ export class AudioEnergyFilter {
...
+
+/** @internal */
+/**
+ * A task that can be cancelled.
+ *
+ * @param T - The type of the task result
+ */
+export class Task<T> {
   private resultFuture: Future<T>;
```
Shubhrakanti, line 372: When should we use `Task` vs `Future`? Can we make the claim that a task does everything a `Future` does?

agents/src/utils.ts
```diff
@@ -351,6 +366,18 @@ export class AbortableTask<T> {
...
+   * Creates a new task from a function.
+   *
+   * @param fn - The function to run
+   * @param controller - The abort controller to use
+   * @returns A new task
+   */
+  static from<T>(fn: (controller: AbortController) => Promise<T>, controller?: AbortController) {
+    const abortController = controller ?? new AbortController();
+    return new Task(fn, abortController);
+  }
```
Shubhrakanti, line 392: Ah I see why you want the task constructor to take in an `AbortController`.

I'm worried that this design allows the user to do something like this.

```typescript
// don't specify an abort controller for the parent task
parentTask = Task.from( (controller) => {
    ...
    // create a separate controller for the child task
    childController =  new AbortController()
    childTask = Task.from( (controller) => {...},  childController)
    ....
});
// cancelling the parent task won't cancel the child task
parent.cancel()
```

Allowing the `Task` to take in an `AbortController` might lead to unintended behavior if not used correctly . Maybe it's better to just have all cancellations stem from `AbortController.abort()` and have the user of the `Task` interface handle the actual cancellation logic.

agents/src/voice/io.ts
toubatbrian: I need to use this for STT stream resetting, so I guess yeah we probably need it. Also, it'd be helpful to hide abort logic inside ReadableStream when we define it

agents/src/utils.ts
toubatbrian: Yes, we have. This lets us determine whether a task was cancelled due to a timeout, explicitly aborted (e.g., an AbortError was raised), or completed normally (e.g., while (!signal.aborted) ...). We can then use this information for further processing.

In my opinion, we should return some status indicator, even if we don’t currently use the status value. It’s still important to include it.

It’s similar to apis like `err = openFile(...)` where the operation may not success. Some might call `openFile()` directly without checking the return value, but it should still be there by design.

agents/src/utils.ts
toubatbrian: I think these are fundamentally different concepts.

A `Future` represents a value or result that will be available at some point and typically needs to be awaited.

A `Task`, on the other hand, is an asynchronous operation (essentially a promise) that encapsulates cancellation logic. It allows the developer to design the async operation with an abort signal in mind, making it explicitly abortable.

agents/src/utils.ts
toubatbrian:
```
// Don't specify an abort controller for the parent task
const parentTask = Task.from((controller) => {
  ...
  // Create a separate controller for the child task
  const childController = new AbortController();
  const childTask = Task.from((controller) => { ... }, childController);
  ...
});
// Cancelling the parent task won't cancel the child task
parent.cancel();
```

I feel like this code is not inherently wrong, as there might be valid scenarios where we want to cancel the outer task while keeping the inner task running. 

The reason `.from` accepts an optional controller is for 2 reasons:
	1.	It allows a single controller to be passed to nested tasks, to help unify cancellation logic.
	2.	It supports cases where a controller is already defined externally (like an abortController defined in class we want to use), which avoid the need for the task to manage its own.


Without this, achieving unified cancellation (as in point 1) would require additional boilerplate. For example: right now we can write
```tsx
const parent = Task.from((controller) => {
  const child1 = Task.from(() => { ... }, controller);
  const child2 = Task.from(() => { ... }, controller);
});
parent.cancel();
```
to cancel entire async task.
However, If `.from` didn’t accept a controller, we'd need something more verbose like:
```tsx
const parent = Task.from((controller) => {  
  const child1 = Task.from(() => { ... });
  const child2 = Task.from(() => { ... });

  controller.signal.addEventListener("abort", () => {
    child1.cancel();
    child2.cancel();
  });  
});
parent.cancel();
```
In my view, the latter is unnecessarily messy.

As for passing controller vs. signal, if the goal is convenience and we’re not passing the controller further, we can simply do:
```tsx
const parent = Task.from(({ signal }) => { ... });
```
That should be cleaner.

Since this primitive is intended for internal use, we can mitigate misuse through some clear docs and examples to convey its intended usage. Let me know what you think!

agents/src/utils.ts
Shubhrakanti: When do we timeout a task in the pipeline logic? the only place I can think of is denouncing the end of turn?

agents/src/utils.ts
Shubhrakanti: sounds good. maybe add some comments on how we should setup nested tasks - but other than that this is lgtm

agents/src/voice/io.ts
Shubhrakanti: Do we need to have it in this PR? I don't see `nodeOptions` defined anywhere.

Shubhrakanti (approved):

PR #435, opened 2025-06-11
Shubhra/ajs 78 wait for audio playout in pipeline reply task

Shubhrakanti:
Adds a basic mechanism to interrupt the agent after LLM/TTS is done, but there is still audio waiting to be played out in real time.

This flow is very confusing in the python codebase so here's a diagram that roughly explains what's supposed to happen.
mermaid-diagram-2025-06-11-191218


agents/src/pipeline/speech_handle.ts
```diff
@@ -224,8 +224,8 @@ export class SpeechHandle {
     return this.playoutDoneFut.await;
   }
 
-  async waitIfNotInterrupted(aw: Promise<void>[]): Promise<void> {
-    const fs: Promise<void>[] = [...aw, this.interruptFut.await];
+  async waitIfNotInterrupted(aw: Promise<unknown>[]): Promise<void> {
+    const fs: Promise<unknown>[] = [...aw, this.interruptFut.await];
```
Shubhrakanti: This is so that we can pass in a `Promise` with a return any return type and wait for it. I think we want [`unknown` over `any`](https://stackoverflow.com/questions/51439843/unknown-vs-any). Correct me if I'm wrong.

agents/src/voice/generation.ts
```diff
@@ -220,14 +218,15 @@ async function forwardAudio(
     throw error;
   } finally {
     reader?.releaseLock();
+    audioOuput.flush();
```
Shubhrakanti, line 221: There's a common pattern in the python code base to muddle clean-up logic and core pipeline logic in the `finally` clause. This is because `asyncio.cancel` will run anything in the finally block before exiting the task. This makes the code super difficult to read. Nothing much we can do about it - just something to keep in mind.

agents/src/voice/generation.ts
```diff
@@ -182,15 +183,15 @@ export function performTextForwarding(
   ];
 }
 
-export interface _AudioOutput {
+export interface _AudioOut {
```
Shubhrakanti, line 186: Rename to keep it consistent with python for now. Not a great name since we also have `AudioOuput` in the `AgentActivity` class.

agents/src/voice/room_io.ts
```diff
@@ -46,10 +204,6 @@ export class RoomIO {
     this.room.on(RoomEvent.TrackSubscribed, this.onTrackSubscribed);
   }
 
-  private cleanup() {
-    this.room.off(RoomEvent.TrackSubscribed, this.onTrackSubscribed);
-  }
-
```
Shubhrakanti, line 52: This was unused

agents/src/voice/room_io.ts
```diff
@@ -16,8 +16,167 @@ import {
...
 import type { AgentSession } from './agent_session.js';
 
+export interface PlaybackFinishedEvent {
+  // How much of the audio was played back
+  playbackPosition: number;
+  // Interrupted is True if playback was interrupted (clearBuffer() was called)
+  interrupted: boolean;
+  // Transcript synced with playback; may be partial if the audio was interrupted
+  // When null, the transcript is not synchronized with the playback
+  synchronizedTranscript: string | null;
```
Shubhrakanti: Should this be undefined? Python equivalent - https://github.com/livekit/agents/blob/be69745c2b52620a3502f33f553be1a09cbd5a75/livekit-agents/livekit/agents/voice/io.py#L80-L88

agents/src/voice/room_io.ts
```diff
@@ -16,8 +16,167 @@ import {
...
+}
+
+export interface AudioOutputOptions {
+  sampleRate: number;
+  numChannels: number;
+  trackPublishOptions: TrackPublishOptions;
+  queueSizeMs?: number;
+}
+export class ParticipantAudioOutput {
```
Shubhrakanti, line 38: There core logic is split into an [abstract class](https://github.com/livekit/agents/blob/be69745c2b52620a3502f33f553be1a09cbd5a75/livekit-agents/livekit/agents/voice/io.py#L91-L196) and [Livekit room specific class](https://github.com/livekit/agents/blob/be69745c2b52620a3502f33f553be1a09cbd5a75/livekit-agents/livekit/agents/voice/room_io/_output.py#L20-L123) in python since the user can custom define an `AudioOutput`. For now I left it one class so the PR is easier to review/follow.

agents/src/voice/room_io.ts
```diff
@@ -16,8 +16,167 @@ import {
...
+    }
+
+    return this.lastPlaybackEvent;
+  }
+
+  private async waitForPlayoutTask(abortController: AbortController): Promise<void> {
+    const waitForPlayout = this.audioSource.waitForPlayout();
+    const waitForInterruption = new Promise<void>((resolve) => {
+      abortController.signal.addEventListener('abort', () => resolve());
+    });
```
Shubhrakanti: is there a cleaner way to do this? Like abstract waiting for the abort signal? Also don't we need to clean up this listener too?

Shubhrakanti: @lukasIO how would we clean up the listener after the race finishes?

lukasIO: in cases where you never actually abort, the promise will just be kept around, I believe. 

an easy solution would be to wrap both of them in a single promise. something like

```
new Promise((resolve, reject) => {
        abortController.signal.addEventListener('abort', () => resolve());
        await this.audioSource.waitForPlayout().then(resolve()).catch(reject)
```

agents/src/voice/room_io.ts
lukasIO: probably should be called on disconnections?

Shubhrakanti: Yeah I'm going to do the disconnect logic and clean up logic in a separate PR. This is just cursor added code I didn't remove.

agents/src/voice/speech_handle.ts
```diff
@@ -201,8 +201,8 @@ export class SpeechHandle {
     return this.playoutDoneFut.await;
   }
 
-  async waitIfNotInterrupted(aw: Promise<void>[]): Promise<void> {
-    const fs: Promise<void>[] = [...aw, this.interruptFut.await];
+  async waitIfNotInterrupted(aw: Promise<unknown>[]): Promise<void> {
```
Shubhrakanti, line 204: @toubatbrian thanks to #412 I learned about `unknown`

agents/src/voice/agent_activity.ts
```diff
@@ -396,7 +401,20 @@ export class AgentActivity implements RecognitionHooks {
       await Promise.allSettled(
         tasks.map((task) => task.cancelAndWait(AgentActivity.REPLY_TASK_CANCEL_TIMEOUT)),
       );
-      // TODO(shubhra): add waiting for audio playout in audio output and syncronizher transcripts
+      // TODO(shubhra): add syncronizher and transcripts
```
Shubhrakanti: add ticket number

agents/src/voice/room_io.ts
toubatbrian: `synchronizedTranscript?: string;`

agents/src/voice/room_io.ts
```diff
@@ -16,8 +16,173 @@ import {
...
+        .waitForPlayout()
+        .then(() => {
+          abortController.signal.removeEventListener('abort', abortHandler);
+          resolve(false);
+        })
+        .catch(() => {
+          abortController.signal.removeEventListener('abort', abortHandler);
+          resolve(false);
+        });
+    });
```
toubatbrian: Just a thought, maybe this should looks cleaner?

```tsx
const abortFuture = new Future<bool>();

const resolveAbort = () => {
  if (!abortFuture.done) abortFuture.resolve(true);
}

abortController.signal.addEventListener('abort', resolveAbort);

this.audioSource.waitForPlayout().finally(() => {
  abortController.signal.removeEventListener('abort', resolveAbort);
  if (!abortFuture.done) abortFuture.resolve(false);
}),

const interrupted = await abortFuture;
```

toubatbrian: These two version should be equivalent; however, just want you to make sure that it doesn't actually cancel underlining `waitForPlayout` promise, which still proceed to finish

toubatbrian (commented):
Add some comments

agents/src/voice/room_io.ts
Shubhrakanti: good suggestion - implemented.

toubatbrian (approved):

PR #438, opened 2025-06-12
brianyin/ajs-88-add-tool-class-generic-tool-typing

toubatbrian:
Setup new tool user-facing APIs and refactored the schema to align with updated interfaces.
- Defined new tool interfaces modeled similar to lk-agent v1.0, Vercel, and Mastra’s tool interface.
- Enabled agent definitions using the new tool interfaces (definition only for now—execution logic will follow in upcoming PRs).
- Refactored parts of the existing implementation from the old function context to the new tool context.

Most of the diffs involve copying existing helper functions, renaming parameters, and structural refactoring. Happy to walk through it if a read-through would help!

### DIFF
--- a/agents/src/llm/index.ts
+++ b/agents/src/llm/index.ts
@@ -28,7 +28,6 @@ export {
   LLMEvent,
   LLMStream,
   type ChatChunk,
-  type Choice,
   type ChoiceDelta,
   type CompletionUsage,
   type LLMCallbacks,
--- a/agents/src/llm/llm.ts
+++ b/agents/src/llm/llm.ts
@@ -6,10 +6,8 @@ import { EventEmitter } from 'node:events';
 import { log } from '../log.js';
 import type { LLMMetrics } from '../metrics/base.js';
 import { AsyncIterableQueue } from '../utils.js';
-import type { FunctionCallOutput } from './chat_context.js';
 import { type ChatContext, type ChatRole, type FunctionCall } from './chat_context.js';
 import type { ToolContext } from './tool_context.js';
-import { executeToolCall } from './utils.js';
 
 export interface ChoiceDelta {
   role: ChatRole;
@@ -23,14 +21,9 @@ export interface CompletionUsage {
   totalTokens: number;
 }
 
-export interface Choice {
-  delta: ChoiceDelta;
-  index: number;
-}
-
 export interface ChatChunk {
-  requestId: string;
-  choices: Choice[];
+  id: string;
+  delta?: ChoiceDelta;
   usage?: CompletionUsage;
 }
 
@@ -65,7 +58,6 @@ export abstract class LLMStream implements AsyncIterableIterator<ChatChunk> {
   protected output = new AsyncIterableQueue<ChatChunk>();
   protected queue = new AsyncIterableQueue<ChatChunk>();
   protected closed = false;
-  protected _functionCalls: FunctionCall[] = [];
   protected abortController = new AbortController();
   protected logger = log();
   abstract label: string;
@@ -94,7 +86,7 @@ export abstract class LLMStream implements AsyncIterableIterator<ChatChunk> {
 
     for await (const ev of this.queue) {
       this.output.put(ev);
-      requestId = ev.requestId;
+      requestId = ev.id;
       if (!ttft) {
         ttft = process.hrtime.bigint() - startTime;
       }
@@ -121,11 +113,6 @@ export abstract class LLMStream implements AsyncIterableIterator<ChatChunk> {
     this.#llm.emit(LLMEvent.METRICS_COLLECTED, metrics);
   }
 
-  /** List of called functions from this stream. */
-  get functionCalls(): FunctionCall[] {
-    return this._functionCalls;
-  }
-
   /** The function context of this stream. */
   get toolCtx(): ToolContext | undefined {
     return this.#toolCtx;
@@ -134,15 +121,6 @@ export abstract class LLMStream implements AsyncIterableIterator<ChatChunk> {
   /** The initial chat context of this stream. */
   get chatCtx(): ChatContext {
     return this.#chatCtx;
-  }
-
-  /** Execute all deferred functions of this stream concurrently. */
-  async executeFunctions(): Promise<FunctionCallOutput[]> {
-    return await Promise.all(
-      this._functionCalls.map(async (f) => {
-        return await executeToolCall(f, this.#toolCtx!);
-      }),
-    );
   }
 
   next(): Promise<IteratorResult<ChatChunk>> {
--- a/agents/src/voice/generation.ts
+++ b/agents/src/voice/generation.ts
@@ -56,14 +56,23 @@ export function performLLMInference(
           data.generatedText += chunk;
           await outputWriter.write(chunk);
           // TODO(shubhra): better way to check??
-        } else if ('choices' in chunk) {
-          const content = chunk.choices[0]?.delta.content;
-          if (!content) continue;
-          data.generatedText += content;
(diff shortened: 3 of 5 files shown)


agents/src/llm/utils.ts
```diff
@@ -95,3 +98,157 @@ export async function serializeImage(image: ImageContent): Promise<SerializedIma
...
+  if (!tool) {
+    throw new Error(`AI tool ${toolName} not found`);
+  }
+
+  return FunctionCall.create({
+    callId: toolCallId,
+    name: toolName,
+    args: rawArgs,
+  });
+};
```
toubatbrian: These functions are copied over from the existing `function_context.ts` file. No need to review them

agents/src/llm/index.ts
```diff
@@ -1,15 +1,25 @@
...
 // SPDX-License-Identifier: Apache-2.0
+// export {
+//   oaiBuildFunctionInfo,
+//   oaiParams,
+//   type CallableFunction,
+//   type CallableFunctionResult,
+//   type FunctionCallInfo,
+//   type FunctionContext,
+//   type inferParameters,
+// } from './function_context.js';
```
Shubhrakanti: why not just delete?

agents/src/voice/agent.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    const agent = createAgent({ instructions, tools });
+
+    const tools1 = agent.tools;
+    const tools2 = agent.tools;
+
+    // Should return different object references
+    expect(tools1).not.toBe(tools2);
+    expect(tools1).not.toBe(tools);
+
+    // But should have the same content
```
Shubhrakanti: Let's be careful of the AI generated comments.

agents/src/voice/agent.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    expect(agentTools).toHaveProperty('getTool2');
+
+    // Verify tool properties with proper checks
+    expect(agentTools.getTool1?.name).toBe('getTool1');
+    expect(agentTools.getTool1?.description).toBe('First test tool');
+    expect(agentTools.getTool2?.name).toBe('getTool2');
+    expect(agentTools.getTool2?.description).toBe('Second test tool');
+  });
+
+  it('should return a copy of tools, not the original reference', () => {
```
Shubhrakanti, line 62: These tests feel really unnecessary. I think we should only have tests for edge cases or describe core behavior of a class. @theomonnom @lukasIO would be good to hear to perspective on tests.

agents/src/voice/agent.ts
```diff
@@ -23,33 +30,42 @@ export class StopResponse extends Error {
...
+  tools?: ToolContext;
+  turnDetection?: TurnDetectionMode;
+  stt?: STT;
+  vad?: VAD;
+  llm?: LLM; // TODO: support realtime model
+  tts?: TTS;
+  mcpServers?: any[]; // TODO: support MCP servers
+  allowInterruptions?: boolean;
+  minConsecutiveSpeechDelay?: number;
+}
```
Shubhrakanti, line 45: I like this, but let's try to do this in a separate PR (next time) - esp when this one is getting so big. PRs should aim to do one logical unit of work.

Shubhrakanti: In the future the whole restricting go the Agent class parameters should be one PR - it's just code clean up and has nothing to do with tool calling.

agents/src/voice/agent.ts
```diff
@@ -69,13 +84,18 @@ export class Agent {
...
   }
 
   get instructions(): string {
     return this._instructions;
   }
 
+  get tools(): ToolContext {
+    return { ...this._tools };
+  }
+
```
Shubhrakanti, line 98: this the only real `tool` related change for this file.

examples/src/basic_agent.ts
```diff
@@ -20,9 +20,10 @@ export default defineAgent({
     proc.userData.vad = await silero.VAD.load();
   },
   entry: async (ctx: JobContext) => {
-    const agent = new voice.Agent(
-      "You are a helpful assistant, you can hear the user's message and respond to it.",
-    );
+    const agent = new voice.Agent({
+      instructions:
+        "You are a helpful assistant, you can hear the user's message and respond to it.",
+    });
```
Shubhrakanti, line 26: Did the linter do this?

agents/src/llm/provider_format/openai.test.ts
```diff
@@ -485,7 +485,7 @@ describe('toChatCtx', () => {
         {
           type: 'audio_content',
           frame: [],
-        } as any,
+        } as any, // eslint-disable-line @typescript-eslint/no-explicit-any
```
Shubhrakanti: why do we have `any` in the first place?

Shubhrakanti (commented):
The title of the PR is "add-tool-class-generic-tool", but it's doing 3 different things 

> Defined new tool interfaces modeled similar to lk-agent v1.0, Vercel, and Mastra’s tool interface.
Enabled agent definitions using the new tool interfaces (definition only for now—execution logic will follow in upcoming PRs).
Refactored parts of the existing implementation from the old function context to the new tool context.


+ refactors the `Agent.ts` to use options as input.
+ renaming function -> tool

I think with this PR it's _okay_, since it won't break the dev branch. But in the future it's best to keep PRs small. If we introduce a bug - the first thing we want to do is revert - so that no one else is blocked. the smaller the PR the easier that becomes. 

It's also easier to review when a PR does just 1-3 things.

agents/src/voice/agent.test.ts
toubatbrian: The lk-agent python package returns a copy of the tools, which is why I think it’s important to explicitly list this test.

examples/src/basic_agent.ts
toubatbrian: Linter cannot recognize this, I manually changed those lines

agents/src/llm/provider_format/openai.test.ts
toubatbrian: This should not have linting issue without the comment, I just removed it

examples/src/basic_agent.ts
Shubhrakanti: Ah I mistook this for a different issue I had earlier.

agents/src/voice/agent.test.ts
Shubhrakanti: I guess no second thought it's not a bad idea to have more tests. No real downside.

agents/src/voice/agent.ts
toubatbrian: Yep, I wouldn't involve anything else for now, just trying to keep this PR size manageable.

Shubhrakanti (approved):

agents/src/voice/agent.test.ts
toubatbrian: I guess one reason behind not returning original reference is to prevent user editing the mutable tool list during runtime, which might create undefined behavior

agents/src/voice/agent.ts
```diff
@@ -23,33 +30,42 @@ export class StopResponse extends Error {
...
 }
 
+export interface AgentOptions {
+  instructions: string;
+  chatCtx?: ChatContext;
+  tools?: ToolContext;
+  turnDetection?: TurnDetectionMode;
+  stt?: STT;
+  vad?: VAD;
+  llm?: LLM; // TODO: support realtime model
```
Shubhrakanti, line 40: Can we add ticket numbers to this? (Ex. https://linear.app/livekit/issue/AJS-32/add-support-for-realtime-model)

PR #440, opened 2025-06-12
brianyin/ajs-97-refactor-chatcontext-to-lk-agent-v1

toubatbrian:


agents/src/llm/misc.ts
```diff
@@ -0,0 +1,5 @@
+import { v4 as uuidv4 } from 'uuid';
+
+export function shortuuid(prefix: string): string {
+  return `${prefix}_${uuidv4().slice(0, 12)}`;
+}
```
Shubhrakanti: ~~Let's do this in a separate PR. I have a ticket for this. (AJS-60)~~

agents/src/llm/provider_format/openai.test.ts
```diff
@@ -0,0 +1,577 @@
...
+import { serializeImage } from '../utils.js';
+import { toChatCtx } from './openai.js';
+
+// Mock the serializeImage function
+vi.mock('../utils.js', () => ({
+  serializeImage: vi.fn(),
+}));
+
+describe('toChatCtx', () => {
+  const serializeImageMock = vi.mocked(serializeImage);
```
Shubhrakanti, line 17: Personally I'm a bigger fan of having fewer tests that test really the edge cases and explain non intuitive behavior - what do you think?

Shubhrakanti (approved):
This PR is over 2000 lines of code making it pretty hard to review.


I just took a brief glance and seems fine.

Shubhrakanti:
>  The changes appear quite large, but the actual impact is smaller than it seems.
> 
> Most of the updates are related to refactoring, such as method call adjustments, which is either 0% or 100%. A significant portion of the diff consists of unit tests. The actual new code is minimal and primarily introduces a new approach for converting the LiveKit 1.0 chat context to the OpenAI format.
> 

My recommendation is then you make one PR with all the functional changes and a follow up PR that actually implements those changes/add tests. Or at least add a comment highlighting the functional changes. Otherwise the reviewer has to drudge through all the code to figure out what the core changes are.

agents/src/llm/misc.ts
Shubhrakanti: I think it's fine for now, but can you add a comment on this function saying `//TODO(AJS-60) refactor all calls to randomUUID to use this`

agents/src/llm/utils.test.ts
```diff
@@ -0,0 +1,463 @@
...
+      // Decode and verify the pattern
+      const { decodedImage } = await decodeImageToRaw(result.base64Data!);
+      const decoded = decodedImage.data;
+
+      // Check each pixel - PNG is lossless so values should be exact
+      expectPixel(decoded, 0, { r: 255, g: 0, b: 0, a: 255 }); // Top-left (red)
+      expectPixel(decoded, 4, { r: 0, g: 255, b: 0, a: 255 }); // Top-right (green)
+      expectPixel(decoded, width * 4, { r: 0, g: 0, b: 255, a: 255 }); // Bottom-left (blue)
+      expectPixel(decoded, (width + 1) * 4, { r: 255, g: 255, b: 255, a: 255 }); // Bottom-right (white)
+    });
```
Shubhrakanti: It's fine to have AI generate the tests, but it tends to have overly verbose comments. I think as you go through the tests that it writes you can go thoughts and remove them.

agents/src/voice/audio_recognition.ts
```diff
@@ -190,7 +190,7 @@ export class AudioRecognition {
     }
 
     chatCtx = chatCtx.copy();
-    chatCtx.append({ role: ChatRole.USER, text: this.audioTranscript });
+    chatCtx.addMessage({ role: 'user', content: this.audioTranscript });
```
Shubhrakanti, line 193: I think the enum is better? I guess the python side uses raw strings so we're trying to keep it the same?

I would at least add a follow up ticket to think about making this an enum.

package.json
```diff
@@ -40,11 +41,11 @@
     "eslint-plugin-standard": "^5.0.0",
     "eslint-plugin-tsdoc": "^0.2.17",
     "prettier": "^3.2.5",
+    "tsup": "^8.4.0",
     "turbo": "^1.13.3",
     "typedoc": "^0.25.13",
     "typescript": "^5.4.5",
-    "vitest": "^3.2.2",
-    "tsup": "^8.4.0"
+    "vitest": "^3.2.2"
```
Shubhrakanti, line 48: no-op same with `"@types/node": "^22.13.10",`. Did prettier move these lines?

Either way, it's good to go through the PR and minimize no-op line changes if possible or comment why they happened.

agents/src/llm/provider_format/openai.test.ts
toubatbrian: Agree. These tests are generally test the core path & normal cases. I'm putting this test file because this function is part of the critical path of the openai's `toChatCtx` function. 

We should definitely make test sets smaller and add edge case testing only when necessary in the future

package.json
toubatbrian: I guess when I install a new package by pnpm install, it automatically re-format the order of dependency. Will make sure I make a comment on these details in the following PRs.

agents/src/llm/provider_format/openai.test.ts
Shubhrakanti: Yeah. Honestly I should be writing more tests too. helps explain behavior.

package.json
Shubhrakanti: Sounds good. Main thing I look for is making sure there are no version changes in existing packages that I should be aware of.

agents/src/llm/utils.test.ts
toubatbrian: Sounds good! btw, I’ve also reviewed all the AI-generated tests, they should cover the intended behaviors.

agents/src/voice/audio_recognition.ts
toubatbrian: The strings here are type-safe too, in case that’s a main concern. I feel like using strings makes more sense, since it allows users to construct the message payload just like they would when passing arguments to OpenAI’s chat completion calls—no need to import anything extra.

Shubhrakanti: Ohhh I see. Yeah let's def keep it a string here.

agents/src/llm/provider_format/index.ts
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: 2025 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import type { ChatContext } from '../chat_context.js';
+import { toChatCtx as toChatCtxOpenai } from './openai.js';
+
+export type ProviderFormat = 'openai';
+
+export async function toChatCtx(
+  format: ProviderFormat,
```
Shubhrakanti, line 10: @toubatbrian - this can just go in `/llm/utils.ts` no? misc and utils might be confusing.

PR #445, opened 2025-06-16
Pull up `TextOutput` and `AudioOutput` into abstract classes users can inherit. 

Shubhrakanti:
Pulling up `TextOutput` and `AudioOutput` into public interfaces users can inherit. This will also make Future PRs easier to read since we already define the base classes.

**There are not functional changes in this PR**

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,11 +17,17 @@
   "rules": {
     "tsdoc/syntax": "warn",
     "space-before-function-parens": 0,
-    "@typescript-eslint/no-unused-vars": "error",
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      {
+        "argsIgnorePattern": "^_"
+      }
+    ],
     "import/export": 0,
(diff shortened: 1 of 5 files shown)


.eslintrc
```diff
@@ -17,11 +17,17 @@
   "rules": {
     "tsdoc/syntax": "warn",
     "space-before-function-parens": 0,
-    "@typescript-eslint/no-unused-vars": "error",
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      {
+        "argsIgnorePattern": "^_"
+      }
+    ],
```
Shubhrakanti, line 25: In TypeScript, abstract methods cannot have implementations. So for `AudioOutput.captureFrame`, we never actually use the parameter. I think it's okay to have unused parameters if we explicitly flag them with an `_`

Shubhrakanti: https://johnnyreilly.com/typescript-eslint-no-unused-vars

PR #449, opened 2025-06-17
fix: use slice() instead of subarray() to correctly copy VAD audio buffer

wuhao-ouyang:
The previous implementation used `subarray()` which returns a view of the original buffer, causing the copied data to be inadvertently overwritten during subsequent processing.

Although a past fix removed the padding from the copied data to mitigate the issue, this made the `prefixPaddingDuration` parameter ineffective.

This commit replaces `subarray()` with `slice()` to ensure a proper copy is made, preventing unwanted side effects and restoring correct behavior of prefix padding.

### DIFF
--- a/plugins/silero/src/vad.ts
+++ b/plugins/silero/src/vad.ts
@@ -209,7 +209,7 @@ export class VADStream extends baseStream {
 
           // convert data to f32
           inferenceData = Float32Array.from(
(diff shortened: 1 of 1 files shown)


MatheusLealFSAI:
What is the livekit team waiting to approve this PR? I've gotten this issue in my application which small words like "John" or "Chad" were being clipped, I've implemented this solution as a patch in my application and this problem has been fixed. Thank you @wuhao-ouyang

PR #453, opened 2025-06-19
brianyin/ajs-114-tool-execution-code-path

toubatbrian:
This PR primarily focus on tool execution code path.

### DIFF
--- a/agents/src/llm/utils.ts
+++ b/agents/src/llm/utils.ts
@@ -255,3 +255,10 @@ export async function executeToolCall(
     });
   }
 }
+
+export function toError(error: unknown): Error {
+  if (error instanceof Error) {
+    return error;
+  }
+  return new Error(String(error));
+}
--- a/agents/src/voice/generation.ts
+++ b/agents/src/voice/generation.ts
@@ -3,17 +3,23 @@
 // SPDX-License-Identifier: Apache-2.0
 import type { AudioFrame } from '@livekit/rtc-node';
 import type { ReadableStream, ReadableStreamDefaultReader } from 'stream/web';
-import { FunctionCall, type ChatContext } from '../llm/chat_context.js';
+import { type ChatContext, FunctionCall } from '../llm/chat_context.js';
 import type { ChatChunk } from '../llm/llm.js';
 import { shortuuid } from '../llm/misc.js';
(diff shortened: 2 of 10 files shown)


agents/src/llm/utils.ts
```diff
@@ -255,3 +255,10 @@ export async function executeToolCall(
     });
   }
 }
+
+export function toError(error: unknown): Error {
+  if (error instanceof Error) {
```
toubatbrian, line 260: Any better measurement for this? @lukasIO @Shubhrakanti

agents/src/voice/generation.ts
```diff
@@ -251,3 +307,179 @@ export function performAudioForwarding(
...
+              function: toolCall.name,
+              speech_id: speechHandle.id,
+            },
+            'exception occurred while executing tool',
+          );
+          jsOut.exception = toError(rawError);
+        } finally {
+          toolOutput.output.push(jsOut);
+        }
+      };
```
toubatbrian, line 424: The reason I made an `waitUntilAborted` method and use this inside tool execution task is because in python there's a `task.add_done_callback` function which executed after task been completed, regardless if task is cancelled. I use this to simulate the behavior so that even if task is aborted, we can still post-process tool output non-interrupted by the abort signal.

@Shubhrakanti @lukasIO happy to discuss any better alternatives

agents/src/llm/utils.ts
Shubhrakanti: Yeah I'm not sure there is but this is really annoying.

agents/src/voice/generation.ts
```diff
@@ -3,21 +3,63 @@
...
+
+  constructor(toolCall: FunctionCall, output: any, exception: Error | undefined) {
+    this.toolCall = toolCall;
+    this.output = output;
+    this.exception = exception;
+  }
+
+  static create(params: {
+    toolCall: FunctionCall;
+    output?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
Shubhrakanti: unknown?

Shubhrakanti (approved):

agents/src/llm/utils.ts
lukasIO: I don't know a better way to handle this... 
Looks like a decent workaround

PR #454, opened 2025-06-20
[DO NOT MERGE NOW] brianyin/ajs-115-tool-calling-examples

toubatbrian:


agents/src/llm/provider_format/openai.ts
```diff
@@ -34,6 +35,8 @@ export async function toChatCtx(chatCtx: ChatContext, injectDummyUserMessage: bo
     }
   }
 
+  log().debug({ messages }, 'toChatCtx');
+
```
Shubhrakanti: we need to make this  a class variable or we have weird issues. Not sure of the reason why but I remember running into them.

Shubhrakanti (approved):

agents/src/llm/provider_format/openai.ts
toubatbrian: Ignore the debug logs for now — I’ll clean them up before merging. They’re just there to support the stacked PRs.

agents/src/voice/generation.ts
```diff
@@ -111,6 +114,17 @@ export function performLLMInference(
...
+                {
+                  function: tool.name,
+                  call_id: toolCall.callId,
+                  args: tool.args,
+                  tool_count: data.generatedToolCalls.length + 1,
+                },
+                'performLLMInference: LLM generated tool call',
+              );
+
               data.generatedToolCalls.push(toolCall);
```
toubatbrian: Will clean this up.

plugins/openai/src/llm.ts
```diff
@@ -479,6 +479,17 @@ export class LLMStream extends llm.LLMStream {
...
+        tools,
+        opts,
+        n,
+        parallelToolCalls,
+        temperature,
+      },
+      'tools',
+    );
+
     try {
```
toubatbrian: ^

PR #461, opened 2025-06-23
brianyin/ajs-94-agent-handoff-function

toubatbrian:


agents/src/voice/audio_recognition.ts
```diff
@@ -415,6 +415,13 @@ export class AudioRecognition {
       });
   }
 
+  async aclose() {
```
Shubhrakanti, line 418: I usually just call it `close` since JS is async native, but this is fine too.

agents/src/voice/generation.ts
```diff
@@ -402,7 +456,7 @@ export function performToolExecutions({
     const signal = controller.signal;
     const reader = toolCallStream.getReader();
 
-    const tasks: Promise<any>[] = [];
+    const tasks: Promise<any>[] = []; // eslint-disable-line @typescript-eslint/no-explicit-any
```
Shubhrakanti, line 459: `unknown` over `any`? (I liked your usage of `catch (error: unknown)`.

agents/src/voice/agent_activity.ts
```diff
@@ -193,16 +243,26 @@ export class AgentActivity implements RecognitionHooks {
...
   }
 
-  private async mainTask(): Promise<void> {
+  private async mainTask(signal: AbortSignal): Promise<void> {
     while (true) {
+      if (signal.aborted) {
+        break;
+      }
+
       await this.q_updated.await;
```
Shubhrakanti, line 266: don't we need to race the `this.q_updated.await;` with the abort signal?

agents/src/voice/agent_activity.ts
```diff
@@ -47,9 +48,16 @@ export class AgentActivity implements RecognitionHooks {
...
   private q_updated: Future;
+  private lock = new Mutex();
+  private speechTasks: Set<Promise<void>> = new Set();
+
   agent: Agent;
   agentSession: AgentSession;
 
+  /** @internal */
+  _mainTask?: Task<void>;
+  _userTurnCompletedTask?: Task<void>;
```
Shubhrakanti, line 59: If we don't use `cancelAndAwait` might be better to just use a `Promise<void>`?

PR #463, opened 2025-06-24
Run test on PRs with changes in workflow files

Shubhrakanti:
https://github.com/livekit/agents-js/pull/448 missed adding an API key to one of the configs. This error was muted because tests weren't running on the PRs when `workflow` files were edited.


Shubhrakanti:
Looking at https://github.com/livekit/agents-js/actions/runs/15856358839/job/44702587230?pr=463 we can see the tests actually ran this time.

PR #467, opened 2025-06-25
brianyin/ajs-91-agent-activity-draining

toubatbrian:


agents/src/vad.ts
```diff
@@ -220,7 +224,15 @@ export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
...
-    this.input.writable.close();
+    this.logger.debug('VAD stream closing');
+    this.logger.debug('VAD stream releasing output writer lock');
+    this.outputWriter.releaseLock();
+    this.logger.debug('VAD stream cancelling output reader');
+    this.outputReader.cancel();
+    this.logger.debug('VAD stream closing output stream');
+    this.output.writable.close();
+    // this.logger.debug('VAD stream cancelling output readable');
+    // this.output.readable.cancel();
```
Shubhrakanti: why is this commented out?

agents/src/voice/agent_session.ts
```diff
@@ -201,10 +201,10 @@ export class AgentSession<
...
     this.nextActivity = new AgentActivity(agent, this);
 
-    // if (this.activity) {
-    //   await this.activity.drain();
-    //   await this.activity.aclose();
-    // }
+    if (this.activity) {
+      await this.activity.drain();
+      await this.activity.close();
+    }
```
Shubhrakanti, line 207: Can you make a ticket ticket (or do it in this PR) to check draining in `say` and `generateReply`?

agents/src/voice/audio_recognition.ts
```diff
@@ -311,49 +310,69 @@ export class AudioRecognition {
   }
 
   private createVadTask() {
-    return async (controller: AbortController) => {
+    return async ({ signal }: AbortController) => {
```
Shubhrakanti: Why not just `signal: AbortSignal`

agents/src/voice/audio_recognition.ts
```diff
@@ -311,49 +310,69 @@ export class AudioRecognition {
...
-            this.speaking = true;
+      const abortHandler = () => {
+        this.logger.debug('VAD task aborted, detaching input stream');
+        vadStream.detachInputStream();
+        this.logger.debug('VAD task aborted, closing vad stream');
+        vadStream.close();
+        this.logger.debug('VAD task aborted, removing abort handler');
+        signal.removeEventListener('abort', abortHandler);
+      };
+      signal.addEventListener('abort', abortHandler);
```
Shubhrakanti: can we just have one log here? Do we really need all 3? 

It's okay for now if we really need it, but we should def clean up before release

Shubhrakanti (approved):

agents/src/voice/audio_recognition.ts
```diff
@@ -311,49 +310,63 @@ export class AudioRecognition {
...
   setInputAudioStream(audioStream: ReadableStream<AudioFrame>) {
-    const mergedStream = mergeReadableStreams(
-      audioStream as any,
-      this.silenceAudioTransform.readable as any,
-    );
-    this.deferredInputStream.setSource(mergedStream as any);
+    // const mergedStream = mergeReadableStreams(
+    //   audioStream as any,
+    //   this.silenceAudioTransform.readable as any,
+    // );
```
Shubhrakanti, line 354: why is this commented out?

Shubhrakanti:
Nice job making the PR small makes it easy to review

agents/src/voice/audio_recognition.ts
toubatbrian: We can, it should be the same. But the outer function directly pass the controller I have to deconstruct like this

agents/src/voice/audio_recognition.ts
toubatbrian: I still need to work on this. Basically if we merge silence stream with the input audio stream. We can't un-merge it, which makes the hand off draining failed due to the locking behavior for merged stream. I need to come up with another way to handle this merge

agents/src/voice/audio_recognition.ts
Shubhrakanti: I prefer to keep it consistent in the code base and `signal: AbortSignal` is just way more obvious. If the function only needs a signal, it should only take a signal. 

Small nit though - don't care too much.

agents/src/voice/audio_recognition.ts
Shubhrakanti: Makes sense. Can you add a detailed comment before requesting for review? This is not at all obvious to the reviewer.

PR #468, opened 2025-06-25
Shubhra/ajs 64 add say to agent

Shubhrakanti:
User can specify the agent to say a specific string.

### DIFF
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -133,27 +133,20 @@ export class AgentActivity implements RecognitionHooks {
   say(
     text: string | ReadableStream<string>,
     options?: {
-      audio?: ReadableStream<AudioFrame> | null;
-      allowInterruptions?: boolean | null;
+      audio?: ReadableStream<AudioFrame>;
+      allowInterruptions?: boolean;
       addToChatCtx?: boolean;
     },
   ): SpeechHandle {
-    if (options === undefined) {
-      options = {};
-    }
-    if (options.addToChatCtx === undefined) {
-      options.addToChatCtx = true;
-    }
+    const { audio, allowInterruptions, addToChatCtx = true } = options ?? {};
 
-    if (!options.audio && !this.tts && this.agentSession.audioOutput) {
+    if (!audio && !this.tts && this.agentSession.audioOutput) {
       throw new Error('trying to generate speech from text without a TTS model');
     }
 
-    const handle = SpeechHandle.create(options.allowInterruptions ?? this.allowInterruptions);
+    const handle = SpeechHandle.create(allowInterruptions ?? this.allowInterruptions);
 
-    this.logger.debug({ text, options }, 'say in agent activity');
-
-    this.ttsTask(handle, text, options.addToChatCtx, options.audio).then(() => {
+    this.ttsTask(handle, text, addToChatCtx, audio).then(() => {
       this.onPipelineReplyDone();
     });
     this.scheduleSpeech(handle, SpeechHandle.SPEECH_PRIORITY_NORMAL);
@@ -355,8 +348,6 @@ export class AgentActivity implements RecognitionHooks {
     addToChatCtx: boolean,
     audio?: ReadableStream<AudioFrame> | null,
   ): Promise<void> {
-    this.logger.debug({ speechHandle, text, addToChatCtx, audio }, 'ttsTask');
-
     const transcriptionOutput = this.agentSession._transcriptionOutput;
     const audioOutput = this.agentSession.audioOutput;
 
@@ -442,7 +433,6 @@ export class AgentActivity implements RecognitionHooks {
     }
 
     if (speechHandle.interrupted) {
-      this.logger.debug('ttsTask: speech handle interrupted');
       replyAbortController.abort();
       await Promise.allSettled(
         tasks.map((task) => task.cancelAndWait(AgentActivity.REPLY_TASK_CANCEL_TIMEOUT)),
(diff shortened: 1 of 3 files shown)


agents/src/voice/agent_activity.ts
```diff
@@ -130,6 +130,36 @@ export class AgentActivity implements RecognitionHooks {
     this.audioRecognition.clearUserTurn();
   }
 
+  say(
+    text: string | ReadableStream<string>,
+    options?: {
+      audio?: ReadableStream<AudioFrame> | null;
+      allowInterruptions?: boolean | null;
+      addToChatCtx?: boolean;
+    },
```
toubatbrian, line 139:
```tsx
    options?: {
      audio?: ReadableStream<AudioFrame>;
      allowInterruptions?: boolean;
      addToChatCtx?: boolean;
    },
```

agents/src/voice/agent_activity.ts
```diff
@@ -130,6 +130,36 @@ export class AgentActivity implements RecognitionHooks {
...
+  ): SpeechHandle {
+    if (options === undefined) {
```
toubatbrian:
```tsx
const { audio, allowInterruptions, addToChatCtx = true } = options;
```
to assign default value

agents/src/voice/agent_activity.ts
Shubhrakanti: Ahh this is to differentiate between the user leaving it undefined and explicitly setting it to null. Look for `NOT_GIVEN_OR` in python.

toubatbrian: We should try avoid using `null` anytime since it's quite annoying to deal with in JS for some cases.  https://medium.com/@hbarcelos/why-i-banned-null-from-my-js-code-and-why-you-should-too-13df90323cfa

toubatbrian: To represent NOT_GIVEN we can simply make it type to be `T | undefined` or in this case `variable?: T`. 
When user are not specifying this argument, it is by default undefined which is NOT_GIVEN

Shubhrakanti: We want three separate state with `NOT_GIVEN`. 1. the user didn't specify, 2. it's explicitly set to null and 3. it's given by the user.

agents/src/voice/agent_activity.ts
```diff
@@ -130,6 +130,36 @@ export class AgentActivity implements RecognitionHooks {
...
+      options.addToChatCtx = true;
+    }
+
+    if (!options.audio && !this.tts && this.agentSession.audioOutput) {
+      throw new Error('trying to generate speech from text without a TTS model');
+    }
+
+    const handle = SpeechHandle.create(options.allowInterruptions ?? this.allowInterruptions);
+
+    this.logger.debug({ text, options }, 'say in agent activity');
```
toubatbrian: maybe clean this before merge?

Shubhrakanti: Yeah I didn't request for review yet. Still need to clean this up.

agents/src/voice/agent_activity.ts
toubatbrian: Discussed offline with @Shubhrakanti, we will stick to use `T | undefined` for those only have two possibility: NOT_GIVEN or a concrete value.

agents/src/voice/agent_activity.ts
```diff
@@ -130,6 +130,29 @@ export class AgentActivity implements RecognitionHooks {
...
+
+    if (!audio && !this.tts && this.agentSession.audioOutput) {
+      throw new Error('trying to generate speech from text without a TTS model');
+    }
+
+    const handle = SpeechHandle.create(allowInterruptions ?? this.allowInterruptions);
+
+    this.ttsTask(handle, text, addToChatCtx, audio).then(() => {
+      this.onPipelineReplyDone();
+    });
```
toubatbrian, line 151: suggest to use `.finally` in case error throwed inside ttsTask

Shubhrakanti: we should also do this for `generateReply` good catch

agents/src/voice/agent_activity.ts
```diff
@@ -215,8 +238,8 @@ export class AgentActivity implements RecognitionHooks {
         speechHandle.authorizePlayout();
         await speechHandle.waitForPlayout();
         this.currentSpeech = undefined;
-        this.q_updated = new Future();
       }
+      this.q_updated = new Future();
```
toubatbrian: nice catch!

Shubhrakanti: Hmm this shouldn't be there. Was fixed in #469

agents/src/voice/agent_activity.ts
```diff
@@ -319,6 +342,126 @@ export class AgentActivity implements RecognitionHooks {
...
+      } else {
+        // use the provided audio
+        const [forwardTask, _audioOut] = performAudioForwarding(
+          audio,
+          audioOutput,
+          replyAbortController,
+        );
+        tasks.push(forwardTask);
+        audioOut = _audioOut;
+      }
```
toubatbrian, line 425: I noticed there’s quite a bit of overlapping code with the pipeline reply task. Since πython have the same setup, let’s just go ahead do this, but would be better make a ticket to merge some logics during polishing phase.

toubatbrian (approved):

agents/src/voice/agent_activity.ts
Shubhrakanti: yeah I had the same thought. TBH we should do this like dead last. Make sure _everything_ is perfect and then come touch this stuff. There's also a log of logic in the audio and text stream forwarding that can be a lot cleaner with streams I want to refactor.

agents/src/voice/agent_activity.ts
Shubhrakanti: I'll do the rest in a separate PR.

PR #478, opened 2025-06-25
Fix TTS Hanging when Receiving empty text stream

toubatbrian:
Issue is that when perform tool calling, the text stream will be a empty stream. This won't trigger `pushText` method in ts's `SynthesizeStream` object. However, pushText lazily starts `monitorMetrics` task, which controlling the close lifecycle stage of the tts stream. Without any text chunk, we cannot start `monitorMetrics` which make the TTS stream unable to close so it become hanging. 

This PR separate the close outside of monitorMetric task so it call `this.output.close()` as soon as the main task finishes

### DIFF
--- a/agents/src/tts/tts.ts
+++ b/agents/src/tts/tts.ts
@@ -157,6 +157,11 @@ export abstract class SynthesizeStream
       this.logger.error(error, 'Error reading deferred input stream');
     } finally {
       reader.releaseLock();
+      // Ensure output is closed when the stream ends
+      if (!this.#monitorMetricsTask) {
+        // No text was received, close the output directly
+        this.output.close();
+      }
     }
   }
 
@@ -204,7 +209,6 @@ export abstract class SynthesizeStream
     if (requestId) {
       emit();
     }
-    this.output.close();
   }
 
   updateInputStream(text: ReadableStream<string>) {
@@ -216,6 +220,8 @@ export abstract class SynthesizeStream
   pushText(text: string) {
     if (!this.#monitorMetricsTask) {
       this.#monitorMetricsTask = this.monitorMetrics();
+      // Close output when metrics task completes
+      this.#monitorMetricsTask.finally(() => this.output.close());
     }
     this.#metricsText += text;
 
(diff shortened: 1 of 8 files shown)


examples/src/basic_tool_call_agent.ts
```diff
@@ -34,9 +34,9 @@ export default defineAgent({
         location: z.string().describe('The location to get the weather for'),
       }),
       execute: async ({ location }) => {
-        if (Math.random() < 0.5) {
-          throw new llm.ToolError('Internal server error, please try again later.');
-        }
+        // if (Math.random() < 0.5) {
+        //   throw new llm.ToolError('Internal server error, please try again later.');
+        // }
```
Shubhrakanti, line 39: we should clean these up before sharing with the team on Monday

agents/src/utils.ts
```diff
@@ -355,19 +359,35 @@ export class Task<T> {
...
   }
 
   private async runTask() {
-    return this.fn(this.controller)
+    const run = async () => {
+      if (this.name) {
+        this.#logger.debug(`Task.runTask: task ${this.name} started`);
+      }
+      return await this.fn(this.controller);
+    };
```
Shubhrakanti, line 377: amazing

agents/src/tts/tts.ts
```diff
@@ -157,6 +157,11 @@ export abstract class SynthesizeStream
       this.logger.error(error, 'Error reading deferred input stream');
     } finally {
       reader.releaseLock();
+      // Ensure output is closed when the stream ends
+      if (!this.#monitorMetricsTask) {
+        // No text was received, close the output directly
+        this.output.close();
+      }
```
Shubhrakanti, line 164: this seems a little flaky?

toubatbrian: I agree, I'll add a ticket and come back during polishing phase. Right now the functionality works correctly (confirmed with extensive trials of debug loggings)

Shubhrakanti: you can just add TODO AJS-30 refactor with streams. look around that file there should be there one already

PR #481, opened 2025-06-26
Shubhra/ajs 50 update speech handle to 10 framework

Shubhrakanti:
Minor refactoring and removing unused code.

### DIFF
--- a/agents/src/voice/speech_handle.ts
+++ b/agents/src/voice/speech_handle.ts
@@ -1,11 +1,9 @@
 // SPDX-FileCopyrightText: 2024 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import type { ChatMessage } from '../llm/index.js';
 import { shortuuid } from '../llm/misc.js';
 import { Future } from '../utils.js';
 
-// TODO(AJS-50): Update speech handle to 1.0
 export class SpeechHandle {
   /** Priority for messages that should be played after all other messages in the queue */
   static SPEECH_PRIORITY_LOW = 0;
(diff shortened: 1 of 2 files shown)


agents/src/voice/speech_handle.ts
```diff
@@ -14,58 +13,56 @@ export class SpeechHandle {
...
 
-  static create(allowInterruptions: boolean = false, stepIndex: number = 0, parent?: SpeechHandle) {
-    return new SpeechHandle(shortuuid('speech'), allowInterruptions, stepIndex, parent);
-  }
+  constructor(
+    readonly id: string,
+    readonly allowInterruptions: boolean,
+    readonly stepIndex: number,
+    readonly parent?: SpeechHandle,
+  ) {}
```
Shubhrakanti, line 27: We _could_ do something like 

```typescript
class SpeechHandle {
  private _id: string;
  private _allowInterruptions: boolean;
  private _stepIndex: number;
  private _parent?: SpeechHandle;

  constructor(
    id: string,
    allowInterruptions: boolean,
    stepIndex: number,
    parent?: SpeechHandle,
  ) {
    this._id = id;
    this._allowInterruptions = allowInterruptions;
    this._stepIndex = stepIndex;
    this._parent = parent;
  }

  get id(): string { return this._id; }
  get allowInterruptions(): boolean { return this._allowInterruptions; }
  get stepIndex(): number { return this._stepIndex; }
  get parent(): SpeechHandle | undefined { return this._parent; }
}
```

But that's more verbose and tbh I don't see the value right now. If we ever need to change it it's fine because the usage syntax won't change.

agents/src/voice/speech_handle.ts
```diff
@@ -14,58 +13,56 @@ export class SpeechHandle {
...
-    return this.#id;
-  }
+  static create(
+    options: {
+      allowInterruptions?: boolean;
+      stepIndex?: number;
+      parent?: SpeechHandle;
+    } = {},
+  ) {
+    const { allowInterruptions = false, stepIndex = 0, parent } = options;
```
toubatbrian: I would suggest
```tsx
  static create(
    options?: {
      allowInterruptions?: boolean;
      stepIndex?: number;
      parent?: SpeechHandle;
    },
  ) {
    const { allowInterruptions = false, stepIndex = 0, parent } = options ?? {};
```
so if user don't specify any of the parameter they can just call `SpeechHandle.create()`

Shubhrakanti: oh snap good point.

agents/src/voice/speech_handle.ts
Shubhrakanti: @lukasIO @theomonnom would be good to get our thoughts too since `SpeechHandle` is a public facing interface.

PR #483, opened 2025-06-26
Fix Merge Silence Audio Stream to Unblock Draining

toubatbrian:


agents/src/voice/audio_recognition.ts
```diff
@@ -348,11 +350,7 @@ export class AudioRecognition {
   }
 
   setInputAudioStream(audioStream: ReadableStream<AudioFrame>) {
-    // const mergedStream = mergeReadableStreams(
-    //   audioStream as any,
-    //   this.silenceAudioTransform.readable as any,
-    // );
-    this.deferredInputStream.setSource(audioStream as any);
```
toubatbrian, line 355: We wire audioStream directly to deferred stream, so that detaching deferred stream will release lock on upstream audioStream object itself

agents/src/voice/audio_recognition.ts
```diff
@@ -76,7 +77,7 @@ export class AudioRecognition {
     this.deferredInputStream = new DeferredReadableStream<AudioFrame>();
     const [vadInputStream, sttInputStream] = this.deferredInputStream.stream.tee();
     this.vadInputStream = vadInputStream;
-    this.sttInputStream = sttInputStream;
+    this.sttInputStream = mergeReadableStreams(sttInputStream, this.silenceAudioTransform.readable);
```
toubatbrian, line 80: We merge the stream with the tee-ed version of incoming audioStream from deferred stream. This way, when we detach deferred stream. original input audioStream gets unlocked to allows transfer and wire to stt in a new agent activity

PR #484, opened 2025-06-26
Fix bug with speech duration calculation

Shubhrakanti:
some minor clean up as well

### DIFF
--- a/plugins/silero/src/vad.ts
+++ b/plugins/silero/src/vad.ts
@@ -35,9 +35,9 @@ export interface VADOptions {
 
 const defaultVADOptions: VADOptions = {
   minSpeechDuration: 50,
-  minSilenceDuration: 250,
+  minSilenceDuration: 550,
   prefixPaddingDuration: 500,
-  maxBufferedSpeech: 60000,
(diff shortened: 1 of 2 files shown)


plugins/silero/src/vad.ts
```diff
@@ -250,9 +250,9 @@ export class VADStream extends baseStream {
           }
 
           if (pubSpeaking) {
-            pubSpeechDuration += inferenceDuration;
+            pubSpeechDuration += windowDuration;
           } else {
-            pubSilenceDuration += inferenceDuration;
+            pubSilenceDuration += windowDuration;
```
Shubhrakanti, line 255: main fix

theomonnom (approved):

plugins/silero/src/vad.ts
toubatbrian: just curious, what breaks did the previous version cause?

Shubhrakanti: the `speechDuration` is incorrect so [this line](https://github.com/livekit/agents-js/pull/485/files#r2169954692) won't work right

PR #488, opened 2025-06-26
Prefix agentActivity with _

Shubhrakanti:
A little clearer to users it's internal use only.

### DIFF
--- a/agents/src/voice/agent.ts
+++ b/agents/src/voice/agent.ts
@@ -89,7 +89,7 @@ export class Agent<UserData = any> {
     this._vad = vad;
     this._llm = llm;
     this._tts = tts;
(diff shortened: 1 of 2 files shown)


theomonnom (approved):

agents/src/voice/agent.ts
```diff
@@ -89,7 +89,7 @@ export class Agent<UserData = any> {
     this._vad = vad;
     this._llm = llm;
     this._tts = tts;
-    this.agentActivity = undefined; // TODO(shubhra): add type
+    this._agentActivity = undefined; // TODO(shubhra): add type
```
Shubhrakanti: don't love setting something explicitly to undefined but that is for a different PR

PR #490, opened 2025-06-26
Add Misc Missing Components before Restaurant Agent Example

toubatbrian:


agents/src/llm/chat_context.ts
```diff
@@ -197,7 +197,7 @@ export class ChatContext {
     return new ChatContext([]);
   }
 
-  get items(): ReadonlyArray<ChatItem> {
+  get items(): ChatItem[] {
```
toubatbrian, line 200: We want to change `items` in place so it can't be readonly. lk python has another class called readonly chat context, we should adopt that approach as well. Will add a ticket

agents/src/llm/chat_context.ts
```diff
@@ -296,34 +296,20 @@ export class ChatContext {
...
-      if (first.type === 'function_call' || first.type === 'function_call_output') {
-        newItems.shift();
-      } else {
-        break;
-      }
+    while (
+      newItems.length > 0 &&
+      ['function_call', 'function_call_output'].includes(newItems[0]!.type)
+    ) {
+      newItems.shift();
```
toubatbrian, line 310: code simplification

agents/src/llm/chat_context.ts
```diff
@@ -296,34 +296,20 @@ export class ChatContext {
...
-      const item = newItems[i]!;
-      if (item.type !== 'function_call' && item.type !== 'function_call_output') {
-        nonFunctionItemIdx = i;
-        break;
-      }
-    }
-
-    // Trim chat ctx start on any function_call or function_call_output
-    newItems = newItems.slice(nonFunctionItemIdx);
-
```
toubatbrian, line 326: those are removed from latest python package

PR #494, opened 2025-06-27
Shubhra/ajs 41 remove hard coding of audio input sample rate

Shubhrakanti:
Add RoomIO input and output options. Will actually apply the flags in a later PR.

### DIFF
--- a/agents/src/voice/room_io/room_io.ts
+++ b/agents/src/voice/room_io/room_io.ts
@@ -29,11 +29,47 @@ import {
   ParticipantTranscriptionOutput,
 } from './_output.js';
 
+export interface RoomInputOptions {
+  audioSampleRate: number;
+  audioNumChannels: number;
+  textEnabled: boolean;
+  audioEnabled: boolean;
+  videoEnabled: boolean;
+  participantIdentity?: string;
+}
+
+export interface RoomOutputOptions {
+  transcriptionEnabled: boolean;
+  audioEnabled: boolean;
+  audioSampleRate: number;
+  audioNumChannels: number;
+  syncTranscription: boolean;
+  audioPublishOptions: TrackPublishOptions;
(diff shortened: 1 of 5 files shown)


examples/src/basic_tool_call_agent.ts
```diff
@@ -142,7 +142,10 @@ export default defineAgent({
       userData: { number: 0 },
     });
 
-    session.start(routerAgent, ctx.room);
+    session.start({
+      agent: routerAgent,
+      room: ctx.room,
+    });
```
Shubhrakanti, line 148: @toubatbrian feel like this is cleaner?

agents/src/voice/room_io/room_io.ts
```diff
@@ -47,24 +83,36 @@ export class RoomIO {
...
+    agentSession: AgentSession;
+    room: Room;
+    participant?: Participant;
+    inputOptions?: Partial<RoomInputOptions>;
+    outputOptions?: Partial<RoomOutputOptions>;
+  }) {
     this.agentSession = agentSession;
     this.room = room;
+    this.inputOptions = { ...DEFAULT_ROOM_INPUT_OPTIONS, ...inputOptions };
+    this.outputOptions = { ...DEFAULT_ROOM_OUTPUT_OPTIONS, ...outputOptions };
```
Shubhrakanti, line 102: we do this instead of `inputOptions = DEFAULT_ROOM_INPUT_OPTIONS` in the constructor because we want to override the defaults with any options specified by the user.

Shubhrakanti:
Also solve AJS-55

examples/src/basic_tool_call_agent.ts
toubatbrian: +1

PR #496, opened 2025-06-27
Fix LLM Instruction Initialization

toubatbrian:


agents/src/voice/agent.ts
```diff
@@ -127,8 +124,8 @@ export class Agent<UserData = any> {
     return { ...this._tools };
   }
 
-  get session(): AgentSession {
-    return this.getActivityOrThrow().agentSession;
+  get session(): AgentSession<UserData> {
+    return this.getActivityOrThrow().agentSession as AgentSession<UserData>;
```
toubatbrian, line 128: type-safety

agents/src/voice/agent_activity.ts
```diff
@@ -68,6 +68,17 @@ export class AgentActivity implements RecognitionHooks {
...
+
+    try {
+      updateInstructions({
+        chatCtx: this.agent._chatCtx,
+        instructions: this.agent.instructions,
+        addIfMissing: true,
+      });
+    } catch (error) {
+      this.logger.error('failed to update the instructions', error);
+    }
```
Shubhrakanti, line 80: why try catch? Do we want to continue if we fail to set the instructions?

Shubhrakanti (approved):

agents/src/voice/agent_activity.ts
toubatbrian: to follow python's code path. Inside update instruction method it could throw an error if the found message with id = INSTRUCTIONS_MESSAGE_ID is not a chat message, although this won't happen since we assign this id internally and we always assign it to system message, but in case user mess this up.

PR #501, opened 2025-06-30
Revert Audio Input Stream

toubatbrian:
This reverts commit 1636b99778050d1cc1f8ce4b8ebf936cc28c412d.

### DIFF
--- a/agents/src/voice/room_io/room_io.ts
+++ b/agents/src/voice/room_io/room_io.ts
@@ -1,10 +1,21 @@
 // SPDX-FileCopyrightText: 2025 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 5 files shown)


agents/src/voice/room_io/room_io.ts
```diff
@@ -13,7 +24,6 @@ import {
 } from '../agent_session.js';
 import type { AudioOutput, TextOutput } from '../io.js';
 import { TranscriptionSynchronizer } from '../transcription/synchronizer.js';
-import { ParticipantAudioInputStream } from './_input.js';
 import {
```
toubatbrian, line 27: I didn't delete the `_input.ts` file, only reverted changes in room_io and examples

toubatbrian: @Shubhrakanti

PR #502, opened 2025-07-01
brianyin/ajs-141-fix-duplicated-user-transcription

toubatbrian:


agents/src/voice/room_io/room_io.ts
```diff
@@ -164,13 +172,40 @@ export class RoomIO {
...
+
+        const event = value;
+        await this.userTranscriptOutput?.captureText(event.transcript);
+        if (event.isFinal) {
+          this.userTranscriptOutput?.flush();
+        }
+      }
+
+      reader.releaseLock();
+    } catch (error) {
```
toubatbrian, line 194: push transcribe event into an stream to make sure events are correctly ordered

agents/src/voice/room_io/room_io.ts
```diff
@@ -234,6 +269,11 @@ export class RoomIO {
       this.agentTranscriptOutput,
     );
 
+    // Start the transcript processing task
+    this.userTranscriptProcessingTask = Task.from(({ signal }) =>
+      this.processTranscriptStream(signal),
+    );
+
     // -- set the room event handlers --
```
toubatbrian, line 265: Passing abort signal here, tho we don't have method to cancel room io yet (not on python side either)

agents/src/voice/room_io/room_io.ts
```diff
@@ -164,13 +172,40 @@ export class RoomIO {
...
+    signal.addEventListener('abort', abortHandler);
+
+    const reader = this.userTranscriptStream.readable.getReader();
+    try {
+      while (!signal.aborted) {
+        const { done, value } = await Promise.race([reader.read(), abortFuture.await]);
+        if (done) break;
+
+        const event = value;
+        await this.userTranscriptOutput?.captureText(event.transcript);
```
toubatbrian, line 189: IMPORTANT, IMPORTANT, IMPORTANT!!!!!!!! Add `await` on `captureText`

lukasIO: maybe add an inline comment around this, too?

Shubhrakanti: +1 let's add a comment here

agents/src/voice/room_io/room_io.ts
```diff
@@ -83,6 +84,11 @@ export class RoomIO {
   private participantAvailableFuture: Future<Participant> = new Future();
   private roomConnectedFuture: Future<void> = new Future();
 
+  // Use stream API for transcript queue
+  private userTranscriptStream = new IdentityTransform<UserInputTranscribedEvent>();
```
Shubhrakanti, line 88: Why do we need this? We can just do this with an event listener in JS no?

agents/src/voice/room_io/room_io.ts
```diff
@@ -164,13 +172,40 @@ export class RoomIO {
...
 
   private onUserInputTranscribed = (ev: UserInputTranscribedEvent) => {
-    this.logger.debug({ ev }, 'user input transcribed');
-    this.userTranscriptOutput?.captureText(ev.transcript);
-    if (ev.isFinal) {
-      this.userTranscriptOutput?.flush();
-    }
+    this.userTranscriptWriter.write(ev).catch((error) => {
+      this.logger.error({ error }, 'Failed to write transcript event to stream');
+    });
```
Shubhrakanti, line 177: why do we need an explicit catch?

agents/src/voice/room_io/room_io.ts
Shubhrakanti: then let's keep it a promise? In general if we don't need something let's not add it in. Especially if python side doesn't use it.

Shubhrakanti: When we read the code and see a `Task` we usually assumed it's going to be cancelled. Would be confusing if it's not.

Shubhrakanti (approved):
Nice catch. I forgot to buffer the transcription data

agents/src/voice/room_io/room_io.ts
toubatbrian: just to make sure we properly catch any potential error

Shubhrakanti: Right the async task won't throw by default

PR #503, opened 2025-07-01
Revert "Revert Audio Input Stream (#501)"

Shubhrakanti:
This reverts commit bd96bbeb1144014bf88968660d9d328a8e02d1c8.

We also added in a bug fix done in #508 which caused the audio forwarding task to hang due to an incorrect future initialization in `ParticipantAudioOutputStream`

### DIFF
--- a/examples/src/restaurant_agent.ts
+++ b/examples/src/restaurant_agent.ts
@@ -164,8 +164,7 @@ class BaseAgent extends voice.Agent<UserData> {
     });
 
     await this.updateChatCtx(chatCtx);
-    // this.session.generateReply({ toolChoice: 'none' });
-    this.session.say("Hello, I'm a game agent. I can help you with your tasks.");
+    this.session.generateReply({ toolChoice: 'none' });
   }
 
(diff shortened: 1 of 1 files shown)


agents/src/voice/room_io/room_io.ts
```diff
@@ -141,6 +115,17 @@ export class RoomIO {
     await this.participantAudioOutput?.start();
   }
 
+  private onConnectionStateChanged = (state: ConnectionState) => {
+    this.logger.debug({ state }, 'connection state changed');
+    if (
+      state === ConnectionState.CONN_CONNECTED &&
+      this.room.isConnected &&
+      !this.roomConnectedFuture.done
+    ) {
```
Shubhrakanti, line 133: I _think_ this is what was causing the issue.

Shubhrakanti: Update this was not the issue. #508 was the issue

toubatbrian (approved):
LGTM!

PR #413, opened 2025-05-31
Add Async Event Primitive

toubatbrian:
Wrote `asyncio.Event` equivalent primitive

PR #508, opened 2025-07-02
Bugfix with subscription future

Shubhrakanti:


agents/src/voice/room_io/_output.ts
```diff
@@ -334,14 +334,12 @@ export class ParticipantAudioOutput extends AudioOutput {
...
   }
 
-  get queueSizeMs(): number {
-    return this.options.queueSizeMs ?? 100000;
+  get subscribed(): boolean {
+    return this.startedFuture.done;
   }
 
   async start(): Promise<void> {
-    this.startedFuture = new Future();
```
Shubhrakanti, line 342: We already initialize this in the constructor - so the first `captureFrame` was waiting for a Future that would never be resolved. 

I really we should never do `= new Future()` outside the constructor. We only do it in the `mainTask` of `AgentActivity`. Might make sense to make a class like asyncio's event.

toubatbrian: Time to merge https://github.com/livekit/agents-js/pull/413 :) ?

Shubhrakanti: Yeah I think we should be using that in the agent activity.

PR #509, opened 2025-07-02
Fix Tool Parsing to Allow Object Return Value

toubatbrian:


Shubhrakanti (approved):
make sure you test is manually. `0.x` is really flaky.

PR #511, opened 2025-07-03
brianyin/ajs-147-agent-session-interrupt

toubatbrian:


agents/src/voice/agent_activity.ts
```diff
@@ -393,6 +393,30 @@ export class AgentActivity implements RecognitionHooks {
...
+      future.resolve();
+    } else {
+      currentSpeech.then(() => {
+        if (future.done) return;
+        future.resolve();
+      });
+    }
+
+    return future;
+  }
```
toubatbrian, line 418: This PR is just forwarding the `interrupt` method from AgentActivity to `AgentSession` so user can use it

Shubhrakanti (approved):

PR #513, opened 2025-07-03
brianyin/ajs-140-generate-replysay-inside-tool-calling

toubatbrian:


agents/src/llm/llm.ts
```diff
@@ -42,12 +42,14 @@ export abstract class LLM extends (EventEmitter as new () => TypedEmitter<LLMCal
...
     toolCtx,
+    toolChoice,
     temperature,
     n,
     parallelToolCalls,
   }: {
     chatCtx: ChatContext;
     toolCtx?: ToolContext;
+    toolChoice?: ToolChoice;
     temperature?: number;
```
toubatbrian, line 53: We need to pass toolChoice all the way to OpenAI client call if the toolChoice is set to be not `undefined`

agents/src/voice/agent_activity.ts
```diff
@@ -61,8 +61,14 @@ export class AgentActivity implements RecognitionHooks {
...
-    // relies on JavaScript's built-in lexicographic array comparison behavior, which checks elements of the array one by one
-    this.speechQueue = new Heap<[number, number, SpeechHandle]>(Heap.maxComparator);
+    /**
+     * Custom comparator to prioritize speech handles with higher priority
+     * - Prefer higher priority
+     * - Prefer earlier timestamp (so calling a sequence of generateReply() will execute in FIFO order)
+     */
+    this.speechQueue = new Heap<[number, number, SpeechHandle]>(([p1, t1, _], [p2, t2, __]) => {
+      return p1 === p2 ? t1 - t2 : p2 - p1;
+    });
```
toubatbrian, line 71: A bug that had been bugging me for a while, finally tracked it down 😅

The issue stems from relying on JavaScript’s lexicographic ordering, which ends up comparing arrays in their stringified form rather than element by element.

What we actually want is to ensure that earlier scheduled speech handles are played first. This guarantees that any `generateReply` inside a tool call is triggered before the tool’s response speech. Otherwise, you end up hearing the tool response first, followed by the generateReply—in reverse order.

agents/src/voice/generation.ts
```diff
@@ -607,10 +607,12 @@ export function performToolExecutions({
...
-      const toolExecution = tool.execute(parsedArgs, {
-        ctx: new RunContext(session, speechHandle, toolCall),
-        toolCallId: toolCall.callId,
-        abortSignal: signal,
+      const toolExecution = asyncLocalStorage.run({ functionCall: toolCall }, async () => {
+        return await tool.execute(parsedArgs, {
+          ctx: new RunContext(session, speechHandle, toolCall),
+          toolCallId: toolCall.callId,
+          abortSignal: signal,
+        });
```
toubatbrian, line 619: Equivalent to python `ContextVar` to make sure async call stack get a global context variable

agents/src/voice/agent_activity.ts
Shubhrakanti: I see this makes sense. Good catch

PR #515, opened 2025-07-08
Shubhra/ajs 149 outline realtime model abstract class

Shubhrakanti:
Copies over https://github.com/livekit/agents/blob/fbcaca8498ce304b8ea7964c77e9c264c5946843/livekit-agents/livekit/agents/llm/realtime.py#L108

### DIFF
--- a/agents/src/llm/realtime.ts
+++ b/agents/src/llm/realtime.ts
@@ -13,7 +13,7 @@ export interface RealtimeCapabilities {
 }
 
 export abstract class RealtimeModel {
-  protected _capabilities: RealtimeCapabilities;
+  private _capabilities: RealtimeCapabilities;
 
   constructor(capabilities: RealtimeCapabilities) {
     this._capabilities = capabilities;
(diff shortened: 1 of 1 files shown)


agents/src/llm/realtime.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  abstract close(): Promise<void>;
+}
+
+export abstract class RealtimeSession {
+  protected _realtimeModel: RealtimeModel;
+
+  constructor(realtimeModel: RealtimeModel) {
+    this._realtimeModel = realtimeModel;
+  }
```
Toubat: any reason we use `protected`?

Toubat (approved):

agents/src/llm/realtime.ts
Shubhrakanti: Good catch. since we have `get realtimeModel` we don't need this to be protected anymore. Can just make it private.

Shubhrakanti: same with `capabilities`

PR #521, opened 2025-07-10
Shubhra/ajs 153 refactor open ai plugin to work with new interface

Shubhrakanti:
- I chose to use `AsyncIterableQueue` over streams because I want to keep things identical to python before we change any implementation details
- Public methods of `RealtimeSession` are not implemented yet to keep this PR small. https://github.com/livekit/agents-js/blob/f20dbed8bcada3cee0831226ddd4918d20ebec8b/agents/src/llm/realtime.ts#L57


agents/src/llm/realtime.ts
```diff
@@ -14,13 +15,13 @@ export interface InputSpeechStoppedEvent {
 
 export interface MessageGeneration {
   messageId: string;
-  textStream: ReadableStream<string>;
-  audioStream: ReadableStream<AudioFrame>;
+  textStream: AsyncIterableQueue<string>;
+  audioStream: AsyncIterableQueue<AudioFrame>;
```
Shubhrakanti, line 19: Keeping it consistent with python for now.

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -1,561 +1,438 @@
...
+  model: 'gpt-4o-realtime-preview',
+  voice: 'alloy',
+  temperature: DEFAULT_TEMPERATURE,
+  inputAudioTranscription: DEFAULT_INPUT_AUDIO_TRANSCRIPTION,
+  turnDetection: DEFAULT_TURN_DETECTION,
+  toolChoice: DEFAULT_TOOL_CHOICE,
+  maxResponseOutputTokens: DEFAULT_MAX_RESPONSE_OUTPUT_TOKENS,
+  maxSessionDuration: DEFAULT_MAX_SESSION_DURATION,
+};
+export class RealtimeModel extends llm.RealtimeModel {
```
Shubhrakanti, line 105: This class should be identical to the python oai plugin

turbo.json
```diff
@@ -20,6 +20,7 @@
     "LOG_LEVEL",
     "OCTOAI_TOKEN",
     "OPENAI_API_KEY",
+    "OPENAI_API_VERSION",
```
Shubhrakanti, line 23: did we even test this before?

PR #527, opened 2025-07-11
Shubhra/ajs 152 wire up public methods to realtime model

Shubhrakanti:
- `updateTools` will be done in a follow up PR


agents/src/audio.ts
```diff
@@ -51,13 +51,16 @@ export class AudioByteStream {
...
       new AudioFrame(
         new Int16Array(this.#buf.buffer),
         this.#sampleRate,
         this.#numChannels,
         this.#buf.length / 2,
       ),
     ];
+
+    this.#buf = new Int8Array(); // Clear buffer after flushing
+    return frames;
```
Shubhrakanti, line 64: this was a nasty bug. Would've caused a memory leak.

agents/src/llm/utils.ts
```diff
@@ -262,3 +268,85 @@ export function toError(error: unknown): Error {
...
+
+/**
+ * Standard dynamic-programming LCS to get the common subsequence
+ * of IDs (in order) that appear in both old_ids and new_ids.
+ *
+ * @param oldIds - The old list of IDs.
+ * @param newIds - The new list of IDs.
+ * @returns The longest common subsequence of the two lists of IDs.
+ */
+function computeLCS(oldIds: string[], newIds: string[]): string[] {
```
Shubhrakanti, line 280: copied over from `llm/utils.py` in python

agents/src/llm/utils.ts
```diff
@@ -262,3 +268,85 @@ export function toError(error: unknown): Error {
...
+}
+
+/**
+ * Compute the minimal list of create/remove operations to transform oldCtx into newCtx.
+ *
+ * @param oldCtx - The old chat context.
+ * @param newCtx - The new chat context.
+ * @returns The minimal list of create/remove operations to transform oldCtx into newCtx.
+ */
+export function computeChatCtxDiff(oldCtx: ChatContext, newCtx: ChatContext): DiffOps {
```
Shubhrakanti, line 329: copied from python

PR #532, opened 2025-07-14
brian/readonly-chat-ctx

toubatbrian:
- Add immutable array proxy to allows define readonly array.
- Add `ReadonlyChatContext` using the immutable array.

### DIFF
--- a/agents/src/llm/chat_context.ts
+++ b/agents/src/llm/chat_context.ts
@@ -264,7 +264,7 @@ export class ChatContext {
   protected _items: ChatItem[];
 
   constructor(items?: ChatItem[]) {
-    this._items = items !== undefined ? items : [];
+    this._items = items ? items : [];
   }
 
   static empty(): ChatContext {


agents/src/llm/chat_context.test.ts
```diff
@@ -279,3 +282,169 @@ describe('ChatContext.toJSON', () => {
...
+      }),
+    ];
+    const readonlyContext = new ReadonlyChatContext(items);
+
+    expect(readonlyContext.readonly).toBe(true);
+  });
+
+  it('should prevent setting items property', () => {
+    const items: ChatItem[] = [
+      new ChatMessage({
```
toubatbrian, line 304: all tests has been verified

agents/src/llm/chat_context.ts
```diff
@@ -261,10 +262,10 @@ export class FunctionCallOutput {
 export type ChatItem = ChatMessage | FunctionCall | FunctionCallOutput;
 
 export class ChatContext {
-  private _items: ChatItem[];
+  protected _items: ChatItem[];
 
   constructor(items?: ChatItem[]) {
-    this._items = items ? [...items] : [];
+    this._items = items !== undefined ? items : [];
```
toubatbrian: following python's implementation to not pass a copy when initializing / setting items

agents/src/llm/chat_context.ts
```diff
@@ -276,7 +277,7 @@ export class ChatContext {
   }
 
   set items(items: ChatItem[]) {
-    this._items = [...items];
+    this._items = items;
```
toubatbrian, line 279: ^

agents/src/llm/immutable_array.ts
```diff
@@ -0,0 +1,66 @@
...
+  'unshift',
+  'splice',
+  'sort',
+  'reverse',
+  'fill',
+  'copyWithin',
+] as const;
+
+export function createImmutableArray<T>(array: T[], additionalErrorMessage: string = ''): T[] {
+  return new Proxy(array, {
```
Shubhrakanti: Great comments. Maybe throw in a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy? above this line?

agents/src/llm/immutable_array.ts
```diff
@@ -0,0 +1,66 @@
...
+    setPrototypeOf() {
+      throw new TypeError(
+        `Cannot change prototype of a read-only array. ${additionalErrorMessage}`.trim(),
+      );
+    },
+  });
+}
+
+export function isImmutableArray(array: unknown): boolean {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Shubhrakanti: Makes sense to disable here. Nice

Shubhrakanti: Curious if there's any way around this?

agents/src/llm/immutable_array.ts
```diff
@@ -0,0 +1,66 @@
...
+      throw new TypeError(
+        `Cannot define property "${String(prop)}" on a read-only array. ${additionalErrorMessage}`.trim(),
+      );
+    },
+    // block changing the prototype
+    setPrototypeOf() {
+      throw new TypeError(
+        `Cannot change prototype of a read-only array. ${additionalErrorMessage}`.trim(),
+      );
+    },
```
Shubhrakanti: what does changing the prototype do? Just curious.

agents/src/llm/immutable_array.ts
```diff
@@ -0,0 +1,66 @@
...
+  'fill',
+  'copyWithin',
+] as const;
+
+export function createImmutableArray<T>(array: T[], additionalErrorMessage: string = ''): T[] {
+  return new Proxy(array, {
+    get(target, key) {
+      if (key === READONLY_SYMBOL) {
+        return true;
+      }
```
Shubhrakanti: Very neat

agents/src/llm/immutable_array.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      // Intercept mutation methods
+      if (typeof key === 'string' && MUTATION_METHODS.includes(key as any)) {
```
Shubhrakanti: why do we need as `key as any`? we already checked it's a string?

agents/src/llm/immutable_array.ts
```diff
@@ -0,0 +1,66 @@
+// SPDX-FileCopyrightText: 2025 LiveKit, Inc.
```
Shubhrakanti: Should this file be in like utils folder? why under llm?

Shubhrakanti: I guess because we only use it in one place for now.

agents/src/llm/chat_context.ts
Shubhrakanti: can't we just do `this._items = items ? items : []`

agents/src/llm/immutable_array.ts
toubatbrian: `SYMBOL` probably the easiest way to check immutability, as using proxy changes the underling object behavior, not really making it a different type.

agents/src/llm/immutable_array.ts
```diff
@@ -0,0 +1,66 @@
...
+        `Cannot delete read-only array index "${String(prop)}". ${additionalErrorMessage}`.trim(),
+      );
+    },
+    // block adding or modifying property descriptors
+    defineProperty(_, prop) {
+      throw new TypeError(
+        `Cannot define property "${String(prop)}" on a read-only array. ${additionalErrorMessage}`.trim(),
+      );
+    },
+    // block changing the prototype
```
Shubhrakanti: these comments are kind of unnecessary? It literally says what the code does

agents/src/llm/immutable_array.ts
toubatbrian: It's a pretty low level operation that change the set of methods on a object like:

```
const newProto = {
  greet() { return "hi"; }
};

const arr = [1, 2, 3];
const success = Object.setPrototypeOf(arr newProto);
arr.greet();
```

Pretty much the earliest way to create a object with a "class", but rarely do people use this type of syntax nowadays.

Shubhrakanti: thanks!

agents/src/llm/immutable_array.ts
toubatbrian: Getting this:
```
Argument of type 'string' is not assignable to parameter of type '"push" | "pop" | "shift" | "unshift" | "splice" | "sort" | "reverse" | "fill" | "copyWithin"'
```

Since `MUTATION_METHODS` only accept type `"push" | "pop" | "shift" | "unshift" | "splice" | "sort" | "reverse" | "fill" | "copyWithin"'`; however string is a more broader type which the method won't accept

agents/src/llm/immutable_array.ts
```diff
@@ -0,0 +1,66 @@
...
+  'push',
+  'pop',
+  'shift',
+  'unshift',
+  'splice',
+  'sort',
+  'reverse',
+  'fill',
+  'copyWithin',
+] as const;
```
Shubhrakanti: why `as const`?

agents/src/llm/immutable_array.ts
toubatbrian: I moved to utils.py

agents/src/llm/immutable_array.ts
toubatbrian: Just a convention if we are not really making changes to the array object

PR #533, opened 2025-07-15
brianyin/ajs-122-use-zod-to-json-schema-for-better-unified-tool-arg-parsing

toubatbrian:
Update `llm.oaiParams` to use library `zod-to-json-schema`

### DIFF
--- a/agents/src/llm/utils.ts
+++ b/agents/src/llm/utils.ts
@@ -122,20 +122,6 @@ export const createToolOptions = <UserData extends UnknownUserData>(
   return { ctx: { userData }, toolCallId } as unknown as ToolOptions<UserData>;
 };
 
-// XXX: Zod is victim to the dual-package hazard. this is a hacky sorta-fix
-// until Zod v4.0.0 is released.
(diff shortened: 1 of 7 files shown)


agents/src/llm/utils.ts
```diff
@@ -122,78 +122,19 @@ export const createToolOptions = <UserData extends UnknownUserData>(
...
-    properties[key] = processZodType(field);
-
-    if (!(field instanceof z.ZodOptional)) {
-      requiredProperties.push(key);
-    }
-  }
+  // TODO(AJS-162): make zod to JSON parsing able to handle z.optional(v.field())
+  const { properties, required, additionalProperties } = zodToJsonSchema(p, {
+    target: 'openAi',
+  }) as OpenAIFunctionParameters;
```
toubatbrian, line 131: Currently tool have poor support for optional field (we can use z.nullable() for a workaround).

toubatbrian: Added a ticket for this, should be a low priority feature to handle

PR #535, opened 2025-07-15
Shubhra/add cset brain

Shubhrakanti:
revert #534

### DIFF
--- a/plugins/openai/src/llm.ts
+++ b/plugins/openai/src/llm.ts
@@ -636,27 +636,36 @@ const buildMessage = async (msg: llm.ChatMessage, cacheKey: any) => {
       break;
   }
 
-  if (typeof msg.content === 'string') {
-    oaiMsg.content = msg.content;
-  } else if (Array.isArray(msg.content)) {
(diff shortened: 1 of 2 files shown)


toubatbrian (approved):

plugins/openai/src/llm.ts
```diff
@@ -636,27 +636,36 @@ const buildMessage = async (msg: llm.ChatMessage, cacheKey: any) => {
...
-        } else {
-          throw new Error('ChatAudio is not supported');
-        }
-      }),
-    )) as OpenAI.ChatCompletionContentPart[];
-  } else if (msg.content === undefined) {
-    oaiMsg.content = '';
+  if (msg.role === llm.ChatRole.TOOL) {
+    try {
+      const toolCallOutput = JSON.stringify(msg.content);
```
davidzhao, line 641: this would encode strings as `"string body"` right? we should probably not serialize strings again

Shubhrakanti: @toubatbrian

PR #538, opened 2025-07-15
brianyin/ajs-111-support-raw-json-schema-for-function-calling

toubatbrian:
Support passing raw JSON schema to define tool


Shubhrakanti (approved):
don't have time to dig into the weeds. 

I would run a lot of this typing logic in the function definitions by @lukasIO.

PR #541, opened 2025-07-15
Shubhra/ajs 137 add resampler to input

Shubhrakanti:
Res-sample the input based on user config


agents/src/utils.ts
```diff
@@ -587,3 +590,27 @@ export function isImmutableArray(array: unknown): boolean {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   return typeof array === 'object' && !!(array as any)[READONLY_SYMBOL];
 }
+
+export function resampleStream({
```
Shubhrakanti, line 603: We will probably need this in the plugins. That's why I kept it a separate utility.

toubatbrian: can you add a brief function spec to this function?

Shubhrakanti: Like a JS doc you mean?

agents/src/utils.ts
```diff
@@ -587,3 +590,27 @@ export function isImmutableArray(array: unknown): boolean {
...
+      }
+      for (const frame of resampler.push(chunk)) {
+        controller.enqueue(frame);
+      }
+      for (const frame of resampler.flush()) {
+        controller.enqueue(frame);
+      }
+    },
+  });
+  return stream.pipeThrough(transformStream);
```
toubatbrian, line 624: note that this will lock the input stream argument. When doing resource cleanup, we need to keep this in mind to avoid potential locking issue.

Shubhrakanti: this a good callout. I'll add a comment for the function.

toubatbrian: I wonder if we make it a class like "ResampledStream" and we expose method for unlock, cancel, just like deferred readable stream.

I think instead of using `pipeThrough` we should probably use manual `getReader()` to pipe the input stream to the transformStream just to make things consistent, because we might need to deal with bugs if using `pipeThrough` as we lose control of input stream's lock. What do you think?

agents/src/voice/room_io/_input.ts
```diff
@@ -83,7 +84,12 @@ export class ParticipantAudioInputStream {
     }
     this.publication = publication;
     this.logger.debug({ track, publication, participant }, 'track subscribed');
-    this.deferredStream.setSource(this.createStream(track));
+    this.deferredStream.setSource(
+      resampleStream({
```
toubatbrian, line 88: ^ same, we need to take care of un-`pipeThrough` whenever we are going to use it.

Shubhrakanti: What do you mean "un-pipeThrough"? If you call `cancel` on the output stream then it will signal it back down the chain?

agents/src/utils.ts
toubatbrian: yeah, just a one or two line brief description would be fine

agents/src/voice/room_io/_input.ts
toubatbrian: If we cancel the `pipeThrough` the source and destination stream will be cancelled as well and stop stream immediately, but in some situation, like deferred readable stream, when we"detach" source stream we still want to keep source stream "alive" and pipe to other stream later. Not sure if we'd ever want to "detach" source stream here tho

Shubhrakanti: This is a great point. Let me look through the code base and see what makes sense.

Shubhrakanti: > when we"detach" source stream we still want to keep source stream "alive" and pipe to other stream later

you can do this via `preventCancel/Abort?` [[source](https://streams.spec.whatwg.org/#rs-prototype)]

Shubhrakanti: readable = stream.[pipeThrough](https://streams.spec.whatwg.org/#rs-pipe-through)({ [writable](https://streams.spec.whatwg.org/#dom-readablewritablepair-writable), [readable](https://streams.spec.whatwg.org/#dom-readablewritablepair-readable) }[, { [preventClose](https://streams.spec.whatwg.org/#dom-streampipeoptions-preventclose), [preventAbort](https://streams.spec.whatwg.org/#dom-streampipeoptions-preventabort), [preventCancel](https://streams.spec.whatwg.org/#dom-streampipeoptions-preventcancel), [signal](https://streams.spec.whatwg.org/#dom-streampipeoptions-signal) }])

agents/src/utils.ts
Shubhrakanti: We can just use the [streampipe options](https://streams.spec.whatwg.org/#dom-streampipeoptions-preventclose)?

toubatbrian: I haven't tried it yet, but feel free to try it.

agents/src/voice/room_io/_input.ts
toubatbrian: I think we would set `preventCancel` to true in this case? Screenshot 2025-07-15 at 4 19 30 PM

Shubhrakanti: Yeah I think you're right. Going to test it with restaurant agent and handoff to make sure it works okay. Then I'll let you know.

Shubhrakanti: hand off worked correctly so I think we don't need it here.

PR #542, opened 2025-07-16
Shubhra/ajs 60 stanrdize id generator with python codebase

Shubhrakanti:
Make sure we are using the same standards as python

### DIFF
--- a/agents/src/utils.ts
+++ b/agents/src/utils.ts
@@ -524,7 +524,7 @@ export function withResolvers<T = unknown>() {
  * @param prefix - The prefix to add to the UUID.
  * @returns A short UUID with the prefix.
  */
-export function shortuuid(prefix: string): string {
+export function shortuuid(prefix: string = ''): string {
   return `${prefix}${uuidv4().slice(0, 12)}`;
 }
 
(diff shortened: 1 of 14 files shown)


agents/src/utils.ts
```diff
@@ -518,9 +518,14 @@ export function withResolvers<T = unknown>() {
...
-export function shortuuid(prefix: string): string {
-  return `${prefix}_${uuidv4().slice(0, 12)}`;
+/**
+ * Generates a short UUID with a prefix. Mirrors the python agents implementation.
+ *
+ * @param prefix - The prefix to add to the UUID.
+ * @returns A short UUID with the prefix.
+ */
+export function shortuuid(prefix: string = ''): string {
+  return `${prefix}${uuidv4().slice(0, 12)}`;
```
Shubhrakanti, line 528: I removed the `_` to keep it the exact same as the python implementation. @toubatbrian if we have the same function copied over from python with subtle differences we should either: 

1. Document it 
2. Bring it up with Theo and Long and change the behavior in python as well.

This will stop us from having a bunch of subtle bugs all over the place.

PR #544, opened 2025-07-16
brianyin/ajs-151-add-function-calling-to-realtime-model

toubatbrian:
Implement realtime model tool calling flow. TODO: realtime reply task, ticket already added.


Shubhrakanti (approved):
Didn't go super in-depth. But over all looks good to me.

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -617,7 +670,20 @@ export class RealtimeSession extends llm.RealtimeSession {
...
+              this.#logger.debug(
+                `(client) -> ${JSON.stringify(
+                  {
+                    ...event,
+                    audio: event.audio.slice(0, 10) + '...',
+                  },
+                  null,
+                  2,
+                )}`,
+              );
```
Shubhrakanti: This is unrelated to function calling. Would prefer a separate PR next time.

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -601,6 +645,15 @@ export class RealtimeSession extends llm.RealtimeSession {
     return ws;
   }
 
+  #isSilentAudio(audio: string): boolean {
```
Shubhrakanti: Don't use `#` for private. Just use `private`. This is there from the legacy code and we need to get rid of it.

toubatbrian: This function is for debug purpose, I forgot to remove it :)

plugins/openai/src/realtime/realtime_model.ts
toubatbrian: yep, forgot to cleanup this debug log, removed now

PR #545, opened 2025-07-17
[Do Not Merge] Stream Channel

toubatbrian:
Create a `StreamChannel` object mimicking python `Chan` but also preserves JS feature by exposing `ReadbleStream` object. This help repace existing AsyncIterableQueue, which is needed to implement tool calling as current `performToolExecution` accepts a ReadbleStream.


agents/src/stream/stream_channel.ts
```diff
@@ -0,0 +1,32 @@
...
+    stream: () => transform.readable,
+    close: async () => {
+      try {
+        return await writer.close();
+      } catch (e) {
+        if (e instanceof Error && e.name === 'TypeError') {
+          // Ignore error if the stream is already closed
+          return;
+        }
+        throw e;
```
toubatbrian, line 28: A utility to use instead of AsyncIterableQueue. It expose a readable stream object that we can pass around without the need to indirectly pipeling from queue to stream

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -959,13 +931,19 @@ export class RealtimeSession extends llm.RealtimeSession {
...
+
+    itemGeneration.audioChannel.write(
       new AudioFrame(
-        new Int16Array(data.buffer),
+        new Int16Array(bytes.buffer),
         api_proto.SAMPLE_RATE,
         api_proto.NUM_CHANNELS,
-        data.length / 2,
+        bytes.length / 2,
       ),
```
toubatbrian, line 947: Based on openai-agent-js: https://github.com/openai/openai-agents-js/blob/af73bfb5178a732d54934bbf1a2fdc37da5f3839/packages/agents-realtime/src/openaiRealtimeWebsocket.ts#L216

agents/src/index.ts
```diff
@@ -14,21 +14,22 @@ import * as ipc from './ipc/index.js';
...
+export * from './plugin.js';
 export * from './transcription.js';
-export * from './inference_runner.js';
+export * from './utils.js';
+export * from './vad.js';
+export * from './version.js';
+export * from './worker.js';
 
-export { cli, stt, tts, llm, multimodal, tokenize, metrics, ipc, voice };
+export { cli, ipc, llm, metrics, multimodal, stream, stt, tokenize, tts, voice };
```
Shubhrakanti, line 35: what are all these changes?

agents/src/stream/stream_channel.ts
Shubhrakanti: Can you map all the `AsyncIterableQueue` public methods to this interface? 

what is the equivalent of `close` and `closed()`?

agents/src/index.ts
toubatbrian: from linting

Shubhrakanti: Next time can you flag that as a comment? or do it in a separate PR?

agents/src/stream/stream_channel.ts
Shubhrakanti: Mainly how do we know when the stream is closed? because if we can check that we can get rid of `AsyncIterableQueue` completely and refactor everything to use this.

toubatbrian: > Can you map all the AsyncIterableQueue public methods to this interface?

Any reason we'd like to keep the same public interface? (for checking is `closed` I will definitely add it later) For `[Symbol.asyncIterator]()` and `next` calling `stream()` should already have those since stream itself is a async iterator.

Shubhrakanti: > Any reason we'd like to keep the same public interface?

Nothing wrong with this. I just want to remove `AsyncIterableQueue`

agents/src/index.ts
toubatbrian: yep for sure, sometimes I might forgot those minor details, I'll make a github PR template right now to list those requirements.

agents/src/stream/stream_channel.ts
toubatbrian: SG, will do in a follow up PR.

Shubhrakanti: this is not urgent. We should do this when have a broader discussion around plugins next week.

PR #550, opened 2025-07-19
brianyin/ajs-171-realtime-reply-task

toubatbrian:


plugins/openai/src/realtime/api_proto.ts
```diff
@@ -344,6 +344,7 @@ export interface ResponseCreateEvent extends BaseClientEvent {
     tool_choice: ToolChoice;
     temperature: number;
     max_output_tokens: number | 'inf';
+    metadata?: Record<string, string>;
```
toubatbrian, line 347: Need this in order to identify wether response is user initiated

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -1102,10 +1117,15 @@ export class RealtimeSession extends llm.RealtimeSession {
...
 
+    const response: api_proto.ResponseCreateEvent['response'] = {};
+    if (instructions) response.instructions = instructions;
+    if (userInitiated) response.metadata = { client_event_id: eventId };
+
     this.sendEvent({
       type: 'response.create',
       event_id: eventId,
-    } as api_proto.ResponseCreateEvent);
+      response: Object.keys(response).length > 0 ? response : undefined,
```
toubatbrian, line 1127: only pass defined key-value to response

plugins/openai/src/realtime/api_proto.ts
Shubhrakanti: should it be `?` or just there all the time?

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -19,7 +19,7 @@ interface RealtimeOptions {
   model: api_proto.Model;
   voice: api_proto.Voice;
   temperature: number;
-  toolChoice: llm.ToolChoice;
+  toolChoice?: llm.ToolChoice;
```
Shubhrakanti, line 22: why is this now optional?

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -1243,7 +1263,7 @@ function createMockAudioItem(durationSeconds: number = 2): llm.ChatMessage {
   });
 }
 
-function toOaiToolChoice(toolChoice: llm.ToolChoice | null): api_proto.ToolChoice {
+function toOaiToolChoice(toolChoice: llm.ToolChoice | undefined): api_proto.ToolChoice {
```
Shubhrakanti: `toolChoice?: llm.ToolChoice`?

Shubhrakanti (approved):

plugins/openai/src/realtime/api_proto.ts
toubatbrian: can be undefined I think

plugins/openai/src/realtime/realtime_model.ts
toubatbrian: python has this as well, doesn't need to be defined since undefined will use the 'auto' mode by default

PR #552, opened 2025-07-19
Shubhra/ajs 159 add agent session events

Shubhrakanti:
The python events are serialize by default because they are all pydantic models. We will need to think about do to make the js side serializable in the future.


toubatbrian:
for serialization, maybe we can just do a `JSON.stringify`?

PR #556, opened 2025-07-20
Shubhra/allow users to set participants

Shubhrakanti:
Since users can now set participant to be `null` explicitly I changed the type everywhere to be `participant | null`


Shubhrakanti:
Also handles https://linear.app/livekit/issue/AJS-106/add-participant-disconnect-and-re-connect-support

Shubhrakanti:
Also does AJS-106

Shubhrakanti:
I tested manually and it works fine.

PR #559, opened 2025-07-21
brianyin/ajs-170-wait-for-message-item-create-future-in-realtime-model

toubatbrian:
This reverts commit ba30c3268bf81d9be7f208900ac4bdaee09676ee.

### DIFF
--- a/agents/src/voice/generation.ts
+++ b/agents/src/voice/generation.ts
@@ -86,9 +86,13 @@ export class _SanitizedOutput {
 }
 
 function isValidToolOutput(toolOutput: unknown): boolean {
-  const validTypes = ['string', 'number', 'boolean', undefined, null];
+  const validTypes = ['string', 'number', 'boolean'];
 
   if (validTypes.includes(typeof toolOutput)) {
+    return true;
+  }
+
+  if (toolOutput === undefined || toolOutput === null) {
     return true;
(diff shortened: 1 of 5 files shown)


agents/src/voice/generation.ts
```diff
@@ -86,12 +86,16 @@ export class _SanitizedOutput {
...
+  const validTypes = ['string', 'number', 'boolean'];
 
   if (validTypes.includes(typeof toolOutput)) {
     return true;
   }
 
+  if (toolOutput === undefined || toolOutput === null) {
+    return true;
+  }
+
```
toubatbrian, line 98: `undefined !== typeof undefined` in typescript :((, so we handle undefined and null separately

plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -406,14 +407,40 @@ export class RealtimeSession extends llm.RealtimeSession {
...
+      return;
+    }
+
+    try {
+      // wait for futures to resolve or timeout
+      await Promise.race([
+        Promise.all(futures),
+        delay(5000).then(() => {
+          throw new Error('Chat ctx update events timed out');
+        }),
```
toubatbrian, line 436: wait or timeout

PR #566, opened 2025-07-22
Shubhra/ajs 165 bug with realtime model interrupts

Shubhrakanti:
We weren't actually waiting for all the tasks to complete.


agents/src/voice/agent_activity.ts
```diff
@@ -1282,7 +1282,7 @@ export class AgentActivity implements RecognitionHooks {
           }
           outputs.push([msg.messageId, textOut, audioOut]);
         }
-        await Promise.allSettled(forwardTasks);
+        await Promise.allSettled(forwardTasks.map((task) => task.result));
```
toubatbrian, line 1285: nice catch!

PR #569, opened 2025-07-22
Shubhra/ajs 173 fix stt resampling issue

Shubhrakanti:
We should be resampling the audio before we write send it to the STT - not hard coding the sample rate of the room. Tested all the pipeline agents manually.


examples/src/restaurant_agent.ts
```diff
@@ -384,9 +384,7 @@ export default defineAgent({
     const vad = ctx.proc.userData.vad! as silero.VAD;
     const session = new voice.AgentSession({
       vad,
-      stt: new deepgram.STT({
-        sampleRate: 24000,
-      }),
```
Shubhrakanti, line 389: We should never expect the user to know this.

PR #570, opened 2025-07-22
brianyin/ajs-164-make-sure-audiorecognition-works-with-realtime-model

toubatbrian:
Allow audioRecognition VAD/EOU co-work with realtime model

### DIFF
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -29,7 +29,7 @@ import type {
   TTSMetrics,
   VADMetrics,
 } from '../metrics/base.js';
-import type { STT, SpeechEvent } from '../stt/stt.js';
+import { STT, type SpeechEvent } from '../stt/stt.js';
 import { splitWords } from '../tokenize/basic/word.js';
 import { TTS } from '../tts/tts.js';
 import { Future, Task, cancelAndWait, waitFor } from '../utils.js';
@@ -891,6 +891,7 @@ export class AgentActivity implements RecognitionHooks {
     const speechHandle = this.generateReply({ userMessage, chatCtx });
 
     const eouMetrics: EOUMetrics = {
+      type: 'eou_metrics',
       timestamp: Date.now(),
       endOfUtteranceDelay: info.endOfUtteranceDelay,
       transcriptionDelay: info.transcriptionDelay,
@@ -900,7 +901,7 @@ export class AgentActivity implements RecognitionHooks {
 
     this.agentSession.emit(
       AgentSessionEventTypes.MetricsCollected,
-      createMetricsCollectedEvent(eouMetrics),
+      createMetricsCollectedEvent({ metrics: eouMetrics }),
     );
   }
 
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -334,8 +334,13 @@ export class AgentActivity implements RecognitionHooks {
   }
 
   attachAudioInput(audioStream: ReadableStream<AudioFrame>): void {
+    /**
+     * We need to add a deferred ReadableStream layer on top of the audioStream from the agent session.
+     * The tee() operation should be applied to the deferred stream, not the original audioStream.
+     * This is important because teeing the original stream directly makes it very difficult—if not
+     * impossible—to implement stream unlock logic cleanly.
+     */
     this.audioStream.setSource(audioStream);
-
     const [realtimeAudioStream, recognitionAudioStream] = this.audioStream.stream.tee();
 
     if (this.realtimeSession) {
@@ -1819,16 +1824,6 @@ export class AgentActivity implements RecognitionHooks {
       await this.realtimeSession?.close();
       await this.audioRecognition?.close();
       await this._mainTask?.cancelAndWait();
-
-      if (this.llm instanceof RealtimeModel) {
-        this.realtimeSession?.off('generation_created', this.onGenerationCreated);
-        this.realtimeSession?.off('input_speech_started', this.onInputSpeechStarted);
-        this.realtimeSession?.off('input_speech_stopped', this.onInputSpeechStopped);
-        this.realtimeSession?.off(
-          'input_audio_transcription_completed',
-          this.onInputAudioTranscriptionCompleted,
-        );
-      }
     } finally {
       unlock();
     }
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -93,6 +93,81 @@ export class AgentActivity implements RecognitionHooks {
 
     this.turnDetectionMode =
       typeof this.turnDetection === 'string' ? this.turnDetection : undefined;
+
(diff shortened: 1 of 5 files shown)


agents/src/voice/agent_activity.ts
```diff
@@ -103,6 +106,81 @@ export class AgentActivity implements RecognitionHooks {
...
+      if (this.turnDetectionMode === 'realtime_llm' && !this.llm.capabilities.turnDetection) {
+        this.logger.warn(
+          'turnDetection is set to "realtime_llm", but the LLM is not a RealtimeModel or the server-side turn detection is not supported/enabled, ignoring the turnDetection setting',
+        );
+        this.turnDetectionMode = undefined;
+      }
+
+      if (this.turnDetectionMode === 'stt') {
+        this.logger.warn(
+          'turnDetection is set to "stt", but the LLM is a RealtimeModel, ignoring the turnDetection setting',
```
toubatbrian, line 141: copied from python

agents/src/voice/agent_activity.ts
```diff
@@ -257,13 +333,22 @@ export class AgentActivity implements RecognitionHooks {
     }
   }
 
-  updateAudioInput(audioStream: ReadableStream<AudioFrame>): void {
-    // TODO(AJS-164): might need to tee the streams here.
+  attachAudioInput(audioStream: ReadableStream<AudioFrame>): void {
+    this.audioStream.setSource(audioStream);
+
+    const [realtimeAudioStream, recognitionAudioStream] = this.audioStream.stream.tee();
+
```
toubatbrian: We need to add a deferred ReadableStream layer on top of the audioStream from the agent session. The tee() operation should be applied to the deferred stream, not the original audioStream. This is important because teeing the original stream directly makes it very difficult—if not impossible—to implement stream unlock logic cleanly.

The deferred stream acts as a protective layer, shielding the source stream from being locked. Once all operations on the deferred stream (or its derived streams) are complete, we can detach the underlying stream. This automatically pauses the lower-level streams while keeping the upper-level stream in an unlocked state.

agents/src/voice/audio_recognition.ts
```diff
@@ -184,6 +184,7 @@ export class AudioRecognition {
         }
         break;
       case SpeechEventType.INTERIM_TRANSCRIPT:
+        this.logger.debug({ transcript: ev.alternatives?.[0]?.text }, 'interim transcript');
```
toubatbrian, line 187: Let's keep some important debug logs like interm transcripts, VAD start/end.

agents/src/voice/agent_activity.ts
```diff
@@ -103,6 +106,81 @@ export class AgentActivity implements RecognitionHooks {
 
     this.turnDetectionMode =
       typeof this.turnDetection === 'string' ? this.turnDetection : undefined;
+
+    if (this.turnDetectionMode === 'vad' && this.vad === undefined) {
+      this.logger.warn(
+        'turnDetection is set to "vad", but no VAD model is provided, ignoring the turnDdetection setting',
+      );
+      this.turnDetectionMode = undefined;
```
Shubhrakanti, line 114: I don't love assigning values to undefined. `undefined` should represent values that are not yet assigned. If we are setting a value explicitly we should use `null`. 

I think these should be typed as `private variable: TYPE | null = null`. And set by default to null.

Shubhrakanti: https://gibdon.com/2006/05/javascript-difference-between-null-and.html

agents/src/voice/agent_activity.ts
Shubhrakanti: > The tee() operation should be applied to the deferred stream, not the original audioStream

Nice! how did you figure this out? Also I'd add a short comment in the code about this.

agents/src/voice/agent_activity.ts
```diff
@@ -778,7 +868,7 @@ export class AgentActivity implements RecognitionHooks {
       this.realtimeSession?.interrupt();
     }
 
-    const userMessage = ChatMessage.create({
+    let userMessage: ChatMessage | undefined = ChatMessage.create({
```
Shubhrakanti, line 876: `let userMessage?`

agents/src/voice/agent_activity.ts
```diff
@@ -1699,11 +1815,20 @@ export class AgentActivity implements RecognitionHooks {
...
-      this.agent._agentActivity = undefined;
+      if (this.llm instanceof RealtimeModel) {
+        this.realtimeSession?.off('generation_created', this.onGenerationCreated);
+        this.realtimeSession?.off('input_speech_started', this.onInputSpeechStarted);
+        this.realtimeSession?.off('input_speech_stopped', this.onInputSpeechStopped);
+        this.realtimeSession?.off(
+          'input_audio_transcription_completed',
+          this.onInputAudioTranscriptionCompleted,
+        );
+      }
```
Shubhrakanti: This is already done. Just look up a few lines. 

```typescript
      if (this.realtimeSession) {
        this.realtimeSession.off('generation_created', this.onGenerationCreated);
        this.realtimeSession.off('input_speech_started', this.onInputSpeechStarted);
        this.realtimeSession.off('input_speech_stopped', this.onInputSpeechStopped);
        this.realtimeSession.off(
          'input_audio_transcription_completed',
          this.onInputAudioTranscriptionCompleted,
        );
        this.realtimeSession.off('metrics_collected', this.onMetricsCollected);
      }
```

agents/src/voice/audio_recognition.ts
Shubhrakanti: Isn't this super noisy?

Shubhrakanti (approved):
Great job! Feel free to merge after you go through the comments.

agents/src/voice/agent_activity.ts
toubatbrian: This one is quite tricky, took me hours to figure out lol. By introducing a deferred stream turned out to be super helpful! It allows you to create multiple derived stream objects from an incoming source while maintaining a clean and consistent way to cancel them all.

agents/src/voice/audio_recognition.ts
toubatbrian: This is just intermediate transcripts, it won't fire very frequently

agents/src/voice/agent_activity.ts
toubatbrian: ok SG. Changing them all to using `null` would requires change the interfaces on multiple places, I'll add a ticket for this.

agents/src/voice/agent_activity.ts
toubatbrian: I don't think `var?` is a valid syntax for declaring a optional var in ts?

Shubhrakanti: no for the typing we can just do `?` instead of `| undefined`

agents/src/voice/agent_activity.ts
Shubhrakanti: nice yeah this isn't super urgent which we can do later.

PR #573, opened 2025-07-23
shubhra/ajs 176 create agent input and output interfaces

Shubhrakanti:
Add `AgentInput` and `AgentOutput` interfaces.

### DIFF
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -334,6 +334,11 @@ export class AgentActivity implements RecognitionHooks {
   }
 
   attachAudioInput(audioStream: ReadableStream<AudioFrame>): void {
+    if (this.audioStream.isSourceSet) {
+      this.logger.debug('detaching existing audio input in agent activity');
+      this.audioStream.detachSource();
+    }
+
     /**
      * We need to add a deferred ReadableStream layer on top of the audioStream from the agent session.
      * The tee() operation should be applied to the deferred stream, not the original audioStream.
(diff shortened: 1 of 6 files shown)


Shubhrakanti:
AJS-176 ignore the branch name it's wrong

agents/src/voice/agent_activity.ts
```diff
@@ -334,6 +334,11 @@ export class AgentActivity implements RecognitionHooks {
   }
 
   attachAudioInput(audioStream: ReadableStream<AudioFrame>): void {
+    if (this.audioStream.isSourceSet) {
+      this.logger.debug('detaching existing audio input in agent activity');
+      this.audioStream.detachSource();
+    }
+
```
Shubhrakanti, line 341: this is slightly different from python where we handle this in the `AgentSession.onAudioInputChanged`

Shubhrakanti: I think it's okay but let me know your thoughts @toubatbrian

toubatbrian: Have you tried if this work? Right now I don't think deferred stream would handle a re-attach after detaching the source.

agents/src/voice/agent_session.ts
```diff
@@ -127,11 +123,23 @@ export class AgentSession<
     this.turnDetection = turnDetection;
     this._userData = userData;
 
+    // configurable IO
+    this._input = new AgentInput(this.onAudioInputChanged);
+    this._output = new AgentOutput(this.onAudioOutputChanged, this.onTextOutputChanged);
```
toubatbrian, line 128:
```
() => this.onAudioInputChanged()
```
?

agents/src/voice/agent_session.ts
```diff
@@ -317,4 +325,19 @@ export class AgentSession<
...
   }
+
+  // -- User changed input/output streams/sinks --
+  private onAudioInputChanged(): void {
+    if (!this.started) {
+      return;
+    }
+
+    if (this.activity && this._input.audio) {
+      this.activity.attachAudioInput(this._input.audio.stream);
```
toubatbrian, line 336: I exposed a `detachAudioInput` method in `AgentActivity`, which is intended to `detach` and then `attach` again when audio/video changes

agents/src/voice/agent_session.ts
Shubhrakanti: Oh man good catch.

PR #576, opened 2025-07-24
Shubhra/ajs 177 add close methods to all inputs and outputs 

Shubhrakanti:
Add everything needed to close the `AgentSession`


agents/src/voice/room_io/_input.ts
```diff
@@ -135,4 +135,11 @@ export class ParticipantAudioInputStream extends AudioInput {
       noiseCancellation: this.noiseCancellation,
     });
   }
+
+  async close() {
+    this.room.off(RoomEvent.TrackSubscribed, this.onTrackSubscribed);
+    this.room.off(RoomEvent.TrackUnpublished, this.onTrackUnpublished);
+    this.closeStream();
+    this.deferredStream.stream.cancel();
```
toubatbrian, line 142: would be safer to call `this.deferredStream.detachSource` before calling `cancel`

agents/src/voice/room_io/room_io.ts
```diff
@@ -184,10 +188,10 @@ export class RoomIO {
     });
   };
 
-  private async forwardUserTranscript(): Promise<void> {
+  private async forwardUserTranscript(signal: AbortSignal): Promise<void> {
     const reader = this.userTranscriptStream.readable.getReader();
     try {
-      while (true) {
+      while (!signal.aborted) {
         const { done, value } = await reader.read();
```
toubatbrian, line 263: can we make sure this `reader.read()` won't block? i.e. we may also need to close reader as well

toubatbrian: Or we do a Promise.race on a abort future with reader.read()

agents/src/voice/room_io/room_io.ts
```diff
@@ -361,4 +365,23 @@ export class RoomIO {
...
+  async close() {
+    this.room.off(RoomEvent.ParticipantConnected, this.onParticipantConnected);
+    this.room.off(RoomEvent.ConnectionStateChanged, this.onConnectionStateChanged);
+    this.room.off(RoomEvent.ParticipantDisconnected, this.onParticipantDisconnected);
+
+    await this.initTask?.cancelAndWait();
+    await this.forwardUserTranscriptTask?.cancelAndWait();
+
+    this.userTranscriptStream.writable.close();
+    this.userTranscriptStream.readable.cancel();
```
toubatbrian: I recalled that cancelling one side of them (writable / readable) should be sufficient; cancelling both could raise an error.

agents/src/voice/agent_session.ts
```diff
@@ -340,4 +348,40 @@ export class AgentSession<
...
+      await this.activity?.interrupt().await;
+    } catch (error) {
+      // uninterruptible speech
+      // TODO(shubhra): force interrupt or wait for it to finish?
+      // it might be an audio played from the error callback
+      this.logger.error(error, 'Failed to interrupt activity');
+    }
+
+    await this.activity?.drain();
+    await this.activity?.currentSpeech?.waitForPlayout();
```
toubatbrian, line 367: we also need to call `await this.activity?.close()` here as well?

agents/src/voice/room_io/_input.ts
Shubhrakanti: Good catch I detach the source of the audio input in the `AgentSession`

agents/src/voice/room_io/room_io.ts
Shubhrakanti: Great catch. I close just the writable side before aborting this signal. I added a comment for clarity.

agents/src/voice/room_io/room_io.ts
```diff
@@ -440,4 +444,23 @@ export class RoomIO {
...
+    this.agentSession.off(AgentSessionEventTypes.UserInputTranscribed, this.onUserInputTranscribed);
+    this.agentSession.off(AgentSessionEventTypes.AgentStateChanged, this.onAgentStateChanged);
+
+    await this.initTask?.cancelAndWait();
+
+    // Close stream FIRST so reader.read() in forwardUserTranscript can exit.
+    // This is a workaround for a race condition in the stream API.
+    this.userTranscriptWriter.close();
+    await this.forwardUserTranscriptTask?.cancelAndWait();
+
```
Shubhrakanti, line 461: @toubatbrian you were right on the read blocking!

PR #580, opened 2025-07-25
Send `lk.segment_id` attribute in transcription events so non final transcriptions can be deduplicated client side

1egoman:
The python agents sdk includes a `lk.segment_id` attribute in all transcription events, which clients ([js sdk example](https://github.com/livekit/components-js/blob/6f02e9f9b3e7737863d256b26a26ade914a5066a/packages/core/src/components/textStream.ts#L79-L85)) can use to deduplicate non finalized events into a list of renderable transcription messages.

However, the agents-js sdk doesn't include this field, and because the `id` field for human-generated transcriptions isn't always the same for transcriptions that represent the same speech, there's no way for that deduplication to occur, which at least in [agents-starter-react](https://github.com/livekit-examples/agent-starter-react) results in rendering each partial transcription distinctly in lists, which is bad:

image

So, to address this, this change does a few things:
- Extends the `transcriptionId` field to be used for tracking human-generated transcriptions as well as agent-generated transcriptions
- Adds the `lk.segment_id` attribute with similar semantics as the python agents sdk (ie, two transcriptions representing the same speech at different times = same segment id), which fixed the above [agents-starter-react](https://github.com/livekit-examples/agent-starter-react) issue
- Extends the `transcriptionId` pattern to the multimodal agent as well (tested my changes with [multimodal-agent-node](https://github.com/livekit-examples/multimodal-agent-node)!)

### DIFF
--- a/.changeset/heavy-cherries-yell.md
+++ b/.changeset/heavy-cherries-yell.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+add `lk.segment_id` to all generated trancription events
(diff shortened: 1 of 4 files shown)


bcherry (approved):
thank you! i now remember that this was changed in python after some testing but we forgot to go back and change it in node

PR #584, opened 2025-07-26
Remove NOT_GIVEN Artifact

toubatbrian:


agents/src/llm/realtime.ts
```diff
@@ -84,7 +84,7 @@ export abstract class RealtimeSession extends EventEmitter {
 
   abstract updateTools(tools: ToolContext): Promise<void>;
 
-  abstract updateOptions(options: { toolChoice?: ToolChoice }): void;
+  abstract updateOptions(options: { toolChoice?: ToolChoice | null }): void;
```
toubatbrian, line 87: Note that in this PR I only made toolChoice opts for the `value?: T | null` notation we discussed.

agents/src/voice/agent_activity.ts
```diff
@@ -797,7 +797,8 @@ export class AgentActivity implements RecognitionHooks {
           userInput: userMessage?.textContent,
           instructions,
           modelSettings: {
-            toolChoice: isGiven(toolChoice) ? toolChoice : undefined,
+            // isGiven(toolChoice) = toolChoice !== undefined
+            toolChoice: toOaiToolChoice(toolChoice !== undefined ? toolChoice : this.toolChoice),
```
toubatbrian, line 801: Made a `toOaiToolChoice` helper. This is because for all the other places such as `ModelSettings`, openai provider constructors, etc, we are still using `ToolChoice | undefined` notation, this PR only changes how we handle undefined in agent activity and will have separate PRs to refactor our entire codebase to adopt for the new notation.

For now, `null` value simply reduces to `undefined`, which means using whatever default tool choice mode from model provider.

agents/src/voice/agent_activity.ts
```diff
@@ -1792,7 +1793,7 @@ export class AgentActivity implements RecognitionHooks {
     }
 
     const originalToolChoice = this.toolChoice;
-    if (toolChoice) {
+    if (toolChoice !== undefined) {
```
toubatbrian, line 1796: This check whether toolChoice is "Given"

Shubhrakanti (approved):

PR #594, opened 2025-08-05
brianyin/ajs-208-google-tts

toubatbrian:


plugins/google/src/beta/gemini_tts.ts
```diff
@@ -0,0 +1,301 @@
...
+      ? {
+          vertexai: true,
+          project: finalProject,
+          location: finalLocation,
+        }
+      : {
+          apiKey: finalApiKey,
+        };
+
+    this.#client = new GoogleGenAI(clientOptions);
```
Shubhrakanti, line 135: @toubatbrian I thought we agreed on no # private variables?

plugins/google/src/beta/gemini_tts.ts
```diff
@@ -0,0 +1,301 @@
...
+      }
+
+      sendLastFrame(true);
+    } catch (error: unknown) {
+      const err = error as {
+        code?: number;
+        message?: string;
+        status?: string;
+        type?: string;
+      };
```
Shubhrakanti, line 261: where are we getting these types from? this is not at all obvious when reading the code. 

We should have comments and be very explicit when casting objects like this.

Shubhrakanti: In python these types look very different. 

https://github.com/livekit/agents/blob/c2d189ad17a83de94f52a2a3c0cf1981825e9bb3/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py#L267-L270

PR #595, opened 2025-08-06
brianyin/ajs-218-fix-chunked-stt-stream-latency

toubatbrian:
Improve chunked STT latency by parallelizing synthesis of tokenized sentences, making each sentence non-blocking from previous sentence synthesis

### DIFF
--- a/plugins/google/src/beta/gemini_tts.ts
+++ b/plugins/google/src/beta/gemini_tts.ts
@@ -7,6 +7,7 @@ import {
   APIConnectionError,
   APIStatusError,
   AudioByteStream,
+  isAPIError,
   shortuuid,
   tts,
 } from '@livekit/agents';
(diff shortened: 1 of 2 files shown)


plugins/google/src/beta/gemini_tts.ts
```diff
@@ -177,82 +178,42 @@ export class ChunkedStream extends tts.ChunkedStream {
...
-        lastFrame = frame;
+    try {
+      for await (const response of responseStream) {
+        await this.#processResponse(response, bstream, requestId);
       }
-
-      sendLastFrame(true);
     } catch (error: unknown) {
+      if (isAPIError(error)) throw error;
+
```
Shubhrakanti, line 216: I don't think there's ever a case where is true. `isAPIError` checks for one of our internal errors, this is never thrown by #processResponse?

PR #600, opened 2025-08-11
brianyin/ajs-220-gemini-realtime-bug-fix

toubatbrian:
Fix gemini realtime input config option so we can have transcriptions by default


plugins/google/src/beta/realtime/realtime_api.ts
```diff
@@ -222,12 +222,12 @@ export class RealtimeModel extends llm.RealtimeModel {
       /**
        * The configuration for input audio transcription
        */
-      inputAudioTranscription?: AudioTranscriptionConfig;
+      inputAudioTranscription?: AudioTranscriptionConfig | null;
```
toubatbrian, line 225: user can explicitly set to `null` to indicate no transcription

PR #607, opened 2025-08-14
Shubhra/ajs 188 fix TTS metrics for 11labs

Shubhrakanti:
Some of the options were configured wrong for the 11labs TTS causing the server to never send an `isFinal` event. The metrics emission depends on the `isFinal` event, so it was never being emitted.


plugins/elevenlabs/src/tts.ts
```diff
@@ -251,7 +262,7 @@ export class SynthesizeStream extends tts.SynthesizeStream {
           }
         }
 
-        ws.send(JSON.stringify({ text: text + ' ', try_trigger_generation: false }));
```
Shubhrakanti, line 254: this was the root cause

PR #610, opened 2025-08-15
Support all types of VideoFrames to LLM Multimodal Inference

toubatbrian:
^ titled

### DIFF
--- a/agents/src/llm/utils.test.ts
+++ b/agents/src/llm/utils.test.ts
@@ -658,29 +658,6 @@ describe('serializeImage', () => {
         expect(result.base64Data).toBeDefined();
       }
     });
-
-    it('should throw error for unsupported YUV formats', async () => {
-      const width = 2;
-      const height = 2;
-      const unsupportedFormats = [
-        VideoBufferType.I420,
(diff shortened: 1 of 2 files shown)


agents/src/llm/utils.test.ts
```diff
@@ -658,29 +658,6 @@ describe('serializeImage', () => {
         expect(result.base64Data).toBeDefined();
       }
     });
-
-    it('should throw error for unsupported YUV formats', async () => {
```
toubatbrian, line 662: removed since now we are able to support all kinds of video buffer type

agents/src/llm/utils.ts
```diff
@@ -79,12 +77,14 @@ export async function serializeImage(image: ImageContent): Promise<SerializedIma
       externalUrl: image.image,
     };
   } else if (image.image instanceof VideoFrame) {
+    const frame = convertToRGBA(image.image);
```
toubatbrian, line 80: this is THE one line magic that converts all frame buffer type to RGBA

Shubhrakanti (approved):

PR #611, opened 2025-08-15
Shubhra/ajs 160 add retry logic and error events

Shubhrakanti:


agents/src/llm/llm.ts
```diff
@@ -117,6 +128,58 @@ export abstract class LLMStream implements AsyncIterableIterator<ChatChunk> {
       this.output.close();
       this.closed = true;
     });
+
+    // this is a hack to immitate asyncio.create_task so that mainTask
+    // is run **after** the constructor has finished. Otherwise we get
+    // runtime error when trying to access class variables in the
+    // `run` method.
+    Promise.resolve().then(() => this.mainTask().then(() => this.queue.close()));
```
Shubhrakanti, line 136: Is there a better way to solve this? If this is unclear I can elaborate more on why we need this.

agents/src/llm/utils.ts
```diff
@@ -208,13 +208,6 @@ export async function executeToolCall(
   }
 }
 
-export function toError(error: unknown): Error {
-  if (error instanceof Error) {
-    return error;
-  }
-  return new Error(String(error));
-}
-
```
Shubhrakanti, line 217: I moved it to global utils and packaged the error object inside the error.

plugins/google/src/beta/gemini_tts.ts
```diff
@@ -223,38 +225,38 @@ export class ChunkedStream extends tts.ChunkedStream {
 
       if (err.code && err.code >= 400 && err.code < 500) {
         if (err.code === 429) {
-          throw new APIStatusError(
-            `Gemini TTS: Rate limit error - ${err.message || 'Unknown error'}`,
-            {
+          throw new APIStatusError({
+            message: `Gemini TTS: Rate limit error - ${err.message || 'Unknown error'}`,
+            options: {
```
Shubhrakanti, line 230: I feel like this reads a lot cleaner with the names parameters.

agents/src/llm/llm.ts
toubatbrian: I see why we are doing this, would be nice to make this a helper function if it's been used in multiple places

agents/src/llm/llm.ts
```diff
@@ -117,6 +128,58 @@ export abstract class LLMStream implements AsyncIterableIterator<ChatChunk> {
...
+              `failed to generate LLM completion, retrying in ${retryInterval}s`,
+            );
+          }
+
+          if (retryInterval > 0) {
+            await delay(retryInterval);
+          }
+        } else {
+          this.emitError({ error: toError(error), recoverable: false });
+          throw error;
```
toubatbrian, line 169: Just a thought, I saw some similar pattern in tts component. Might be better to abstract the retry error handling logics to a helper?

agents/src/_exceptions.ts
```diff
@@ -0,0 +1,134 @@
+/**
+ * Raised when accepting a job but not receiving an assignment within the specified timeout.
+ * The server may have chosen another worker to handle this job.
+ */
+export class AssignmentTimeoutError extends Error {
+  constructor(message = 'Assignment timeout occurred') {
+    super(message);
+    this.name = 'AssignmentTimeoutError';
+    Error.captureStackTrace(this, AssignmentTimeoutError);
```
toubatbrian, line 12: what does this do?

agents/src/llm/llm.ts
Shubhrakanti: Yeah keeping it the same as python for now. But will do this after release.

agents/src/_exceptions.ts
Shubhrakanti: this is the same as doing `raise CustomeException from ParentException` in python

agents/src/llm/llm.ts
Shubhrakanti: Yeah I'll do in a follow up pr once all the retries are merged in.

PR #612, opened 2025-08-16
Add retries to TTS

Shubhrakanti:
Same exact logic as llm

### DIFF
--- a/agents/src/tts/tts.ts
+++ b/agents/src/tts/tts.ts
@@ -171,14 +171,14 @@ export abstract class SynthesizeStream
           } else if (i === this._connOptions.maxRetry) {
             this.emitError({ error, recoverable: false });
             throw new APIConnectionError({
-              message: `failed to generate LLM completion after ${this._connOptions.maxRetry + 1} attempts`,
+              message: `failed to generate TTS completion after ${this._connOptions.maxRetry + 1} attempts`,
               options: { retryable: false },
             });
           } else {
             this.emitError({ error, recoverable: true });
             this.logger.warn(
               { tts: this.#tts.label, attempt: i + 1, error },
-              `failed to generate TTS completion, retrying in ${retryInterval}s`,
+              `failed to synthesize speech, retrying in  ${retryInterval}s`,
             );
           }
 


agents/src/tts/tts.ts
```diff
@@ -117,28 +129,82 @@ export abstract class SynthesizeStream
...
       this.input.close();
       this.output.close();
       this.closed = true;
     });
+
+    // this is a hack to immitate asyncio.create_task so that mainTask
+    // is run **after** the constructor has finished. Otherwise we get
+    // runtime error when trying to access class variables in the
+    // `run` method.
+    Promise.resolve().then(() => this.mainTask().then(() => this.queue.close()));
```
Shubhrakanti, line 157: We do this in three separate places. I think it might be worth moving to a helper method.

agents/src/tts/tts.ts
```diff
@@ -117,28 +129,82 @@ export abstract class SynthesizeStream
...
+        if (error instanceof APIStatusError) {
+          const retryInterval = this._connOptions._intervalForRetry(i);
+
+          if (this._connOptions.maxRetry === 0 || !error.retryable) {
+            this.emitError({ error, recoverable: false });
+            throw error;
+          } else if (i === this._connOptions.maxRetry) {
+            this.emitError({ error, recoverable: false });
+            throw new APIConnectionError({
+              message: `failed to generate LLM completion after ${this._connOptions.maxRetry + 1} attempts`,
```
Shubhrakanti: Ahh shoot need to update the error messages.

agents/src/tts/tts.ts
toubatbrian: For example,
```ts
startSoon(async () => {
  await this.mainTask();
  this.queue.close();
});
```
?

I ok with both

agents/src/tts/tts.ts
toubatbrian: "LLM completion" maybe confusing? Should be "TTS generation"?

agents/src/tts/tts.ts
```diff
@@ -294,14 +362,75 @@ export abstract class ChunkedStream implements AsyncIterableIterator<Synthesized
...
     this.#text = text;
     this.#tts = tts;
+    this._connOptions = connOptions;
 
     this.monitorMetrics();
+
+    // this is a hack to immitate asyncio.create_task so that mainTask
+    // is run **after** the constructor has finished. Otherwise we get
+    // runtime error when trying to access class variables in the
+    // `run` method.
```
toubatbrian, line 382: really nice comment here, helped a lot around the context!

agents/src/tts/tts.ts
Shubhrakanti: yeah this is a mistake

PR #613, opened 2025-08-18
Fix video frame auto-conversion

toubatbrian:
Fix video frame conversion error

### DIFF
--- a/agents/src/llm/utils.ts
+++ b/agents/src/llm/utils.ts
@@ -23,14 +23,39 @@ export interface SerializedImage {
   externalUrl?: string;
 }
 
-const RGBA_NUM_CHANNELS = 4;
+function getChannelsFromVideoBufferType(type: VideoBufferType): 3 | 4 {
+  switch (type) {
+    case VideoBufferType.RGBA:
+    case VideoBufferType.ABGR:
+    case VideoBufferType.ARGB:
+    case VideoBufferType.BGRA:
+      return 4;
+    case VideoBufferType.RGB24:
+      return 3;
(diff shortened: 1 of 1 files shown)


agents/src/llm/utils.ts
```diff
@@ -23,14 +23,39 @@ export interface SerializedImage {
...
+    case VideoBufferType.ARGB:
+    case VideoBufferType.BGRA:
+      return 4;
+    case VideoBufferType.RGB24:
+      return 3;
+    default:
+      // YUV formats (I420, I420A, I422, I444, I010, NV12) need conversion
+      throw new Error(`Unsupported VideoBufferType: ${type}. Only RGB/RGBA formats are supported.`);
+  }
+}
```
toubatbrian, line 39: We still need to keep this helper function since to distinguish between RGB24 and the rest RGB based format.

agents/src/llm/utils.ts
```diff
@@ -23,14 +23,39 @@ export interface SerializedImage {
...
+  // If the frame is already in an RGB/RGBA-compatible format, return it directly
+  if (
+    frame.type === VideoBufferType.RGBA ||
+    frame.type === VideoBufferType.BGRA ||
+    frame.type === VideoBufferType.ARGB ||
+    frame.type === VideoBufferType.ABGR ||
+    frame.type === VideoBufferType.RGB24
+  ) {
+    return frame;
+  }
```
toubatbrian, line 51: `frame.convert` would fail if frame is already RGB based format, so we just return it directly

PR #615, opened 2025-08-18
Shubhra/add stt retries

Shubhrakanti:
Same as #611 but for STTs


plugins/deepgram/src/stt.ts
```diff
@@ -134,11 +134,10 @@ export class SpeechStream extends stt.SpeechStream {
       (duration) => this.onAudioDurationReport(duration),
       { duration: 5.0 },
     );
-
-    this.#run();
   }
 
-  async #run(maxRetry = 32) {
+  protected async run() {
+    const maxRetry = 32;
```
Shubhrakanti, line 140: This seems very random. But I will remove in a separate PR.

PR #616, opened 2025-08-18
Shubhra/refactor hack to start function in next step of the event loop

Shubhrakanti:
Refactor from comments in #611 and #612

### DIFF
--- a/agents/src/utils.ts
+++ b/agents/src/utils.ts
@@ -660,3 +660,13 @@ export function toError(error: unknown): Error {
   }
   throw new InvalidErrorType(error);
 }
+
+/**
+ * This is a hack to immitate asyncio.create_task so that
+ * func will be run after the current event loop iteration.
+ *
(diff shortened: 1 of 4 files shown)


agents/src/utils.ts
```diff
@@ -660,3 +660,13 @@ export function toError(error: unknown): Error {
...
 }
+
+/**
+ * This is a hack to immitate asyncio.create_task so that
+ * func will be run after the current event loop iteration.
+ *
+ * @param func - The function to run.
+ */
+export function startSoon(func: () => void) {
+  Promise.resolve().then(func);
```
Shubhrakanti: I actually think `setTimeout(0, func)` is technically more correct because we don't need to pull this forward to the front of the event loop. 

See https://stackoverflow.com/questions/38752620/promise-vs-settimeout for more details.

theomonnom (approved):

PR #618, opened 2025-08-19
Fix Noisy Logging & Turn Detection Warning Log

toubatbrian:


agents/src/voice/agent_activity.ts
```diff
@@ -416,7 +416,7 @@ export class AgentActivity implements RecognitionHooks {
     if (
       this.llm instanceof RealtimeModel &&
       this.llm.capabilities.turnDetection &&
-      !allowInterruptions
+      allowInterruptions === false
```
toubatbrian, line 419: only warn user if they explicitly set `allowInterruptions` to false

PR #623, opened 2025-08-20
Shubhra/fix cartesia plugin

Shubhrakanti:
We were not waiting for the receive task on the web socket to actually terminate. Now that the queue is properly closed we need to fix that.

### DIFF
--- a/plugins/cartesia/src/tts.ts
+++ b/plugins/cartesia/src/tts.ts
@@ -190,46 +190,74 @@ export class SynthesizeStream extends tts.SynthesizeStream {
     };
 
     const recvTask = async (ws: WebSocket) => {
+      let finalReceived = false;
+      let shouldExit = false;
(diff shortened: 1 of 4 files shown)


plugins/cartesia/src/tts.ts
```diff
@@ -59,10 +61,26 @@ export class TTS extends tts.TTS {
...
       );
     }
+
+    if ((this.#opts.speed || this.#opts.emotion) && this.#opts.model !== 'sonic-2-2025-03-07') {
+      const logger = log();
+      logger.warn(
+        { model: this.#opts.model, speed: this.#opts.speed, emotion: this.#opts.emotion },
+        "speed and emotion controls are only supported for model 'sonic-2-2025-03-07', see https://docs.cartesia.ai/developer-tools/changelog for details",
+      );
+    }
```
toubatbrian, line 71: nice!

toubatbrian (approved):

PR #652, opened 2025-08-26
Update pkg version in README

toubatbrian:


Shubhrakanti (approved):
Can we also remove the beta release (it's near line 31)

PR #676, opened 2025-08-29
Fix Job memory monitoring

Shubhrakanti:
## Problem
Memory monitoring was incorrectly tracking the parent process instead of the actual child process running the job, leading to inaccurate memory usage detection.

## Solution
Added [pidusage](https://www.npmjs.com/package/pidusage) dependency to monitor child process memory by PID
Updated memory monitoring to track actual child process usage
Increased default jobMemoryWarnMB from 300MB to 500MB

## Testing
Monitored restaurant agent over 5 minutes - memory usage hovers around ~300MB, matching Python implementation behavior.

## Dependencies
[ pidusage@^4.0.1](https://www.npmjs.com/package/pidusage), @types/pidusage@^2.0.5

### DIFF
--- a/.changeset/eleven-pandas-provide.md
+++ b/.changeset/eleven-pandas-provide.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Fix Job memory monitoring.


Shubhrakanti:
This is not ready for merge.

agents/src/worker.ts
```diff
@@ -190,7 +190,7 @@ export class WorkerOptions {
     port = undefined,
     logLevel = 'info',
     production = false,
-    jobMemoryWarnMB = 300,
+    jobMemoryWarnMB = 500,
```
Shubhrakanti, line 193: same as python

toubatbrian (approved):

PR #694, opened 2025-09-15
Inference Gateway API

toubatbrian:


turbo.json
```diff
@@ -16,6 +16,9 @@
     "GROQ_API_KEY",
     "LIVEKIT_API_KEY",
     "LIVEKIT_API_SECRET",
+    "LIVEKIT_GATEWAY_API_KEY",
```
Shubhrakanti: Needs to be `LIVEKIT_INFERENCE_*` see https://live-kit.slack.com/archives/C07TWVC0W4A/p1758222314623639?thread_ts=1758179225.445529&cid=C07TWVC0W4A for more details

agents/src/inference/llm.ts
```diff
@@ -0,0 +1,485 @@
...
+    parallelToolCalls?: boolean;
+    toolChoice?: llm.ToolChoice;
+    maxCompletionTokens?: number;
+    baseURL?: string;
+    apiKey?: string;
+    apiSecret?: string;
+    maxRetries?: number;
+    timeout?: number;
+    verbosity?: Verbosity;
+    extraKwargs?: LLMOptions<TModel>;
```
bcherry, line 122: can we just call this `extra` or `extraArgs` and also rename to match on the python side? it's not great to leak the word "kwargs" into TypeScript since that's a Python word cc @longcw @theomonnom

agents/src/inference/index.ts
```diff
@@ -0,0 +1,7 @@
+// SPDX-FileCopyrightText: 2025 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+export { LLM, LLMStream, type LLMModels, type LLMOptions, type OpenAIModels } from './llm.js';
```
bcherry:
```suggestion
import * as llm from './llm.js';
import * as stt from './stt.js';
import * as tts from './tts.js';

export { stt, tts, llm };
export { LLM, LLMStream, type LLMModels, type LLMOptions, type OpenAIModels } from './llm.js';
```

Can we also export the whole llm, stt, and tts modules so that people can access the underlying OpenAiOptions, CerebrasOptions, etc types in generated docs? (we can't just export them normally since CartesiaOptions appears in both STT and TTS)

agents/src/inference/stt.ts
```diff
@@ -0,0 +1,395 @@
...
+
+export interface AssemblyaiOptions {
+  format_turns?: boolean; // default: false
+  end_of_turn_confidence_threshold?: number; // default: 0.01
+  min_end_of_turn_silence_when_confident?: number; // default: 0
+  max_turn_silence?: number; // default: not specified
+  keyterms_prompt?: string[]; // default: not specified
+}
+
+export type STTModels = DeepgramModels | CartesiaModels | AssemblyaiModels;
```
bcherry: I think you need `| string` here or at consumption sites so people can adopt new models added to the service that aren't yet added to the type enums (same for the other types)

bcherry:
I see a bunch of this error in my terminal when i run the starter agent in dev with the gateway integrated

```
(node:50680) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 abort listeners added to [AbortSignal]. MaxListeners is 10. Use events.setMaxListeners() to increase limit
```

Shubhrakanti (approved):

Shubhrakanti:
> I see a bunch of this error in my terminal when i run the starter agent in dev with the gateway integrated
> 
> 
> 
> ```
> 
> (node:50680) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 abort listeners added to [AbortSignal]. MaxListeners is 10. Use events.setMaxListeners() to increase limit
> 
> ```

@toubatbrian Maybe too many Tasks?

agents/src/inference/llm.ts
toubatbrian: This would involves refactoring all the base class in TS side as well, let's make it a separate PR. I'll add a ticket

agents/src/inference/stt.ts
toubatbrian: My main concern is type hint. If we add `string` to the type user will no longer get type hints on list of supporting model name to choose from when they specify the model string in their IDE. I added this

```
export type CustomModelType = `custom/${string}`;
```

for specifying unsupported enum. This would still gives user type hints on existing model list and also allow them to define custom model:
```
      llm: 'custom/my-vew-unsupported-model',
```

Let me know what's your thought on this

agents/src/inference/stt.ts
```diff
@@ -0,0 +1,414 @@
...
+  | 'deepgram/nova-3-general'
+  | 'deepgram/nova-3-medical'
+  | 'deepgram/nova-2'
+  | 'deepgram/nova-2-general'
+  | 'deepgram/nova-2-medical'
+  | 'deepgram/nova-2-phonecall';
+
+export type CartesiaModels = 'cartesia' | 'cartesia/ink-whisper';
+
+export type AssemblyaiModels = 'assemblyai' | 'assemblyai/universal-streaming';
```
toubatbrian, line 32: https://live-kit.slack.com/archives/C08SCFCHKGD/p1758557081674519 @bcherry let me know if this is ok

agents/src/inference/stt.ts
bcherry: it's not for custom models, it's for when we add a new model to the service like `xai/grok-4`. devs should be able to support it without requiring an SDK update - that's actually one of the key benefits of the gateway. if we actually had to choose between working intellisense and supporting future models, we should choose supporting future models IMO.

But I asked ChatGPT and it suggested `| (string & NonNullable<unknown>)` to solve this, which in my quick test seems to preserve intellisense while also not giving errors on new unlisted types.

what do you think?

agents/src/inference/llm.ts
```diff
@@ -0,0 +1,490 @@
...
+  APIStatusError,
+  APITimeoutError,
+  DEFAULT_API_CONNECT_OPTIONS,
+  toError,
+} from '../index.js';
+import * as llm from '../llm/index.js';
+import type { APIConnectOptions } from '../types.js';
+import { type CustomModelType, createAccessToken, getModelName } from './utils.js';
+
+export type OpenAIModels =
```
bcherry: I realized we should rename this to `AzureModels` (and also rename the options type too) to be more clear and future-proof for if we do add first-party openai platform as a model provider later.

```suggestion
export type AzureModels =
```

we should do the same in python too @longcw IMO

bcherry (approved):

PR #695, opened 2025-09-15
Gateway LLM

toubatbrian:


agents/package.json
```diff
@@ -44,12 +44,13 @@
   },
   "dependencies": {
     "@livekit/mutex": "^1.1.1",
-    "@livekit/protocol": "^1.29.1",
+    "@livekit/protocol": "^1.41.0",
```
toubatbrian, line 47: upgrade to resolve competibility issue

agents/src/inference/llm.ts
```diff
@@ -0,0 +1,426 @@
...
+    opts?: {
+      temperature?: number;
+      topP?: number;
+      parallelToolCalls?: boolean;
+      toolChoice?: llm.ToolChoice;
+      maxCompletionTokens?: number;
+      baseURL?: string;
+      apiKey?: string;
+      apiSecret?: string;
+      maxRetries?: number;
```
toubatbrian, line 54: most of code was moved from openai `llm.ts`, change few places for making gateway llm to be competible

plugins/openai/src/llm.ts
```diff
@@ -491,230 +484,18 @@ export class LLM extends llm.LLM {
       extras.tool_choice = toolChoice;
     }
 
-    return new LLMStream(this, {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return new LLMStream(this as any, {
```
Shubhrakanti, line 488: why is the typing off here?

plugins/openai/src/llm.ts
```diff
@@ -2,14 +2,7 @@
...
 import type { APIConnectOptions } from '@livekit/agents';
-import {
-  APIConnectionError,
-  APIStatusError,
-  APITimeoutError,
-  DEFAULT_API_CONNECT_OPTIONS,
-  llm,
-  toError,
-} from '@livekit/agents';
+import { DEFAULT_API_CONNECT_OPTIONS, inference, llm } from '@livekit/agents';
```
Shubhrakanti, line 5: why are we changing the open ai plugins for gateway?

agents/src/inference/llm.ts
```diff
@@ -0,0 +1,426 @@
...
+  }: {
+    chatCtx: llm.ChatContext;
+    toolCtx?: llm.ToolContext;
+    connOptions?: APIConnectOptions;
+    parallelToolCalls?: boolean;
+    toolChoice?: llm.ToolChoice;
+    // TODO(AJS-270): Add responseFormat parameter
+    extraKwargs?: Record<string, any>;
+  }): LLMStream {
+    let extras: Record<string, any> = { ...(extraKwargs || {}) }; // eslint-disable-line @typescript-eslint/no-explicit-any
```
Shubhrakanti: unknown ? why any?

Shubhrakanti (approved):

plugins/openai/src/llm.ts
toubatbrian: `this` is subclassed from the base class and have some different fields. JS is complaining about it even if should still work.

plugins/openai/src/llm.ts
toubatbrian: I followed @longcw's PR. Basically we can reuse gateway llm to openai plugin (since gateway llm is already openai-competible).

PR #701, opened 2025-09-16
Add Initial Cursor Rules

toubatbrian:


toubatbrian:
>I am not sure if we want to add this to the repo, is this helpful for users? and another question, how the project structures, examples in the rule will be maintained?

Good question! This is mainly to help developers by reducing the amount of context we need to provide the agent before starting certain tasks.

For project structure and examples, the rule files can be updated as needed, and we can also re-run the tree command periodically to regenerate the structure. Even if it’s not 100% up to date, missing one or two files shouldn’t significantly impact the LLM. imo it’s still more useful to have it than not.

PR #703, opened 2025-09-17
Gateway STT

toubatbrian:


agents/src/utils.ts
```diff
@@ -813,3 +813,14 @@ export async function waitForTrackPublication({
     room.off(RoomEvent.TrackPublished, onTrackPublished);
   }
 }
+
+export async function waitForAbort(signal: AbortSignal) {
+  const abortFuture = new Future<void>();
+  const handler = () => {
+    abortFuture.resolve();
+    signal.removeEventListener('abort', handler);
```
toubatbrian, line 825: helper function (also been used in a next PR)

PR #713, opened 2025-09-19
brianyin/agt-2100-rename-worker-to-agentserver

Shubhrakanti:
Simple rename. 

TODO: 

- [ ] Keep aliases for backwards compatibility

### DIFF
--- a/.changeset/open-mice-bathe.md
+++ b/.changeset/open-mice-bathe.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Rename Worker to AgentServer


Shubhrakanti:
@theomonnom can you link the PR on python side once it's done so I can port it over?

agents/src/generator.ts
```diff
@@ -33,6 +33,11 @@ export function isAgent(obj: unknown): obj is Agent {
  * })
  * ```
  */
-export function defineAgent(agent: Agent): Agent {
+export function defineAgentServer(agent: Agent): Agent {
```
toubatbrian: @theomonnom @Shubhrakanti What do you guys think about renaming this?

toubatbrian: bump

theomonnom: I think its fine to keep defineAgent

PR #725, opened 2025-09-24
brianyin/ajs-282-race-condition-having-generatereply-inside-tool

toubatbrian:
## Description
Resolve race condition of having `generateReply` inside tool call.

## Changes Made
Always wait for all `generateReply` inside tool call to be scheduled first, before dispatching the speech handle of the tool response

## Pre-Review Checklist

## Testing

- [x] All tests pass
- [x] Make sure both `restaurant_agent.ts` and `realtime_agent.ts` work properly (for major changes)

### DIFF
--- a/examples/src/basic_tool_call_agent.ts
+++ b/examples/src/basic_tool_call_agent.ts
@@ -53,16 +53,6 @@ export default defineAgent({
         ctx.session.generateReply({
           userInput: 'Tell user you are queuing the request. Counting down from 3 to 1.',
         });
-        await ctx.waitForPlayout();
-
-        // Mock API call to fetch some data
-        try {
-          const response = await fetch('https://jsonplaceholder.typicode.com/posts/1');
-          const data = await response.json();
-          console.log('Fetched mock data', { data });
-        } catch (error) {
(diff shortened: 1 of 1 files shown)


agents/src/voice/agent.ts
```diff
@@ -21,8 +21,21 @@ import { SynthesizeStream, StreamAdapter as TTSStreamAdapter } from '../tts/inde
...
-export const asyncLocalStorage = new AsyncLocalStorage<{ functionCall?: FunctionCall }>();
+export type ActiveToolCall = {
+  functionCall: FunctionCall;
+  speechHandles: SpeechHandle[];
+};
+
+export function createActiveToolCall(functionCall: FunctionCall): ActiveToolCall {
+  return {
+    functionCall,
+    speechHandles: [],
```
toubatbrian, line 34: We keep track of all `generateReply` created inside tool call so we can wait them to be scheduled

Shubhrakanti (commented):
1. Can you add why the race condition was happening in the first place? 
2. This solution seems to diverge from the python implementation? Is there a way we can solve this similar to how the python side does it?

PR #729, opened 2025-09-25
Shubhra/clean up real time model logs

Shubhrakanti:
Toggle openAI realtime session logs with `LK_OPENAI_DEBUG` env variable. They are very noisy and should be enabled for very specific debugging purposes.

I also cleaned up the logs to keep the events as metadata so it's 100x easier to read them.

### DIFF
--- a/.changeset/every-moons-search.md
+++ b/.changeset/every-moons-search.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-openai': patch
+---
+
+update debug logs to be enabled by env variable


plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -640,11 +643,8 @@ export class RealtimeSession extends llm.RealtimeSession {
     } as api_proto.ConversationItemTruncateEvent);
   }
 
-  /// Truncates the data field of the event to the specified maxLength to avoid overwhelming logs
```
Shubhrakanti, line 643: raw audio data is useless to look at

PR #738, opened 2025-09-30
Add changesets workaround to prevent peerdependency updates to cause …

lukasIO:
…major bump

## Description
adds a workaround for https://github.com/changesets/changesets/issues/1011.

Bumping `@livekit/agents` with a minor tag causes all plugins to be bumped with a major tag without this workaround


patches/@changesets__assemble-release-plan.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/dist/changesets-assemble-release-plan.cjs.js b/dist/changesets-assemble-release-plan.cjs.js
```
toubatbrian, line 1: We need to add the copyright text at the top of this .patch file. Would that be valid syntax-wise?

lukasIO: I'll add `patches` to exclude it from the reuse check

PR #740, opened 2025-09-30
Shubhra/update comments

Shubhrakanti:
Update the comments to follow proper docstring format

### DIFF
--- a/agents/src/voice/room_io/room_io.ts
+++ b/agents/src/voice/room_io/room_io.ts
@@ -475,7 +475,6 @@ export class RoomIO {
     this.initTask = Task.from((controller) => this.init(controller.signal));
 
     /** attach the agent to the session */
-    // Note: Fixed spelling from 'attatch' to 'attach'
     if (this.audioInput) {
       this.agentSession.input.audio = this.audioInput;
     }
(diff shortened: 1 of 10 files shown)


agents/src/voice/room_io/room_io.ts
```diff
@@ -473,7 +474,8 @@ export class RoomIO {
 
     this.initTask = Task.from((controller) => this.init(controller.signal));
 
-    // -- attatch the agent to the session --
+    /** attach the agent to the session */
+    // Note: Fixed spelling from 'attatch' to 'attach'
```
bcherry: note: you can remove the comments cursor adds to your code :)

Shubhrakanti: oh my god. This is horrible. my bad.

agents/src/worker.ts
```diff
@@ -279,7 +279,7 @@ export class Worker {
   #logger = log().child({ version });
   #inferenceExecutor?: InferenceProcExecutor;
 
-  /* @throws {@link MissingCredentialsError} if URL, API key or API secret are missing */
+  /** @throws MissingCredentialsError if URL, API key or API secret are missing */
```
bcherry: is it correct to remove the links? you should try building the docs locally and seeing if they work

Shubhrakanti: yeah we shouldn't remove these

PR #741, opened 2025-09-30
Add language to UserInputTranscribed event

Shubhrakanti:
Add language field to `UserInputTranscribed` Event

### DIFF
--- a/agents/src/voice/events.ts
+++ b/agents/src/voice/events.ts
@@ -80,25 +80,31 @@ export type UserInputTranscribedEvent = {
   type: 'user_input_transcribed';
   transcript: string;
   isFinal: boolean;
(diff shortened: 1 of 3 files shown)


agents/src/voice/events.ts
```diff
@@ -80,25 +80,31 @@ export type UserInputTranscribedEvent = {
   type: 'user_input_transcribed';
   transcript: string;
   isFinal: boolean;
+  // TODO(AJS-106): add multi participant support
+  /** Not supported yet. Always null by default. */
   speakerId: string | null;
```
Shubhrakanti, line 85: shouldn't have added this in the first place :/

theomonnom (approved):

PR #742, opened 2025-10-01
brianyin/ajs-291-update-inference-api-update-model-names

toubatbrian:


agents/src/inference/llm.ts
```diff
@@ -13,124 +13,104 @@ import * as llm from '../llm/index.js';
...
+
+export interface InferenceLLMOptions {
+  model: LLMModels;
+  provider?: string;
   baseURL: string;
   apiKey: string;
   apiSecret: string;
-  verbosity?: Verbosity;
-  extraKwargs: Record<string, any>; // eslint-disable-line @typescript-eslint/no-explicit-any
+  extraKwargs: ChatCompletionOptions;
```
davidzhao: I think we should change this to whatever is native for JS? `extra_kwargs` is really a python term. is it `modelOptions` for JS?

agents/src/inference/llm.ts
```diff
@@ -13,124 +13,104 @@ import * as llm from '../llm/index.js';
...
+
+export type DeepSeekModels = 'deepseek-ai/deepseek-v3';
+
+type ChatCompletionPredictionContentParam = OpenAI.Chat.Completions.ChatCompletionPredictionContent;
+type WebSearchOptions = OpenAI.Chat.Completions.ChatCompletionCreateParams.WebSearchOptions;
+type ToolChoice = OpenAI.Chat.Completions.ChatCompletionCreateParams['tool_choice'];
+type Verbosity = 'low' | 'medium' | 'high';
+
+export interface ChatCompletionOptions extends Record<string, unknown> {
+  frequency_penalty?: number;
```
davidzhao, line 43: should these be snake_case?

toubatbrian: I'm keeping it as-is for now since it gets sent to the server side, and I want it to match the parameter schema in Python. Open to make a follow-up PR that allows users to pass camelCase, and we can handle the conversion to snake_case internally.

PR #750, opened 2025-10-06
brianyin/ajs-282-race-condition-having-generatereply-inside-tool

toubatbrian:


toubatbrian:
Close in favor of https://github.com/livekit/agents-js/pull/752

PR #753, opened 2025-10-07
Changeset for realtime playout fix

toubatbrian:


Shubhrakanti (approved):

agents/src/voice/agent_activity.ts
```diff
@@ -1783,9 +1783,9 @@ export class AgentActivity implements RecognitionHooks {
           !this.currentSpeech.done() &&
           this.currentSpeech !== speechHandle
         ) {
-          this.logger.info(
```
Shubhrakanti, line 1786: We can get rid of this too. Not super important.

Shubhrakanti: Good catch

Shubhrakanti (approved):

PR #777, opened 2025-10-20
brianyin/ajs-306-fix-commonjs-entrypoint-to-resolve-files-from-dist

toubatbrian:
### Summary
Fix the CommonJS export map so `require('@livekit/agents')` resolves to the built files in `dist`.

### Background
In Node 22+ the package export map is authoritative for `require`. Our export map still pointed at `./index.cjs`, which doesn’t exist in the published package, so `require('@livekit/agents')` failed with `MODULE_NOT_FOUND` (it was looking for `node_modules/@livekit/agents/index.cjs`). Updating the export map to `./dist/index.cjs`/`./dist/index.d.cts` fixes the regression.

### Testing
- `node -p "require('./packages/agents/dist').version"`

### DIFF
--- a/agents/package.json
+++ b/agents/package.json
@@ -11,8 +11,8 @@
       "default": "./dist/index.js"
     },
     "require": {
-      "types": "./index.d.cts",
-      "default": "./index.cjs"
+      "types": "./dist/index.d.cts",
+      "default": "./dist/index.cjs"
     }
   },
   "author": "LiveKit",
--- a/.changeset/olive-roses-brush.md
+++ b/.changeset/olive-roses-brush.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+- Fix CommonJS entrypoint to resolve files from `dist`


agents/package.json
```diff
@@ -11,8 +11,8 @@
       "default": "./dist/index.js"
     },
     "require": {
-      "types": "./index.d.cts",
-      "default": "./index.cjs"
+      "types": "./dist/index.d.cts",
+      "default": "./dist/index.cjs"
```
Shubhrakanti, line 15: Just curious how does this actually fix the issue? Also can you update the PR description with what the actually issue/error is?

Shubhrakanti (approved):
Description for this one. It's not obvious what the issue is based on reading the PR.

agents/package.json
toubatbrian: @livekit/agents advertises require: "dist/index.cjs" in its top-level fields, but the exports.require map still pointed to ./index.cjs at the package root. In Node 22+ CJS resolution always consults the export map, so require('@livekit/agents') looked for node_modules/@livekit/agents/index.cjs and threw MODULE_NOT_FOUND.

The fix updates the export map to resolve ./dist/index.cjs and ./dist/index.d.cts. After the change, require('@livekit/agents') succeeds again because Node now reaches the built CommonJS bundle in dist.

toubatbrian: Updated PR description

PR #786, opened 2025-10-24
Fix race condition causing "Writer is not bound to a WritableStream" error in Silero VAD

toubatbrian:
## Problem

The Silero VAD plugin was intermittently throwing unhandled `TypeError: Invalid state: Writer is not bound to a WritableStream` errors when plugin close. 

## Root Cause

A race condition exists between the VAD inference loop and the stream shutdown process:

1. The VAD task runs in a continuous loop, checking `while (!this.closed)` before processing frames
2. When a participant disconnects, the base class `close()` method:
   - Calls `outputWriter.releaseLock()` to unbind the writer from the stream
   - Sets `closed = true`
3. These operations are not atomic, creating a window where the inference loop can pass the `!this.closed` check but then attempt to write to an already-released writer

**Race Condition Timeline**:
```
VAD Task                     |  Close Task
-----------------------------|---------------------------
Check: !this.closed → true   |
                             |  outputWriter.releaseLock()
                             |  closed = true
outputWriter.write(...) ❌   |
→ Writer is not bound!       |
```

## Solution

Implemented graceful error handling following the established pattern in `deferred_stream.ts`.

### DIFF
--- a/agents/src/vad.ts
+++ b/agents/src/vad.ts
@@ -191,10 +191,6 @@ export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
       this.outputWriter.write(event);
       return true;
     } catch (e) {
-      // Ignore writer release errors during stream closure (e.g., on participant disconnect)
-      if (isWriterReleaseError(e)) {
-        return false; // Signal that the stream is closing
-      }
       throw e;
     }
   }
@@ -259,29 +255,4 @@ export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
   [Symbol.asyncIterator](): VADStream {
     return this;
   }
-}
-
-/**
- * Helper function to check if an error is related to writer being released during stream closure.
- * This can happen during shutdown when close() releases the writer while the VAD task is still running.
- *
- * Handles Node.js stream errors with code ERR_INVALID_STATE:
- * - "Invalid state: Writer is not bound to a WritableStream" (after releaseLock())
- * - "Invalid state: WritableStream is closed" (after close())
- */
-function isWriterReleaseError(e: unknown): boolean {
-  if (e instanceof TypeError) {
-    // Check for ERR_INVALID_STATE error code (most reliable)
-    if ('code' in e && e.code === 'ERR_INVALID_STATE') {
-      return true;
-    }
-
-    // Fallback to message checking for compatibility
-    const message = e.message;
-    return (
-      message.includes('Writer is not bound to a WritableStream') ||
-      message.includes('WritableStream is closed')
-    );
-  }
-  return false;
 }


Shubhrakanti (commented):
You mention that the `Close Thread` calls `outputWriter.releaseLock() .... closed = true` but we don't multi thread? It's asynchronous. Are you sure this is the root cause?

Are you able to reliably reproduce this issue? If so can you list the steps?

plugins/silero/src/vad.ts
```diff
@@ -16,6 +17,31 @@ import { OnnxModel, newInferenceSession } from './onnx_model.js';
...
+
+    // Fallback to message checking for compatibility
+    const message = e.message;
+    return (
+      message.includes('Writer is not bound to a WritableStream') ||
+      message.includes('WritableStream is closed')
+    );
+  }
+  return false;
+}
```
theomonnom: Isn't this hacky? Should we have another mechanism of synchronization?

toubatbrian: We have similar approach in `defered_stream.ts` https://github.com/livekit/agents-js/blob/7fc7808ae33619d7bf80d6d31ead7ffc66f2fc34/agents/src/stream/deferred_stream.ts#L17-L28

toubatbrian: The error we're catching (ERR_INVALID_STATE: Writer is not bound) is essentially Node.js telling us the stream is shutting down. It's not really an exceptional condition, it's an expected part of the shutdown sequence that we need to handle gracefully. The safeWriteEvent() method makes this explicit: it catches expected shutdown errors and signals the loop to exit cleanly, while still throwing any truly unexpected errors.

Shubhrakanti: Could this happen in llm/stt/tts as well?

Shubhrakanti (changes requested):
@toubatbrian I _think_ this is what you mean. We have a to await the [model to actually run](https://github.com/livekit/agents-js/blob/7fc7808ae33619d7bf80d6d31ead7ffc66f2fc34/plugins/silero/src/vad.ts#L221-L223) during which the [outputWriter](https://github.com/livekit/agents-js/blob/7fc7808ae33619d7bf80d6d31ead7ffc66f2fc34/agents/src/vad.ts#L91) might be closed.

(model run)
https://github.com/livekit/agents-js/blob/7fc7808ae33619d7bf80d6d31ead7ffc66f2fc34/plugins/silero/src/vad.ts#L221-L223

(output writer definition)
https://github.com/livekit/agents-js/blob/7fc7808ae33619d7bf80d6d31ead7ffc66f2fc34/agents/src/vad.ts#L91

Is there a simpler solution where the [base vad class](https://github.com/livekit/agents-js/blob/7fc7808ae33619d7bf80d6d31ead7ffc66f2fc34/agents/src/vad.ts#L85) has a method that wraps `this.outputWriter.write` with a `if (!this.closed)`? Then every plugin just calls this method and doesn't have to worry about the internals of `outputWriter`?

toubatbrian:
> Is there a simpler solution where the [base vad class](https://github.com/livekit/agents-js/blob/7fc7808ae33619d7bf80d6d31ead7ffc66f2fc34/agents/src/vad.ts#L85) has a method that wraps this.outputWriter.write with a if (!this.closed)? Then every plugin just calls this method and doesn't have to worry about the internals of outputWriter?

@Shubhrakanti Great question! Let's definitely explore that idea later. For now let's focus on fixing the VAD issue within the scope of this PR to unblock our customer first. We can revisit and discuss a cleaner solution afterward.

Shubhrakanti:
> > Is there a simpler solution where the [base vad class](https://github.com/livekit/agents-js/blob/7fc7808ae33619d7bf80d6d31ead7ffc66f2fc34/agents/src/vad.ts#L85) has a method that wraps this.outputWriter.write with a if (!this.closed)? Then every plugin just calls this method and doesn't have to worry about the internals of outputWriter?
> 
> @Shubhrakanti Great question! Let's definitely explore that idea later. For now let's focus on fixing the VAD issue within the scope of this PR to unblock our customer first. We can revisit and discuss a cleaner solution afterward.

All we need to do is add one method to the base vad class 

```typescript 
writeOutput(event: VADEvent) {
    if (!this.closed) {
       this.outputWriter.write(event)
    } else {
     // you can either throw an error or have a boolean which returns if the write was successful or not. Or just do nothing and the next iteration of `this.#task` in silerio.vad the main loop will break since `this.close == true`
   ....
}
```


and then you replace `this.outputWriter.write(` -> `this.writeOutput` in `sliero/vad.ts`. 


It's fine if this goes out Mondaynext week. I'd rather have it done correctly.

agents/src/vad.ts
```diff
@@ -177,6 +177,28 @@ export abstract class VADStream implements AsyncIterableIterator<VADEvent> {
...
+    if (this.closed) {
+      return false;
+    }
+
+    try {
+      this.outputWriter.write(event);
+      return true;
+    } catch (e) {
+      // Ignore writer release errors during stream closure (e.g., on participant disconnect)
+      if (isWriterReleaseError(e)) {
```
Shubhrakanti: If `this.closed` is `false` then the writer should not be closed right? We should never see this error?

plugins/silero/src/vad.ts
```diff
@@ -260,26 +260,30 @@ export class VADStream extends baseStream {
...
-                1,
-                toCopyInt,
-              ),
-            ],
-            speaking: pubSpeaking,
-            rawAccumulatedSilence: silenceThresholdDuration,
-            rawAccumulatedSpeech: speechThresholdDuration,
-          });
+          if (
+            !this.sendVADEvent({
```
Shubhrakanti, line 264: nit: might be easier to just continue the loop and then we will automatically break out in the next iteration of the main loop (`while (this.closed)`).

agents/src/vad.ts
toubatbrian: I think we can drop this, but I’m keeping it for now as an extra safety measure.

Shubhrakanti: Let's drop it for now. If we see this error again that means the writer state is out of sync with `this.closed`. We want to know that so we can fix it.

plugins/silero/src/vad.ts
```diff
@@ -314,19 +318,23 @@ export class VADStream extends baseStream {
...
-                speechDuration: pubSpeechDuration,
-                probability: p,
-                inferenceDuration,
-                frames: [copySpeechBuffer()],
-                speaking: pubSpeaking,
-                rawAccumulatedSilence: 0,
-                rawAccumulatedSpeech: 0,
-              });
+              if (
+                !this.sendVADEvent({
```
Shubhrakanti, line 322:
```typescript 

if (!this.sendVADEvent) {

    continue
}

```


Feels redundant? We can just do `this.sendVADEvent`? We don't need a return type?

Shubhrakanti (approved):
Please look at the two comments before merging.

agents/src/vad.ts
Shubhrakanti: @toubatbrian bump

plugins/silero/src/vad.ts
Shubhrakanti: @toubatbrian bump

agents/src/vad.ts
toubatbrian: ok

plugins/silero/src/vad.ts
toubatbrian: If `this.sendVADEvent` return false, that means the output writer is closed and we need to jump out of the loop, so I think this is not redundant?

Shubhrakanti: yeah it's fine for now

PR #798, opened 2025-10-31
brianyin/ajs-314-preemptive-generation

toubatbrian:
Add preemptive generation, allowing llm / tts generation to start before EOU waits to completion.

### DIFF
--- a/agents/src/stt/stt.ts
+++ b/agents/src/stt/stt.ts
@@ -32,18 +32,18 @@ export enum SpeechEventType {
    */
   FINAL_TRANSCRIPT = 2,
   /**
+   * Indicate the end of speech, emitted when the user stops speaking.
+   * The first alternative is a combination of all the previous FINAL_TRANSCRIPT events.
+   */
+  END_OF_SPEECH = 3,
+  /** Usage event, emitted periodically to indicate usage metrics. */
+  RECOGNITION_USAGE = 4,
+  /**
    * Preflight transcript, emitted before final transcript when STT has high confidence
    * but hasn't fully committed yet. Includes all pre-committed transcripts including
    * final transcript from the previous STT run.
    */
-  PREFLIGHT_TRANSCRIPT = 3,
-  /**
-   * Indicate the end of speech, emitted when the user stops speaking.
-   * The first alternative is a combination of all the previous FINAL_TRANSCRIPT events.
-   */
-  END_OF_SPEECH = 4,
-  /** Usage event, emitted periodically to indicate usage metrics. */
-  RECOGNITION_USAGE = 5,
+  PREFLIGHT_TRANSCRIPT = 5,
 }
 
 /** SpeechData contains metadata about this {@link SpeechEvent}. */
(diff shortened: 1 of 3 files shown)


agents/src/stt/stt.ts
```diff
@@ -31,13 +31,19 @@ export enum SpeechEventType {
...
+  PREFLIGHT_TRANSCRIPT = 3,
   /**
    * Indicate the end of speech, emitted when the user stops speaking.
    * The first alternative is a combination of all the previous FINAL_TRANSCRIPT events.
    */
-  END_OF_SPEECH = 3,
+  END_OF_SPEECH = 4,
   /** Usage event, emitted periodically to indicate usage metrics. */
-  RECOGNITION_USAGE = 4,
+  RECOGNITION_USAGE = 5,
```
Shubhrakanti: Are these changes backwards compatible? Do we need a minor version bump? 

cc: @theomonnom

agents/src/voice/agent_session.ts
```diff
@@ -67,6 +68,7 @@ const defaultVoiceOptions: VoiceOptions = {
   minEndpointingDelay: 500,
   maxEndpointingDelay: 6000,
   maxToolSteps: 3,
+  preemptiveGeneration: false,
```
Shubhrakanti, line 71: We should have this on by default right?

agents/src/voice/agent_session.ts
```diff
@@ -421,7 +423,8 @@ export class AgentSession<
   }
 
   /** @internal */
-  _updateUserState(state: UserState) {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Shubhrakanti: we should just have an `_` prefix? Let's remove line level ignores as much as possible

agents/src/voice/audio_recognition.ts
```diff
@@ -162,34 +175,139 @@ export class AudioRecognition {
...
+          // vad disabled, use stt timestamp
+          this.lastSpeakingTime = Date.now();
+        }
+
+        if (this.turnDetectionMode !== 'manual' || this.userTurnCommitted) {
+          const confidenceVals = [...this.finalTranscriptConfidence, preflightConfidence];
+          this.logger.debug(
+            { transcript: this.audioPreflightTranscript },
+            'triggering preemptive generation (PREFLIGHT_TRANSCRIPT)',
+          );
```
Shubhrakanti, line 263: I think we want to actually print the transcript - truncated with `...` if it's too log?

agents/src/stt/stt.ts
toubatbrian: It should be? Assume user access the event through the actual type STTEvent instead of raw number.. but technically they could just pass a single number, say they want to listen to certain STT events, and this change will break it. I think you are right I will re-order the new event at the end to keep those enum numbers unchanged.

agents/src/voice/agent_session.ts
toubatbrian: In python this is disabled by default, so I guess we should follow the same.

PR #808, opened 2025-11-06
Add AgentHandoff Chat Item

toubatbrian:
Implement `AgentHandoffItem` into chat context

### DIFF
--- a/agents/src/voice/agent_session.ts
+++ b/agents/src/voice/agent_session.ts
@@ -372,7 +372,6 @@ export class AgentSession<
     this.activity = this.nextActivity;
     this.nextActivity = undefined;
 
-    // Insert agent handoff into chat context
     this._chatCtx.insert(
       new AgentHandoffItem({
         oldAgentId: previousActivity?.agent.id,


toubatbrian:
Hey @simllll, those events will be recorded and sent to the observability for traces. It's been filtered before passing to LLM since LLM does not support a "handoff" message type.

agents/src/voice/agent.ts
```diff
@@ -59,6 +59,7 @@ export interface ModelSettings {
 }
 
 export interface AgentOptions<UserData> {
+  id?: string;
   instructions: string;
```
Shubhrakanti, line 63: why is this optional?

agents/src/voice/agent_session.ts
```diff
@@ -370,6 +372,15 @@ export class AgentSession<
     this.activity = this.nextActivity;
     this.nextActivity = undefined;
 
+    // Insert agent handoff into chat context
```
Shubhrakanti: This comment just says what the function does

agents/src/voice/agent.ts
toubatbrian: This also follows python agents:
```py
class Agent:
    def __init__(
        self,
        *,
        instructions: str,
        id: str | None = None,
        chat_ctx: NotGivenOr[llm.ChatContext | None] = NOT_GIVEN,
        tools: list[llm.FunctionTool | llm.RawFunctionTool] | None = None,
        turn_detection: NotGivenOr[TurnDetectionMode | None] = NOT_GIVEN,
        stt: NotGivenOr[stt.STT | STTModels | str | None] = NOT_GIVEN,
        vad: NotGivenOr[vad.VAD | None] = NOT_GIVEN,
        llm: NotGivenOr[llm.LLM | llm.RealtimeModel | LLMModels | str | None] = NOT_GIVEN,
        tts: NotGivenOr[tts.TTS | TTSModels | str | None] = NOT_GIVEN,
        mcp_servers: NotGivenOr[list[mcp.MCPServer] | None] = NOT_GIVEN,
        allow_interruptions: NotGivenOr[bool] = NOT_GIVEN,
        min_consecutive_speech_delay: NotGivenOr[float] = NOT_GIVEN,
        use_tts_aligned_transcript: NotGivenOr[bool] = NOT_GIVEN,
        min_endpointing_delay: NotGivenOr[float] = NOT_GIVEN,
        max_endpointing_delay: NotGivenOr[float] = NOT_GIVEN,
    ) -> None:
        tools = tools or []
        if type(self) is Agent:
            self._id = "default_agent"
        else:
            self._id = id or misc.camel_to_snake_case(type(self).__name__)

```

toubatbrian:
Hey @simllll, I see your thoughts, which totally make sense! The main goal is to trying to achieve as much as parity as possible with the python agent framework. Adding handoff object into chat context is what currently been implemented in python side. 

The other reason tie-ing the handoff to chat context is that, we'll add support for export the context json once a session is finished, so developer can do certain things on it such as eval. Having the agent handoff info as part of the chat context would be useful for that case.

PR #813, opened 2025-11-07
brianyin/ajs-320-bump-openai-to-6.x

toubatbrian:
Bump openai to 6.x, Re-enable critical LLM tests to make sure nothing breaks

### DIFF
--- a/.changeset/grumpy-parks-stand.md
+++ b/.changeset/grumpy-parks-stand.md
@@ -1,0 +1,8 @@
+---
+'@livekit/agents-plugin-google': patch
+'@livekit/agents-plugin-openai': patch
+'@livekit/agents-plugins-test': patch
+'@livekit/agents': patch
+---
+
+bump openai to 6.x


agents/src/utils.ts
```diff
@@ -15,6 +15,22 @@ import { TransformStream, type TransformStreamDefaultController } from 'node:str
...
+  ? T
+  : T extends object
+    ? T extends Array<infer U>
+      ? Array<Expand<U>>
+      : T extends Map<infer K, infer V>
+        ? Map<Expand<K>, Expand<V>>
+        : T extends Set<infer M>
+          ? Set<Expand<M>>
+          : { [K in keyof T]: Expand<T[K]> }
+    : T;
```
toubatbrian, line 33: For example:
```ts
type WebSearchOptions = Expand<OpenAI.Chat.Completions.ChatCompletionCreateParams.WebSearchOptions>;
```

When hover over `WebSearchOptions` in Cursor, it shows:

```ts
type WebSearchOptions = {
    search_context_size?: "low" | "medium" | "high" | undefined;
    user_location?: {
        approximate: {
            city?: string | undefined;
            country?: string | undefined;
            region?: string | undefined;
            timezone?: string | undefined;
        };
        type: "approximate";
    } | null | undefined;
}
```

instead of

```ts
type WebSearchOptions = OpenAI.Chat.Completions.ChatCompletionCreateParams.WebSearchOptions
```

This helps debug

plugins/test/src/llm.ts
```diff
@@ -129,8 +130,8 @@ export const llm = async (llm: llmlib.LLM) => {
 
         expect(calls.length).toStrictEqual(1);
         expect(JSON.parse(calls[0]!.args).name).toStrictEqual('Theo');
```
toubatbrian, line 132: google is still outputting NULL for `.optional()` schema, will add a ticket and investigate it.

samuelcastro: I found this issue with `.optional()` and OpenAI as well https://github.com/openai/openai-node/issues/1180, looks like a known limitation from the official docs: https://platform.openai.com/docs/guides/structured-outputs#all-fields-must-be-required

davidzhao (approved):

Issue #819, opened 2025-11-11
elevenlabs autoMode and SentenceTokenizer


simllll:
### Describe the bug

for elevenlabs setting autoMode to true does not work, ~it's just not doing anything then..~ the output is actually there, but sounds like super weird and stuttering. I have seen in the python library the tokenizer must be set to a SentenceTokenizer:
https://github.com/livekit/agents/blob/758fd13ee0b86eaf42439e0f6604ccee17c5d238/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py#L150-L154

so I tried setting a SentenceTokenizer:
```
 new elevenlabs.TTS({
            autoMode: true,
            wordTokenizer: new tokenize.basic.SentenceTokenizer() as any,
});
```
and even succeeded 👍 

besides this easy fix, to allow a SentenceTokenizer + the check for autoMode + SentenceTokenizer I have found something else though: The SentenceTokenizer in typescript is simple regex implementation, in python it's using blingfire for sentence tokenization. There is a possibilty for loading this via WASM, but hard to say if this is even necessary?

### Describe your environment

linux

PR #820, opened 2025-11-11
fix elevenlabs auto mode

simllll:
## Description
fixes https://github.com/livekit/agents-js/issues/819 (without blingfire implementation)

## Changes Made
- added auto_mode flag (kept it to false by default, in python it is true by default though)
- added checks for sentencetokenizer
- added types for sentencetokenizer in the elevenlabs implementation

## Pre-Review Checklist

## Testing

- [x] manual test locally

### DIFF
--- a/.changeset/fuzzy-boxes-follow.md
+++ b/.changeset/fuzzy-boxes-follow.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-elevenlabs': patch
+---
+
+fix setting autoMode for elevenlabs


simllll:
`WARN (285528): skipping user input, current speech generation cannot be interrupted` when using the "flush" parameter.

the fix seemed to easy, but adding "flush" breaks it somehow....I guess I'm missing something that python has and we don't have in typescript. 
I'm removing the flush parameter for now again, without it it seems to work quite fine.

simllll:
Followup on "flush: true".

The reason why this actually broke thigns is that the livekit implemenation currently is not using the multi stream input api.. it starts a new websocket on each turn.

the python version uses this: https://elevenlabs.io/docs/api-reference/text-to-speech/v-1-text-to-speech-voice-id-multi-stream-input, there we also need the flush parameter. I made a quick implementation, it seems to be portable from python but needs more work.

toubatbrian:
> the python version uses this: https://elevenlabs.io/docs/api-reference/text-to-speech/v-1-text-to-speech-voice-id-multi-stream-input, there we also need the flush parameter. I made a quick implementation, it seems to be portable from python but needs more work.

Thanks for digging into this! Let me know if you need help!

PR #832, opened 2025-11-13
Support input_audio_noise_reduction for OpenAI RealtimeModel

Dimirolz:
This PR adds support for OpenAI’s `input_audio_noise_reduction` option so LiveKit agents can control the near-field/far-field noise filter when creating realtime sessions.

### DIFF
--- a/.changeset/violet-tigers-count.md
+++ b/.changeset/violet-tigers-count.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents-plugin-openai": minor
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/violet-tigers-count.md
```diff
@@ -0,0 +1,5 @@
+---
+"@livekit/agents-plugin-openai": minor
+---
```
Shubhrakanti, line 3: Is this change not backward compatible? If it is then we can do a `patch`.

toubatbrian:
https://github.com/livekit/agents-js/pull/949

Issue #836, opened 2025-11-14
mainTask hangs forever when interrupting a speech task which is on the queue


simllll:
### Describe the bug

mainTask hangs forever when interrupting a speech task which is on the queue

the mainTask in agent_activity.ts hangs forever if something like this happens:
1. create a new task with a speech out (say)
2. cancel it shortly after, but while it's not playing yet
3. the main queue calls _authorizeGeneration, even though the speechHandle is cancelled already (no check for this.interruptFut.done)
4. a new Future is created, which never resolves.

example code:

```ts
   const task = async (
      controller: AbortController
    ): Promise<void> => {
      const reply = agentSession.say("and ciao");
      console.log('reply', reply.id)
      reply.addDoneCallback((say) => {
        console.log('Done', say.id)
      })

      // Listen for abort signal
      const abortListener = () => {
        if (
          agentSession.currentAgent._agentActivity?.currentSpeech === reply
        ) {
          // currently playing
          console.log('not interrupting')
        } else {
          // interrupt queued task
          reply.interrupt();
          console.log('interrupting')
        }
      };
      controller.signal.addEventListener("abort", abortListener);

      try {
        await reply.waitForPlayout();
      } finally {
        controller.signal.removeEventListener("abort", abortListener);
      }
    }

    agentSession.say("Hello from Vienna");
    const taskHandle = Task.from(task);
    
    setTimeout(() => taskHandle.cancel(), 10);

```

this adds a new task with a speech, but cancel it right away (as long as it is not playing yet). 

But what happens:
1. the first speak out is processed
2. main task is at await _waitForGeneration
3. the next speech out task is still on the list, because a interrupt() is not removing it, it's just resolving interruptFut
4. the next loop now comes to this speak out, a new future is created, but it never finishes.

there is not even an optoin to remove a schedulede speech task.

to fix it there are several ways:
1. removing the speech task, if it's not currently playing (yet) from the speechQueue
2. just checking if the speech task is done or canceld when calling _authorizeGeneration and returning e.g. false, if it's false, _waitForGeneration is just skipped and we go on
3. ppbl more solutions ;-)

but what makes me puzzling: I couldn't find any specific handling regarding this in the python version either... is it affected too? Can't believe it that nobody stumbled on this before? What am I missing here?

### how it's working for me
I added two things:
1. in _authorizeGeneration() (speech_handle.ts) I added following code, so it's not generating a new Future, and not pushing anything to the generations.
```
if (this.interruptFut.done) {
[...]

PR #842, opened 2025-11-16
_waitForGeneration hangs when speech task is canceled before play out

simllll:
## Description
fix for #836

open for further investigation or discussion

### DIFF
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -848,14 +848,7 @@ export class AgentActivity implements RecognitionHooks {
         const speechHandle = heapItem[2];
         this._currentSpeech = speechHandle;
         speechHandle._authorizeGeneration();
-        try {
-          await Promise.race([
-            speechHandle._waitForGeneration(),
-            new Promise((_, reject) => setTimeout(() => reject('timeout after 30 seconds'), 30000)),
-          ]);
-        } catch (err) {
-          console.error('wait for generation failed', err, speechHandle);
-        }
+        await speechHandle._waitForGeneration();
         this._currentSpeech = undefined;
       }
 


agents/src/telemetry/utils.ts
```diff
@@ -1,61 +0,0 @@
-// SPDX-FileCopyrightText: 2025 LiveKit, Inc.
-//
-// SPDX-License-Identifier: Apache-2.0
-import { type Span, SpanStatusCode, context as otelContext, trace } from '@opentelemetry/api';
-import type { RealtimeModelMetrics } from '../metrics/base.js';
-import * as traceTypes from './trace_types.js';
-import { tracer } from './traces.js';
-
-export function recordException(span: Span, error: Error): void {
```
toubatbrian, line 9: Those deletions are by accident I assume? Can we revert these?

plugins/livekit/src/turn_detector/constants.ts
```diff
@@ -8,7 +8,7 @@ export const MAX_HISTORY_TURNS = 6;
 
 export const MODEL_REVISIONS: Record<EOUModelType, string> = {
   en: 'v1.2.2-en',
-  multilingual: 'v0.4.0-intl',
+  multilingual: 'v0.4.1-intl',
```
toubatbrian, line 11: Nice to catch this, but let's make it in a separate PR.

simllll: sry, there was a mixup, rebased to origin/main with the only 2 changes relevant for this fix

agents/src/telemetry/utils.ts
simllll: .

agents/src/voice/agent_activity.ts
```diff
@@ -848,7 +848,14 @@ export class AgentActivity implements RecognitionHooks {
         const speechHandle = heapItem[2];
         this._currentSpeech = speechHandle;
         speechHandle._authorizeGeneration();
-        await speechHandle._waitForGeneration();
+        try {
+          await Promise.race([
+            speechHandle._waitForGeneration(),
+            new Promise((_, reject) => setTimeout(() => reject('timeout after 30 seconds'), 30000)),
```
toubatbrian: what would happen if the actual speech is longer than 30 seconds? e.g. User asks tell me a very long story to agent

simllll: absolut good point, for our use case 30 seconds is enough, but there are ppbl cases where this is not the case. As this was mainly for testing and debugging I would say we can skip this for now. The main issue was in the cancelation of the speech before it was played out, and this is solved with the promise.race inside the speech_handle.ts

simllll: An idea for the future: allow setting a timeout, but make it configurable, and disable it by default.

PR #851, opened 2025-11-20
Upload Session Report to Cloud Observability

toubatbrian:
^ working version of session report upload

### DIFF
--- a/agents/package.json
+++ b/agents/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@livekit/agents",
-  "version": "1.0.18",
+  "version": "1.0.22",
   "description": "LiveKit Agents - Node.js",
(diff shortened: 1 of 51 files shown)


agents/package.json
```diff
@@ -61,9 +61,11 @@
     "@opentelemetry/sdk-trace-base": "^1.28.0",
     "@opentelemetry/sdk-trace-node": "^1.28.0",
     "@opentelemetry/semantic-conventions": "^1.28.0",
+    "@types/form-data": "^2.5.2",
```
simllll: This is a stub types definition. form-data provides its own type definitions, so you do not need this installed.

simllll: https://www.npmjs.com/package/@types/form-data

PR #853, opened 2025-11-20
create a new error object on timeout to have a correct stacktrace

simllll:
## Description
small change, but helps to find where a task timeout actually is happening. Otherwise we have a wrong/no real backtrace on the uncaught error object.

### DIFF
--- a/.changeset/beige-jars-sort.md
+++ b/.changeset/beige-jars-sort.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+create a new error object on timeout to have a correct stacktrace


agents/src/utils.ts
```diff
@@ -481,34 +479,30 @@ export class Task<T> {
   async cancelAndWait(timeout?: number) {
     this.cancel();
 
-    try {
```
simllll, line 484: just removed the try / catch, no code change exepct for the throw new Error('Task cancellation timed out'); in line 502

toubatbrian (approved):

PR #863, opened 2025-11-25
Upload Transcripts using HTTP Logger Exporter

toubatbrian:
Wrote a custom logger exporter that bypass OTEL library, upload transcript using application/json instead of protobuf

### DIFF
--- a/.changeset/gentle-areas-float.md
+++ b/.changeset/gentle-areas-float.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Support transcripts & traces upload to livekit cloud observability


agents/src/telemetry/traces.ts
```diff
@@ -411,45 +415,56 @@ export async function uploadSessionReport(options: {
...
+  // This fixes the issue where function_call and function_call_output with same timestamp
+  // get reordered by the dashboard
+  let lastTimestamp = 0;
   for (const item of report.chatHistory.items) {
-    const itemLog = chatItemToProto(item);
+    // Ensure monotonically increasing timestamps for proper ordering
+    // Add 0.001ms (1 microsecond) offset when timestamps collide
+    let itemTimestamp = item.createdAt;
+    if (itemTimestamp <= lastTimestamp) {
+      itemTimestamp = lastTimestamp + 0.001; // Add 1 microsecond
```
theomonnom, line 463: Timestamps cannot collide in JS?

theomonnom (approved):

agents/src/telemetry/traces.ts
toubatbrian: JS's timestamp `Date.now()` is MS precision whereas in python it is nanoseond. For some reason tool call and tool output have the same milisecond. In chat history it is fine since tool results are always after tool call. However, when sent to frontend, frontend reorder items by timestamp, and the ordering of items with the same timestamp will be undefined, sometimes tool output appears before tool call.

I can create another PR to use `process.hrtime.bigint();` by default instead of using `Date.now()` for chat item's timestamp, which is nanosecond precision

PR #866, opened 2025-11-26
Manual shutdown with reason

toubatbrian:


examples/src/manual_shutdown.ts
```diff
@@ -0,0 +1,89 @@
...
+              ])
+              .describe('The reason to end the call'),
+          }),
+          execute: async ({ reason }, { ctx }) => {
+            session.generateReply({
+              userInput: `You are about to end the call due to ${reason}, notify the user with one last message`,
+            });
+            await ctx.waitForPlayout();
+
+            session.shutdown({ reason: `custom_${reason}` });
```
davidzhao: should we just let this be any string? rather than `custom_`?

PR #873, opened 2025-12-01
Fix Pino Logger Export to OTEL Logger

toubatbrian:
Write custom adapter that wires pino log streams to OTEL cloud exporter.

### DIFF
--- a/agents/src/voice/report.ts
+++ b/agents/src/voice/report.ts
@@ -47,7 +47,6 @@ export function createSessionReport(opts: SessionReportOptions): SessionReport {
   };
 }
 
-// TODO(brian): PR5 - Add uploadSessionReport() function that creates multipart form with:
 //   - header: protobuf MetricsRecordingHeader (room_id, duration, start_time)
 //   - chat_history: JSON serialized chat history (use sessionReportToJSON)
 //   - audio: audio recording file if available (ogg format)


agents/src/log.ts
```diff
@@ -28,19 +34,50 @@ export const log = () => {
...
+    return;
+  }
+  otelEnabled = true;
+
+  const { pretty, level } = loggerOptions;
+
+  const logLevel = level || 'info';
+  const streams: { stream: DestinationStream; level: string }[] = [
+    { stream: pretty ? pinoPretty({ colorize: true }) : process.stdout, level: logLevel },
+    { stream: new OtelDestination(), level: 'debug' },
```
toubatbrian, line 79: By default we send all types of logs to OTEL, including debug logs

agents/src/telemetry/traces.ts
```diff
@@ -256,30 +254,32 @@ export async function setupCloudTracer(options: {
...
 
-    const loggerProvider = new LoggerProvider({ resource });
-
-    logs.setGlobalLoggerProvider(loggerProvider);
-
-    const logExporter = new OTLPLogExporter({
-      url: `https://${cloudHostname}/observability/logs/otlp/v0`,
-      headers,
-      compression: CompressionAlgorithm.GZIP,
+    // Initialize standalone Pino cloud exporter (no OTEL SDK dependency)
```
toubatbrian, line 259: We no longer need to use OTEL's log exporter since it never worked. Use a custom exporter instead and sending raw http requests directly.

theomonnom (approved):

PR #876, opened 2025-12-03
Implement RecorderIO for Observability

toubatbrian:
First working version of RecorderIO.

TODO item as follow ups:
- Implement `pause` and `resume` recording

### DIFF
--- a/agents/src/voice/recorder_io/recorder_io.ts
+++ b/agents/src/voice/recorder_io/recorder_io.ts
@@ -107,8 +107,6 @@ export class RecorderIO {
       await cancelAndWait([this.forwardTask!, this.encodeTask!]);
 
       this.started = false;
-
-      const stats = fs.statSync(this._outputPath!);
     } finally {
       unlock();
     }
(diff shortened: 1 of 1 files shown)


agents/src/voice/report.ts
```diff
@@ -17,6 +17,12 @@ export interface SessionReport {
   startedAt: number;
   /** Timestamp when the session report was created (milliseconds), typically at the end of the session */
   timestamp: number;
+  /** Path to the audio recording file (if recording was enabled) */
+  audioRecordingPath?: string;
+  /** Timestamp when the audio recording started (milliseconds) */
+  audioRecordingStartedAt?: number;
+  /** Duration of the session in milliseconds */
+  duration?: number;
```
theomonnom, line 25: I'm wondering if we should use the same units as Python for the SessionReport?

This way you can export to json and it's compatible across SDKs

toubatbrian: Good idea, I think I do manually convert all units to seconds for session report, but I'll double check

PR #884, opened 2025-12-04
fix: silence stream cleanup errors during teardown

Hormold:
## Description

Gracefully handle common edge cases when detaching audio streams:
- Return early from detachSource() if source was never set
- Catch and ignore 'Controller is already closed' errors
- Catch and ignore 'WritableStream is closed' errors

These errors were spamming logs during normal cleanup operations when sessions closed before audio was fully initialized.
image

## Changes Made

- Extended `isStreamReaderReleaseError()` to handle controller/writer closed errors
- Made `detachSource()` return early if source was never set
- Added defensive checks in `detachAudioInput()` and session cleanup

## Testing

- [x] Build passes locally
- [x] No behavioral changes - only suppresses expected errors during cleanup

### DIFF
--- a/agents/src/stream/deferred_stream.ts
+++ b/agents/src/stream/deferred_stream.ts
@@ -9,7 +9,10 @@ import type {
 import { IdentityTransform } from './identity_transform.js';
 
 /**
- * Check if error is related to stream cleanup operations
+ * Check if error is related to stream cleanup operations.
+ *
+ * These errors are expected when calling reader.read() after releaseLock()
+ * or when writing to already closed streams during cleanup:
  *
  * Invalid state: Releasing reader
  * Invalid state: The reader is not attached to a stream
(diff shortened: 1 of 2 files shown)


agents/src/stream/deferred_stream.ts
```diff
@@ -9,15 +9,19 @@ import type {
 import { IdentityTransform } from './identity_transform.js';
 
 /**
- * Check if error is related to reader.read after release lock
+ * Check if error is related to stream cleanup operations
```
toubatbrian: can we include both comments? Just want ppl to be aware that calling `reader.read` after release lock is the root cause of triggering tis error

agents/src/voice/agent_activity.ts
```diff
@@ -417,6 +417,10 @@ export class AgentActivity implements RecognitionHooks {
   }
 
   detachAudioInput(): void {
+    if (!this.audioStream.isSourceSet) {
```
toubatbrian: Do we need this if check given that `detachSource` already handle this?

Hormold: good point, done

PR #891, opened 2025-12-08
Improve AgentActivity resource cleanup

toubatbrian:
## Problem
Previously we have https://github.com/livekit/agents-js/pull/849 which adds `close()` call to STT / TTS / VAD on agent activity close and subsequently https://github.com/livekit/agents-js/pull/883. Turned out these are not correct way of handling resource cleanup, because there could be another AgentActivity starts after the closed activity, i.e. agent transfer. However, the close() method on STT / TTS are "global" w.r.t the STT / TTS class, meaning aborting the class object not only cancels any active stt / tts stream, but also make STT / TTS unusable for any future streams created under this STT / TTS object, making the speeches after agent transfer failed.

## Fix
- Refactor default `sttNode` `llmNode` and `ttsNode` to always call `close()` on respective stream objects created from global `STT` `TTS` `LLM` component, allowing plugins to implement proper resource cleanup inside `SpeechStream` `SynthsizeStream` and `LLMStream` instead of on the root class.
- Some minor bug fixes to improve stability

## Future TODOs
- Ensure proper abort handling for external request in TTS / LLM streams across all plugins

### DIFF
--- a/agents/src/stt/stt.ts
+++ b/agents/src/stt/stt.ts
@@ -296,7 +296,7 @@ export abstract class SpeechStream implements AsyncIterableIterator<SpeechEvent>
 
   protected abstract run(): Promise<void>;
 
-  get abortSignal(): AbortSignal {
+  protected get abortSignal(): AbortSignal {
     return this.abortController.signal;
   }
 
--- a/plugins/deepgram/src/stt.ts
+++ b/plugins/deepgram/src/stt.ts
@@ -260,6 +260,8 @@ export class SpeechStream extends stt.SpeechStream {
         samples100Ms,
       );
 
+      // waitForAbort internally sets up an abort listener on the abort signal
+      // we need to put it outside loop to avoid constant re-registration of the listener
       const abortPromise = waitForAbort(this.abortSignal);
 
       try {


agents/src/stt/stream_adapter.ts
```diff
@@ -23,10 +23,14 @@ export class StreamAdapter extends STT {
...
+
+    this.#stt.on('error', (error) => {
+      this.emit('error', error);
+    });
   }
 
-  _recognize(frame: AudioFrame): Promise<SpeechEvent> {
-    return this.#stt.recognize(frame);
+  _recognize(frame: AudioFrame, abortSignal?: AbortSignal): Promise<SpeechEvent> {
+    return this.#stt.recognize(frame, abortSignal);
```
toubatbrian, line 33: plugins can now attach this abortSignal for any external request cancellation

agents/src/stt/stt.ts
```diff
@@ -173,6 +177,8 @@ export abstract class SpeechStream implements AsyncIterableIterator<SpeechEvent>
   private logger = log();
   private _connOptions: APIConnectOptions;
 
+  protected abortController = new AbortController();
```
toubatbrian, line 180: add an `abortController` in root STT class, which mainly handles any external request cancellation such as `fetch`. `abortSignal` will be `aborted` whenever the underling STT stream is closed.

agents/src/voice/agent.ts
```diff
@@ -260,28 +260,40 @@ export class Agent<UserData = any> {
...
 
       if (!wrapped_stt.capabilities.streaming) {
-        if (!agent.vad) {
+        if (!activity.vad) {
           throw new Error(
             'STT does not support streaming, add a VAD to the AgentTask/VoiceAgent to enable streaming',
           );
         }
-        wrapped_stt = new STTStreamAdapter(wrapped_stt, agent.vad);
+        wrapped_stt = new STTStreamAdapter(wrapped_stt, activity.vad);
```
toubatbrian: Fix a minor bug on non-streaming STT, allowing `vad` on AgentSession to forward to `Agent` if not set explicitly

agents/src/voice/agent.ts
```diff
@@ -314,15 +327,27 @@ export class Agent<UserData = any> {
         connOptions,
         parallelToolCalls: true,
       });
+
+      let cleaned = false;
+      const cleanup = () => {
+        if (cleaned) return;
+        cleaned = true;
+        stream.close();
```
toubatbrian, line 335: ^ `stream.close()` in `STT / TTS / LLM` stream should be the method that plugin uses to do proper resource cancellation.  @simllll @KrishnaShuk

plugins/deepgram/src/stt.ts
```diff
@@ -263,12 +260,11 @@ export class SpeechStream extends stt.SpeechStream {
         samples100Ms,
       );
 
+      const abortPromise = waitForAbort(this.abortSignal);
```
toubatbrian, line 265: minor bug fix: prevent overflowing max #listeners

agents/src/stt/stt.ts
```diff
@@ -290,6 +296,10 @@ export abstract class SpeechStream implements AsyncIterableIterator<SpeechEvent>
 
   protected abstract run(): Promise<void>;
 
+  get abortSignal(): AbortSignal {
```
theomonnom: Should this be exposed publicly?

plugins/deepgram/src/stt.ts
theomonnom, line 265: Can you add a comment? it doesn't seems obvious

agents/src/stt/stt.ts
toubatbrian: I'll make it protected

PR #900, opened 2025-12-09
Sync all package versions with `fixed` 

toubatbrian:
## Summary

Configure changesets to keep all package versions aligned using the `fixed` option.

## Problem

Currently, plugin packages have different versions from `@livekit/agents`:
- `@livekit/agents`: 1.0.25
- `@livekit/agents-plugin-anam`: 1.0.15
- `@livekit/agents-plugin-bey`: 1.0.15
- `@livekit/agents-plugin-inworld`: 1.0.2
- `@livekit/agents-plugin-rime`: 1.0.18

This makes it confusing for users to know which versions are compatible.

## Solution

Added `fixed` configuration to `.changeset/config.json` which ensures all packages in the group **always have the same version**.

When `changeset version` runs, it will automatically bump all packages to the same version.

## Changes

- Added `fixed` array containing all `@livekit/agents*` packages
- Next release will align all packages to the same version

### DIFF
--- a/.changeset/curvy-geckos-search.md
+++ b/.changeset/curvy-geckos-search.md
@@ -1,0 +1,19 @@
+---
+'@livekit/agents': patch
+'@livekit/agents-plugin-anam': patch
+'@livekit/agents-plugin-bey': patch
+'@livekit/agents-plugin-cartesia': patch
+'@livekit/agents-plugin-deepgram': patch
+'@livekit/agents-plugin-elevenlabs': patch
+'@livekit/agents-plugin-google': patch
+'@livekit/agents-plugin-inworld': patch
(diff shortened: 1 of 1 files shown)


.changeset/config.json
```diff
@@ -8,7 +8,25 @@
   ],
   "commit": false,
   "ignore": ["livekit-agents-examples"],
-  "linked": [],
+  "fixed": [
+    [
+      "@livekit/agents",
+      "@livekit/agents-plugin-anam",
```
lukasIO: nit: you could use a glob expr to match all plugins, if that's the intention. 
that way you don't have to remember to add it here when adding a new plugin

PR #910, opened 2025-12-12
Add ConnectionPool for Inference TTS + Fix interruption handling to prevent transcript “spill”

toubatbrian:
- Fix voice interruption behavior by clearing audio buffer immediately on interrupt and only treating playback as “completed” when playbackPosition is near the end, preventing full-text transcript spill after stop.

- Add ConnectionPool to reduce reconnecting TTS websockets for consecutive TTS requests.

Changes has been tested on all TTS including agent handoffs scenarios and interruption handling

### DIFF
--- a/agents/src/voice/transcription/synchronizer.ts
+++ b/agents/src/voice/transcription/synchronizer.ts
@@ -151,14 +151,7 @@ class SegmentSynchronizerImpl {
       return;
     }
 
-    const playbackPosition = _playbackPosition;
-    const epsilonSeconds = 0.05;
-    const nearEnd = playbackPosition >= Math.max(0, this.audioData.pushedDuration - epsilonSeconds);
-
-    // Only mark as fully completed if playback reached (roughly) the end.
-    // This prevents returning the full transcript in cases where the sink reports interrupted=false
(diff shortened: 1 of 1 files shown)


agents/src/voice/transcription/synchronizer.ts
```diff
@@ -151,7 +151,14 @@ class SegmentSynchronizerImpl {
...
     }
 
-    if (!interrupted) {
+    const playbackPosition = _playbackPosition;
+    const epsilonSeconds = 0.05;
+    const nearEnd = playbackPosition >= Math.max(0, this.audioData.pushedDuration - epsilonSeconds);
+
+    // Only mark as fully completed if playback reached (roughly) the end.
+    // This prevents returning the full transcript in cases where the sink reports interrupted=false
+    // but playbackPosition indicates partial playout.
```
davidzhao: what situations would this happen? if it's not interrupted, shouldn't it play out fully?

toubatbrian: I added this when trying to solve the spill bug, but looks like this is not the root cause. I will revert this back.

PR #922, opened 2025-12-18
Fix baseten dependencies

toubatbrian:
The Release pipeline is currently failing in JS due to a dependency issue.

The root cause is that the baseten plugin depends on @livekit/agents-plugin-openai. Version Packages is automatically bumping every occurrence of all plugins to the next version (1.0.29), including nested dependencies in package.json.

However, since the next version of @livekit/agents-plugin-openai hasn't been released yet, the build fails because it can’t find version 1.0.29 under baseten's dependencies.

Fix: removing the plugin-as-a-dependency pattern in the baseten plugin.

### DIFF
--- a/.changeset/ninety-chicken-pull.md
+++ b/.changeset/ninety-chicken-pull.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-baseten': patch
+---
+
+Fix baseten dependency issue by removing the plugin-as-a-dependency


plugins/baseten/src/llm.ts
```diff
@@ -6,9 +6,136 @@
...
+  parallelToolCalls?: boolean;
+  metadata?: Record<string, string>;
+  maxCompletionTokens?: number;
+  serviceTier?: string;
+  store?: boolean;
+  strictToolSchema?: boolean;
+}
+
+const defaultLLMOptions: LLMOptions = {
+  model: 'gpt-4.1',
```
davidzhao: is this the right default? for baseten  I don't think we should have one

plugins/baseten/src/llm.ts
```diff
@@ -6,9 +6,136 @@
...
+  apiKey: process.env.OPENAI_API_KEY,
+  parallelToolCalls: true,
+  strictToolSchema: false,
+};
+
+export class OpenAILLM extends llm.LLM {
```
davidzhao, line 37: can we call this `OpenAICompatLLM` maybe? and any way to avoid this boilerplate copypaste?

toubatbrian: The original `baseten` directly import the openaiLLM class from openai plugin, which was supposed to be the good pattern, but since that somehow fails the CI due to this dependency-as-dependency issue, I want to temporarily make a copy of the openai competible LLM to fix the release. I will later on figure out a better approach to expose a generic OpenAI Compatible LLM class for other plugin packages.

PR #941, opened 2026-01-05
Fix flaky IPC test EPIPE error

toubatbrian:
Add error handler to suppress EPIPE race condition in `supervised_proc.test.ts`

### DIFF
--- a/.changeset/eight-files-destroy.md
+++ b/.changeset/eight-files-destroy.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Fix flaky IPC test EPIPE error


theomonnom (approved):

agents/src/ipc/supervised_proc.test.ts
```diff
@@ -96,6 +96,12 @@ describe('IPC send on dead process', () => {
     const child = fork(childScript, [], { stdio: ['pipe', 'pipe', 'pipe', 'ipc'] });
     const exitPromise = new Promise<void>((r) => child.on('exit', r));
 
+    // Suppress EPIPE errors that can occur due to race conditions between
+    // child.connected check and the actual pipe state
+    child.on('error', (err: NodeJS.ErrnoException) => {
+      if (err.code !== 'EPIPE') throw err;
+    });
```
Shubhrakanti, line 103: what's the root case? Just curious. 

Isn't it a real issue if we're writing to a closed stream?

toubatbrian: This is due to race condition. `child.connected` can still be true briefly after `kill()` while the pipe is already closed. The 20ms interval fires in that window results in EPIPE.

This is a test artifact, not a real app issue.

PR #949, opened 2026-01-07
Update OpenAI realtime GA API

toubatbrian:


.changeset/afraid-beds-bet.md
```diff
@@ -0,0 +1,5 @@
+---
+'@livekit/agents-plugin-openai': patch
```
lukasIO, line 2: given that the now legacy `AudioFormat` was exposed to users and we're renaming it, I think this should be a minor version upgrade.

toubatbrian: I can make it backward competible tho. One thing makes it hard to do minor bump is that we recently sync all plugin packages, along with main agent package, to the same version on every release. Making a minor bump on one plugin would cause all packages to be minor bumped. Any idea on how we should properly handle this?

toubatbrian: Actually, while `AudioFormat` type changed from a string ('pcm16') to an interface `({ type: 'audio/pcm', rate: number })`, users never interact with this type directly - it's only used internally for the WebSocket protocol.

The public API (constructor options) remains fully backward compatible:
Beta:
```ts
new RealtimeModel({  model: 'gpt-4o-realtime',  voice: 'alloy',  temperature: 0.8,  inputAudioTranscription: { model: 'whisper-1' },  turnDetection: { type: 'server_vad' },  // ... other options});
```

GA (same code still works):
```ts
new RealtimeModel({  model: 'gpt-4o-realtime',  voice: 'alloy',  temperature: 0.8,  // kept for compat, marked @deprecated  inputAudioTranscription: { model: 'whisper-1' },  turnDetection: { type: 'server_vad' },  // ... other options  // NEW optional: inputAudioNoiseReduction, tracing});
```

All existing constructor options are preserved, temperature is kept (just deprecated), and new options (inputAudioNoiseReduction, tracing) are additive. Keeping it as a patch should be fine since no user-facing breaking changes.

Let me know how you think @lukasIO

toubatbrian: I just verified, both 

```ts
llm: new openai.realtime.beta.RealtimeModel(),
```

and 

```ts
llm: new openai.realtime.RealtimeModel(),
```

are working properly.

lukasIO: > Actually, while AudioFormat type changed from a string ('pcm16') to an interface ({ type: 'audio/pcm', rate: number }), users never interact with this type directly - it's only used internally for the WebSocket protocol.

if that's the case then we should think about not exporting it from `index.ts` in the first place going forward. Right now it's available for users to interact with as a type even if they don't directly use it.

>  One thing makes it hard to do minor bump is that we recently sync all plugin packages, along with main agent package, to the same version on every release. Making a minor bump on one plugin would cause all packages to be minor bumped.

yeah, this situation is less than ideal imo. Let's discuss this in our sync next week.

PR #958, opened 2026-01-13
add OpenAI Responses plugin

tinalenguyen:
## Description
implements openai responses API plugin

to use:
`llm: new openai.responses.LLM(),`

### DIFF
--- a/plugins/openai/src/responses/llm.test.ts
+++ b/plugins/openai/src/responses/llm.test.ts
@@ -9,8 +9,9 @@ describe('OpenAI Responses', async () => {
   await llm(
     new LLM({
       temperature: 0,
+      strictToolSchema: false,
     }),
-    false,
+    true,
   );
 });
 
--- a/plugins/test/src/llm.ts
+++ b/plugins/test/src/llm.ts
@@ -112,7 +112,7 @@ const toolCtxStrict: llmlib.ToolContext = {
   }),
 };
 
-export const llm = async (llm: llmlib.LLM, isGoogle: boolean) => {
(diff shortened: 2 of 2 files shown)


agents/src/llm/provider_format/openai.ts
```diff
@@ -144,3 +144,106 @@ async function toImageContent(content: ImageContent) {
...
+  }
+
+  return {
+    type: 'input_image' as const,
+    image_url: `data:${serialized.mimeType};base64,${serialized.base64Data}`,
+    detail: serialized.inferenceDetail,
+  };
+}
+
+export async function toResponsesChatCtx(
```
toubatbrian, line 177: Could we generate some unit tests for this?

plugins/openai/src/responses/llm.ts
```diff
@@ -0,0 +1,327 @@
...
+  strictToolSchema?: boolean;
+}
+
+const defaultLLMOptions: LLMOptions = {
+  model: 'gpt-4.1',
+  apiKey: process.env.OPENAI_API_KEY,
+  strictToolSchema: true,
+};
+
+export class LLM extends llm.LLM {
```
toubatbrian, line 35: Could we also add unit tests for this new response API in `llm.test.ts`? Like this
```ts
describe('OpenAI', async () => {
  await llm(
    new LLM({
      temperature: 0,
    }),
    false,
  );
});

describe('OpenAI strict tool schema', async () => {
  await llmStrict(
    new LLM({
      temperature: 0,
      strictToolSchema: true,
    }),
  );
});
```

toubatbrian (approved):
LGTM!

PR #965, opened 2026-01-13
feat(agents): add initial test framework for agent testing

toubatbrian:
## Summary
Add initial test framework support for testing agent behavior without requiring a real LiveKit room connection.

## Changes
- Add `RunResult` class for collecting agent events during test runs
- Add `RunAssert` API for asserting on events (`nextEvent()`, `isMessage()`, `noMoreEvents()`, `at()`, `skipNext()`)
- Add `session.run({ userInput })` method for triggering agent responses in tests
- Make `room` parameter optional in `session.start()` to support roomless testing
- Add `isSpeechHandle` type guard for runtime type checking
- Fix memory leak: only tee LLM stream when TTS is needed
- Add `speechHandle._itemAdded()` calls to collect assistant messages in test results

## Usage
```ts
const session = new AgentSession({ llm });
await session.start({ agent });

const result = session.run({ userInput: 'Hello!' });
await result.wait();

result.expect.nextEvent().isMessage({ role: 'assistant' });
result.expect.noMoreEvents();

await session.close();
```

## Future Work
- Range access (`range()`) with `containsFunctionCall()`, `containsMessage()`
- `judge()` method for LLM-based evaluation
- `mockTools()` utility
- Typed finalOutput support

### DIFF
--- a/.changeset/strong-moose-fly.md
+++ b/.changeset/strong-moose-fly.md
@@ -1,6 +1,0 @@
----
-'@livekit/agents-plugin-openai': patch
-'@livekit/agents': patch
----
-
-fix openai-realtime by removing buggy resolveGeneration method that used wrong key lookup


agents/src/voice/speech_handle.ts
```diff
@@ -91,7 +109,7 @@ export class SpeechHandle {
     this._allowInterruptions = value;
   }
 
-  done(): boolean {
+  get done(): boolean {
```
lukasIO: this appears to be a breaking change?

toubatbrian: uh good catch! I will revert this

examples/src/testing/run_result.test.ts
```diff
@@ -0,0 +1,218 @@
...
+          execute: async () => {
+            return '3:00 PM';
+          },
+        }),
+      },
+    });
+  }
+}
+
+describe('RunResult', { timeout: 120_000 }, () => {
```
toubatbrian, line 59: example usage of test framework

agents/src/voice/agent_activity.ts
```diff
@@ -1350,18 +1350,24 @@ export class AgentActivity implements RecognitionHooks {
...
     tasks.push(llmTask);
 
-    const [ttsTextInput, llmOutput] = llmGenData.textStream.tee();
-
     let ttsTask: Task<void> | null = null;
     let ttsStream: ReadableStream<AudioFrame> | null = null;
+    let llmOutput: ReadableStream<string>;
+
     if (audioOutput) {
+      // Only tee the stream when we need TTS
```
lukasIO, line 1358: was this causing any issues or is this an optimisation?

agents/src/voice/agent_session.ts
```diff
@@ -283,41 +287,45 @@ export class AgentSession<
...
+        outputOptions,
+      });
+      this.roomIO.start();
+    }
 
     let ctx: JobContext | undefined = undefined;
     try {
       ctx = getJobContext();
-    } catch (error) {
+    } catch {
```
lukasIO, line 323: can we detect whether we're within an eval and still log the warning if not?

agents/src/voice/agent_activity.ts
toubatbrian: Yes this is causing issue when running the eval

agents/src/voice/agent_session.ts
toubatbrian: I guess we could, by using a node context variable. The current code is directly ported from python so maybe let's keep the current version and we can figure out a better way with a follow up PR.

PR #980, opened 2026-01-16
Add AdaptiveInterruptionDetector

lukasIO:
## Summary by CodeRabbit

* **New Features**
  - Adaptive interruption detection with configurable thresholds, timing parameters, and probability estimation.
  - HTTP and WebSocket transport options for real-time interruption inference.
  - Stream-based interruption detection and management APIs.
  - Telemetry attributes to track interruption events and metrics.

* **Chores**
  - Added ofetch dependency.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/agents/src/inference/interruption/InterruptionStream.ts
+++ b/agents/src/inference/interruption/InterruptionStream.ts
@@ -299,8 +299,23 @@ export class InterruptionStreamBase {
       transport = createHttpTransport(transportOptions, getState, setState, handleSpanUpdate);
     }
 
+    const eventEmitter = new TransformStream<InterruptionEvent, InterruptionEvent>({
+      transform: (chunk, controller) => {
+        if (chunk.type === InterruptionEventType.INTERRUPTION) {
+          this.model.emit('userInterruptionDetected', chunk);
+        } else if (chunk.type === InterruptionEventType.OVERLAP_SPEECH_ENDED) {
+          this.model.emit('overlapSpeechEnded', chunk);
+        }
+        controller.enqueue(chunk);
+      },
+    });
+
     // Pipeline: input -> audioTransformer -> transport -> eventStream
-    return this.inputStream.stream().pipeThrough(audioTransformer).pipeThrough(transport);
+    return this.inputStream
+      .stream()
+      .pipeThrough(audioTransformer)
+      .pipeThrough(transport)
+      .pipeThrough(eventEmitter);
   }
 
   private ensureInputNotEnded() {


agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,392 @@
...
+        }
+      },
+
+      async flush() {
+        await close();
+      },
+    },
+    { highWaterMark: 2 },
+    { highWaterMark: 2 },
+  );
```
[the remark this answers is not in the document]
lukasIO: That's not applicable as audio is only written to the transport when overlapSpeechStarted is true.
The ws_transport's overlapSpeechStartedAt check is a secondary guard, not the primary control.
The flow is: InterruptionStream gates audio → sends to transport → transport has additional guard

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+    headers: { Authorization: `Bearer ${token}` },
+  });
+
+  const { readable, writable } = webSocketToStream(ws);
+
+  await new Promise<void>((resolve, reject) => {
+    const timeout = setTimeout(() => {
+      ws.terminate();
+      reject(new Error('WebSocket connection timeout'));
+    }, options.timeout);
```
toubatbrian, line 81: Should we call clearTimeout also inside ws.once('end')?

lukasIO: can `end` happen before `open` ?

toubatbrian: Make sense. It's not necessary then

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+  writable: WritableStream<Uint8Array>;
+  ws: WebSocket;
+}> {
+  const baseUrl = options.baseUrl.replace(/^http/, 'ws');
+  const url = `${baseUrl}/bargein`;
+  const token = await createAccessToken(options.apiKey, options.apiSecret);
+
+  const ws = new WebSocket(url, {
+    headers: { Authorization: `Bearer ${token}` },
+  });
```
toubatbrian, line 73: We might want to expose a options object in `connectWebsocket` that can pass into the `new Websocket` option if we would like to make it a general utility method in the future. For now, this is good, maybe add a small comment to make it better for us to track this TODO.

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+  const duplex = createWebSocketStream(ws);
+  duplex.on('error', (err) => log().error({ err }, 'WebSocket stream error'));
+
+  // End the write side when the read side ends
+  duplex.on('end', () => duplex.end());
+
+  const writable = Writable.toWeb(duplex) as WritableStream<Uint8Array>;
+  const readable = Readable.toWeb(duplex) as ReadableStream<Uint8Array>;
+
+  return { readable, writable };
```
toubatbrian, line 56: maybe unrelated: in JS I always need to manually close/end the resource whenever using stream API. I'm not sure in this case if we need to manually close it or like "detach" ws from stream API to prevent memory leak

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+            num_channels: 1,
+            threshold: options.threshold,
+            min_frames: options.minFrames,
+            encoding: 's16le',
+          },
+        });
+        await writer.write(new TextEncoder().encode(sessionCreateMsg));
+
+        // Start reading responses
+        readerTask = processResponses(conn.readable);
```
toubatbrian, line 146: Should we wrap this with `Task.from` and control cancellation logic inside `processResponse`? Otherwise it could become a dangling promise.

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+  async function processResponses(readable: ReadableStream<Uint8Array>): Promise<void> {
+    const reader = readable.getReader();
+    const decoder = new TextDecoder();
+    let buffer = '';
+
+    try {
+      while (true) {
+        const { done, value } = await reader.read();
```
toubatbrian, line 171: Same as ^, we should use abortController explicitly and do `await Promise.race([reader.read(), waitForAbort(signal)]);` to unblock dangling await

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+        buffer = lines.pop() ?? '';
+
+        for (const line of lines) {
+          if (line.trim()) {
+            try {
+              const message: WsMessage = JSON.parse(line);
+              handleMessage(message);
+            } catch {
+              logger.warn({ line }, 'Failed to parse WebSocket message');
+            }
```
toubatbrian, line 187: Could be safer and easier to debug if we write a zod schema for `WsMessage` and to a `safeParseAsync` here.

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+          }
+        }
+
+        // Also try parsing buffer as complete message (for non-newline-delimited)
+        if (buffer.trim()) {
+          try {
+            const message: WsMessage = JSON.parse(buffer);
```
toubatbrian, line 194: ^

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+      case MSG_INTERRUPTION_DETECTED: {
+        const createdAt = message.created_at ?? 0;
+        if (state.overlapSpeechStarted && state.overlapSpeechStartedAt !== undefined) {
+          const existing = state.cache.get(createdAt);
+          const entry = new InterruptionCacheEntry({
+            createdAt,
+            speechInput: existing?.speechInput,
+            totalDurationInS: (performance.now() - createdAt) / 1000,
+            probabilities: message.probabilities,
+            isInterruption: true,
```
toubatbrian, line 224: Same ^, usually I found explicitly defining ClientMessage and ServerMessage zod schema, and manually parsed the raw string to JSON makes it easier to debug & iterate on Ws-based STT/TTS API. This way we don't need to do additional assertions on optional fields.

toubatbrian: e.g. https://github.com/livekit/agents-js/pull/990/changes#diff-f8bf8580746f49acb14c7f9e0e8e11105c2f20b2642fa372aa8286e250e444c2

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+    const audioBytes = new Uint8Array(
+      audioSlice.buffer,
+      audioSlice.byteOffset,
+      audioSlice.byteLength,
+    );
+    const combined = new Uint8Array(8 + audioBytes.length);
+    combined.set(new Uint8Array(header), 0);
+    combined.set(audioBytes, 8);
+
+    await writer.write(combined);
```
toubatbrian, line 330: We probably need to wrap this around try catch to skip any error caused by Writer been written after closed (this might happen due to race condition)

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+  async function close(): Promise<void> {
+    if (writer && ws?.readyState === WebSocket.OPEN) {
+      const closeMsg = JSON.stringify({ type: MSG_SESSION_CLOSE });
+      await writer.write(new TextEncoder().encode(closeMsg));
+      writer.releaseLock();
+      writer = null;
+    }
+    ws?.close(1000);
```
toubatbrian, line 340: What is 1000? Maybe add a comment? Also just curious does closing ws automatically cleanup resources of writable / readable stream attached to it?

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -0,0 +1,387 @@
...
+          logger.error({ err }, 'Failed to send audio data over WebSocket');
+        }
+      },
+
+      async flush() {
+        await close();
+      },
+    },
+    { highWaterMark: 2 },
+    { highWaterMark: 2 },
```
toubatbrian, line 383: Any reason we set high water mark to be 2?

I saw in `IdentityTransform` we make it to be max integer:
```ts
export class IdentityTransform<T> extends TransformStream<T, T> {
  constructor() {
    super(
      {
        transform: (chunk, controller) => controller.enqueue(chunk),
      },
      // By default the transfor stream will buffer only one chunk at a time.
      // In order to follow the python agents channel.py, we set set the capaciy to be effectively infinite.
      { highWaterMark: Number.MAX_SAFE_INTEGER },
      { highWaterMark: Number.MAX_SAFE_INTEGER },
    );
  }
}
```

agents/src/inference/interruption/utils.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  constructor(maxLen: number = 10) {
+    this.maxLen = maxLen;
+  }
+
+  set(key: K, value: V): void {
+    this.cache.set(key, value);
+    if (this.cache.size > this.maxLen) {
+      // Remove the oldest entry (first inserted)
+      const firstKey = this.cache.keys().next().value as K;
```
toubatbrian, line 22: I'm not sure if this is correct. In python I saw it uses `OrderedDict` which preserves insertion order of map items, but I assume JS `Map` is unordered?

agents/src/inference/interruption/ws_transport.ts
lukasIO: yeah, would be good to reuse this if we ever need another websocket transport, leaving it as is for now

agents/src/inference/interruption/ws_transport.ts
lukasIO: this makes sense and it's been significantly refactored already in #991 where the whole `ws` instance is being returned to perform things like manual closing

agents/src/inference/interruption/ws_transport.ts
lukasIO: there's a test case already that ensures closing the websocket closes the streams also. Do you think we need an additional abort signal beyond the ability to close the websocket?

agents/src/inference/interruption/ws_transport.ts
lukasIO: ideally this would be a protobuf as part of the agents protocol, we should keep this in mind for a v2 of the endpoint

agents/src/inference/interruption/ws_transport.ts
lukasIO: answered this above, `1000` is the clean close status for websockets. I'll add a comment

agents/src/inference/interruption/ws_transport.ts
lukasIO: setting it to `2` instead of unbounded ensures we apply backpressure and don't try to overload the barge in server with endless parallel requests in case it doesn't keep up

agents/src/inference/interruption/utils.ts
lukasIO: from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

> The keys in Map are ordered in a straightforward way: A Map object iterates entries, keys, and values in the order of entry insertion.

PR #1000, opened 2026-01-27
fix: preserve thought_signature across parallel tool calls for Gemini 3+

toubatbrian:
Fixes a 400 error when using Gemini 3+ models with parallel tool calls. The `thought_signature` is now correctly propagated to all tool calls in a batch instead of only the first one. Also removes a hardcoded `parallelToolCalls: true` in the voice pipeline that was overriding user configuration.

## Release Notes

* **Bug Fixes**
  * Fixed preservation of context signatures across parallel tool calls for Gemini 3+ models to improve inference quality with the inference gateway.

* **Chores**
  * Updated configuration to use model options for controlling parallel tool call behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.changeset/salty-regions-relate.md
+++ b/.changeset/salty-regions-relate.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+preserve thought_signature across parallel tool calls for Gemini 3+ for inference gateway


agents/src/voice/agent.ts
```diff
@@ -325,16 +325,16 @@ export class Agent<UserData = any> {
...
       const connOptions = activity.agentSession.connOptions.llmConnOptions;
 
+      // parallelToolCalls is not passed here - it will use the value from LLM's modelOptions
+      // This allows users to configure it via: new inference.LLM({ modelOptions: { parallel_tool_calls: false } })
       const stream = activity.llm.chat({
         chatCtx,
         toolCtx,
         toolChoice,
         connOptions,
-        parallelToolCalls: true,
```
toubatbrian, line 337: Note: although this was hardcoded to true, the actual value was still correctly been overriden by user config inside LLM in previous version.

PR #1018, opened 2026-02-03
fix(stream): allow DeferredReadableStream to be reused after detachSource

mshivam019:
## Description
Fix `DeferredReadableStream` to allow `setSource()` to be called again after `detachSource()`, enabling stream reuse during participant reconnection scenarios.

When a participant's connection is interrupted and restored (e.g., browser reload, network interruption, or user canceling the "Leave site?" confirmation dialog), the SDK's `ParticipantAudioInputStream` attempts to reattach the audio track via `onTrackSubscribed`. This triggers:

1. `closeStream()` → calls `deferredStream.detachSource()`
2. Then `deferredStream.setSource(newAudioStream)` to attach the new track

However, `setSource()` throws **"Stream source already set"** because:

1. `detachSource()` releases the reader lock but doesn't reset `sourceReader` to `undefined`
2. `isSourceSet` (which checks `!!this.sourceReader`) still returns `true`
3. Even if that were fixed, the `writer` may already be closed from the previous session

This makes it impossible to reuse the agent session after any participant reconnection event.

## Changes Made
- Reset `sourceReader = undefined` in `detachSource()` so `isSourceSet` returns `false`
- Added `writerClosed` state tracking to detect when transform needs recreation
- Recreate the internal transform stream in `setSource()` if writer was previously closed
- Added check for `!this.sourceReader` before read in `pump()` to handle race condition where `detachSource()` is called while pump is running

## Pre-Review Checklist
- [ ] **Video demo**: N/A - This is an internal stream management fix, not user-facing UI. The fix enables reconnection scenarios that previously crashed.

## Testing
- [x] Automated tests added/updated (if applicable) - All 25 existing tests in `deferred_stream.test.ts` pass
- [x] All tests pass - `pnpm build:agents` and `pnpm test` pass
- [ ] Make sure both `restaurant_agent.ts` and `realtime_agent.ts` work properly - N/A for this internal stream fix

**Manual Testing Performed:**
- Verified fix in production with browser tab close/cancel reconnection scenario
- Verified fix with network interruption simulation
- Confirmed agent session remains functional after participant reconnects

## Additional Notes
This bug affects any scenario where a participant disconnects and reconnects to the same room session. Without this fix, the agent becomes unresponsive after reconnection because audio input cannot be reattached.

---

### DIFF
--- a/agents/src/stream/deferred_stream.ts
+++ b/agents/src/stream/deferred_stream.ts
@@ -37,6 +37,7 @@ export class DeferredReadableStream<T> {
   private transform: IdentityTransform<T>;
   private writer: WritableStreamDefaultWriter<T>;
   private sourceReader?: ReadableStreamDefaultReader<T>;
+  private writerClosed: boolean = false;
 
   constructor() {
     this.transform = new IdentityTransform<T>();
@@ -53,10 +54,18 @@ export class DeferredReadableStream<T> {
 
   /**
    * Call once the actual source is ready.
+   * Can be called again after detachSource() to attach a new source.
    */
   setSource(source: ReadableStream<T>) {
     if (this.isSourceSet) {
       throw new Error('Stream source already set');
+    }
+
+    // If writer was closed from a previous source, recreate the transform
+    if (this.writerClosed) {
+      this.transform = new IdentityTransform<T>();
+      this.writer = this.transform.writable.getWriter();
+      this.writerClosed = false;
     }
 
     this.sourceReader = source.getReader();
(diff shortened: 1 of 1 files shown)


mshivam019:
Hey @toubatbrian, thanks for the pointer! In the meantime, I migrated my architecture to create a fresh room on reconnection instead — the backend preloads the conversation context into the new agent session, so the user picks up right where they left off without needing stream reuse. That eliminated the need for this PR on my end.

That said, keeping the original room open would still be the better approach — with the new room architecture, there's a window where recent conversation data that's still in the agent's context but hasn't been persisted by the backend yet can get lost during the handoff. The stream reuse approach avoids that entirely since the session stays alive.

I checked out #1036 and the deferred stream approach looks solid. I'll give it a try soon and report back if I run into anything.

PR #1022, opened 2026-02-04
feat(tts): add FallbackAdapter for TTS failover support

gokuljs:
This adds a FallbackAdapter for TTS that lets you configure multiple TTS providers and automatically switches to the next one if the current one fails. It handles both connection errors and silent failures where the TTS connects but doesn't return any audio. Failed providers are automatically tested in the background and restored when they come back online. It also normalizes sample rates across different providers so you can mix and match TTS services without worrying about audio format differences.

## Summary by CodeRabbit

* **New Features**
  * Adds a TTS fallback adapter that orchestrates multiple TTS providers with unified streaming and chunked synthesis, per-provider resampling when needed, availability-change events, and public controls to get streaming instances, synthesize, stream, and close adapters.

* **Bug Fixes**
  * More robust error handling, improved shutdown/cleanup, and per-provider health tracking with configurable recovery retries and delays to improve streaming resilience.

### DIFF
--- a/.changeset/warm-pumpkins-punch.md
+++ b/.changeset/warm-pumpkins-punch.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Add FallbackAdapter for TTS failover support


toubatbrian:
@codex review

agents/src/tts/fallback_adapter.ts
```diff
@@ -0,0 +1,541 @@
...
+              });
+              this.audioPushed = true;
+            }
+          }
+        };
+        const forwardInputPromise = forwardInput();
+        const processOutputPromise = processOutput();
+        let outputError: unknown = null;
+        try {
+          await processOutputPromise;
```
toubatbrian: Should we wait for both promise here?

agents/src/tts/fallback_adapter.ts
```diff
@@ -0,0 +1,541 @@
...
+            if (input === SynthesizeStream.FLUSH_SENTINEL) {
+              stream.flush();
+            } else {
+              this.tokenBuffer.push(input);
+              stream.pushText(input);
+            }
+          }
+          stream.endInput();
+        };
+
```
toubatbrian: In python the forward task is created before the retry loop (i.e. `for (let i = 0; i < this.adapter.ttsInstances.length; i++) {`). Can we follow the same logic here by having a single forwardInput tasks running, and each retry loop content creates a new `StreamChannel` that receives data from the outer global forward input tasks and pushText to the tts stream? Basically it'd be better if we have more logic parity with python. Happy to explain in more details.

agents/src/tts/fallback_adapter.ts
gokuljs: @toubatbrian Actually, we should not wait for both to finish, because that would add delay when switching to the new TTS. We would be waiting for the entire text to be generated. I think we should handle it independently and start receiving text from the LLM as it arrives.

gokuljs: will fix it

agents/src/tts/fallback_adapter.ts
gokuljs: @toubatbrian quick clarification before I implement so that we are aligned. I’m planning to move input forwarding outside the retry loop so it executes once globally. It will read from the input stream and store events in order (text, flush, and end) in the token buffer array.

Then, for each retry attempt, we will create a fresh stream/channel, read from the token buffer array, and send those events to the stream channel. A separate function will handle the output, receive the audio, and store it in an array.

toubatbrian: Yep, that sounds good!

gokuljs: @toubatbrian I have implemented the changes. You can review them in the latest commit .

gokuljs:
@toubatbrian Thanks man

PR #1027, opened 2026-02-05
Migrate user span

toubatbrian:
## Summary by CodeRabbit

* **New Features**
  * Expanded telemetry attributes for assistant turns, provider/tools, response timing (TTFB/TTFT), end-to-end latency, and interruption metrics.
  * Enhanced speech-to-text tracing to include STT model/provider and linked-participant context; spans now carry participant metadata via new span-attribute helpers.
  * Added participant accessors to expose local/linked participant info for span attribution.
  * Exposed a Future.result getter to read resolved values.

* **Tests**
  * Added in-memory tests validating audio recognition span creation, parent/child relationships, and telemetry attributes.

### DIFF
--- a/agents/src/voice/audio_recognition_span.test.ts
+++ b/agents/src/voice/audio_recognition_span.test.ts
@@ -146,7 +146,7 @@ describe('AudioRecognition user_turn span parity', () => {
     // creation-time attributes
     expect(userTurn.attributes['lk.participant_id']).toBe('p1');
     expect(userTurn.attributes['lk.participant_identity']).toBe('bob');
-    expect(userTurn.attributes['lk.participant_kind']).toBe('INGRESS');
+    expect(userTurn.attributes['lk.participant_kind']).toBe('AGENT');
     expect(userTurn.attributes['gen_ai.request.model']).toBe('deepgram-nova2');
     expect(userTurn.attributes['gen_ai.provider.name']).toBe('deepgram');
 


agents/src/voice/audio_recognition.ts
```diff
@@ -58,6 +65,22 @@ export interface AudioRecognitionOptions {
   minEndpointingDelay: number;
   maxEndpointingDelay: number;
   rootSpanContext?: Context;
+  /** STT model name for tracing */
+  sttModel?: string;
+  /** STT provider name for tracing */
+  sttProvider?: string;
```
lukasIO, line 85: any reason why the `model` and `provider` getters shouldn't live directly on the stt node ?

agents/src/voice/audio_recognition.ts
```diff
@@ -58,6 +65,22 @@ export interface AudioRecognitionOptions {
...
+}
+
+/**
+ * Minimal participant shape for span attribution.
+ * Compatible with both `LocalParticipant` and `RemoteParticipant` from `@livekit/rtc-node`.
+ */
+export interface ParticipantLike {
+  sid: string | undefined;
+  identity: string;
+  kind: number;
```
lukasIO: should this be of type `ParticipantKind` instead?

agents/src/utils.ts
```diff
@@ -142,19 +144,33 @@ export class Future<T = void> {
     return this.#done;
   }
 
+  get result(): T {
```
lukasIO, line 147: this doesn't feel great. It introduces a the notion of (false) synchronicity in a - by definition - async utility. 

What speaks against using `await` ?

agents/src/voice/audio_recognition.ts
toubatbrian: This is in python codebase as well, mainly to setup STT related span. STT Node is a function of type

```ts
export type STTNode = (
  modelSettings: ModelSettings,
) => Promise<ReadableStream<SpeechEvent | string> | null>;
```

so we cannot assign model and provider args to it.

agents/src/utils.ts
toubatbrian: This is only been used in 1 place currently, in `RoomIO.linkedParticipant`, which is a getter method that expect synchronous access
```ts
  get isParticipantAvailable(): boolean {
    return this.participantAvailableFuture.done;
  }

  get linkedParticipant(): RemoteParticipant | undefined {
    if (!this.isParticipantAvailable) {
      return undefined;
    }

    return this.participantAvailableFuture.result;
  }
```
Agree that this does not looks great but it's kind of a work around. Also, we have the guard of checking `isParticipantAvailable` before getting the `participantAvailableFuture.result` so it should not throw but simply return undefined if the future is not done in this case

agents/src/voice/audio_recognition.ts
lukasIO: oh, yeah that makes sense

PR #1028, opened 2026-02-05
Fix: Parsing lang from stt ctor

adrian-cowham:
## Summary by CodeRabbit

* **New Features**
  * Speech-to-text accepts model identifiers that include a language (e.g., "provider/model:language") and will derive the language automatically.
  * If a separate language is also provided and conflicts, a warning is shown and the explicit language choice is respected.
  * Internal handling updated so selected model and language are applied consistently.

### DIFF
--- a/.changeset/curly-bees-shave.md
+++ b/.changeset/curly-bees-shave.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Fix parsing lang from stt ctor 


agents/src/inference/stt.ts
```diff
@@ -155,9 +155,28 @@ export class STT<TModel extends STTModels> extends BaseSTT {
       throw new Error('apiSecret is required: pass apiSecret or set LIVEKIT_API_SECRET');
     }
 
+    // Parse language from model string if provided: "provider/model:language"
+    let nextModel = model;
```
toubatbrian, line 219: We might want to change the model type to `ModelWithLanguage`

PR #1042, opened 2026-02-11
fix: dispose native FFI resources before process.exit() in job shutdown

Raysharr:
## Summary

Calls `dispose()` from `@livekit/rtc-node` before `process.exit(0)` in the job process shutdown sequence to properly clean up native FFI resources.

## Problem

When a job process shuts down (e.g., after SIP trunk disconnect), the current shutdown sequence is:

```
session.close() → room.disconnect() → shutdown callbacks → process.exit(0)
```

`room.disconnect()` disconnects from the LiveKit room, but does **not** clean up the Rust FFI Server resources — specifically the tokio async/audio runtimes, FfiRoom instances, and native handles in the DashMap. These resources leak on every job shutdown.

## Fix

Add `await dispose()` after all job cleanup completes but before `process.exit(0)`:

```
session.close() → room.disconnect() → callbacks → dispose() → process.exit(0)
```

`dispose()` calls `livekitDispose()` which:
1. Closes all `FfiRoom` instances (drops track handles, awaits task JoinHandles)
2. Clears the `DashMap` of native handles
3. Shuts down both tokio runtimes (async + audio)
4. Invalidates the FfiServer config

The call is wrapped in try/catch so a failed cleanup never blocks process exit.

## Testing

Tested against a production voice agent (SIP trunk + Silero VAD + Google STT + ElevenLabs TTS + LiveKit turn detector). Confirmed via logs that `dispose()` completes successfully on every SIP disconnect.

**Note:** The `libc++abi: mutex lock failed` crash that sometimes appears on `process.exit(0)` is a separate native-layer issue — it persists regardless of JS-level cleanup (including `dispose()`, ONNX session release, and drain delays). It occurs during C++ destructor ordering and is cosmetic: all job work, IPC messaging, and resource cleanup complete before it fires. See node-sdks#564 for the related native crash.

## Changes

- **`agents/src/ipc/job_proc_lazy_main.ts`** — Import `dispose` from `@livekit/rtc-node`, call it before `process.exit(0)` in the shutdown sequence

## Risk

**Minimal.** `dispose()` is idempotent and designed for exactly this purpose. The try/catch ensures it never blocks exit even if it fails.

### DIFF
--- a/.changeset/dispose-native-ffi.md
+++ b/.changeset/dispose-native-ffi.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Dispose native FFI resources before process.exit() in job shutdown to prevent libc++abi mutex crash
--- a/agents/src/ipc/job_proc_lazy_main.ts
+++ b/agents/src/ipc/job_proc_lazy_main.ts
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: 2024 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import { Room, RoomEvent } from '@livekit/rtc-node';
+import { Room, RoomEvent, dispose } from '@livekit/rtc-node';
 import { EventEmitter, once } from 'node:events';
 import { pathToFileURL } from 'node:url';
 import type { Logger } from 'pino';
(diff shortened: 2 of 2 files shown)


toubatbrian (commented):
Nice catch on this bug! Have you tried testing the agent after this change and saw that the error is gone?

Raysharr (commented):
Thanks! I tested this against our production voice agent (SIP trunk, Silero VAD, Google STT, ElevenLabs TTS, LiveKit turn detector) and ran a deeper investigation.

**`dispose()` works correctly** — after patching, the log confirms `native resources disposed` on every shutdown. This properly cleans up FFI rooms, native handles, and tokio runtimes that were previously leaking on each job exit.

**However**, during testing I discovered that the `libc++abi: mutex lock failed` crash is a separate, deeper issue. It persists even after:
- `dispose()` completes ✅
- ONNX sessions released (VAD + turn detector) ✅
- 3s drain delay before exit ✅

The crash fires at `process.exit(0)` itself — during C++ destructor ordering in the native addon layer. All job work, IPC, and cleanup complete successfully before it. It's cosmetic but noisy.

I'd suggest we:
1. **Merge this PR as-is** — `dispose()` is the correct JS-side cleanup and should have been here regardless. It fixes the resource leak.
2. **Track the mutex crash separately** — it needs a fix in the native Rust/C++ teardown (likely related to [node-sdks#564](https://github.com/livekit/node-sdks/issues/564)).

I've also updated the PR description to reflect these findings.

toubatbrian (approved):

sgzrov:
Filed #1375 — same libc++abi mutex error class, but the JS-side dispose-ordering workaround in this PR doesn't address it (the abort fires even after `native resources disposed`). Cross-linking for searchability.

PR #1044, opened 2026-02-11
Ensure resampling is skipped for empty audio frames

lukasIO:
## Description
While testing I ran into a panic on the underlying rust sdk originating from a null pointer. 
Tracing it back it looks like it was triggered by a `0` value that was passed back after trying to resample an empty audio frame.

## Changes Made

- 
- 
- 

## Pre-Review Checklist

## Testing

- [ ] Automated tests added/updated (if applicable)
- [ ] All tests pass

## Additional Notes

---

### DIFF
--- a/.changeset/fair-needles-obey.md
+++ b/.changeset/fair-needles-obey.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Ensure resampling is skipped for empty audio frames


agents/src/utils.ts
```diff
@@ -651,14 +651,22 @@ export function resampleStream({
...
       if (!resampler) {
         resampler = new AudioResampler(chunk.sampleRate, outputRate);
       }
       for (const frame of resampler.push(chunk)) {
         controller.enqueue(frame);
       }
-      for (const frame of resampler.flush()) {
-        controller.enqueue(frame);
+    },
+    flush(controller) {
```
lukasIO, line 665: This moves the flushing to the `flush` of the transform stream.

I don't know if the previous implemenation was intended to always flush after every insertion? 

@toubatbrian do you happen to remember why this code path is flushing after each chunk?

toubatbrian: Here is the original PR: https://github.com/livekit/agents-js/pull/541

What are the difference between flushing for each frames v.s. flushing at the very end? I guess technically they are the same but maybe flushing on every frame would incur some more performance costs?

PR #1036, opened 2026-02-07
feat: Create MultiInputStream API primitive

toubatbrian:
This class could solves the issue where user want the room to stay alive event after participant disconnected and being able to pipe the audio stream from new participants.

Related PR: https://github.com/livekit/agents-js/pull/1018

## Summary by CodeRabbit

* **New Features**
  * MultiInputStream primitive for managing multiple input streams dynamically.
  * Audio input now exposes a close operation to shut down underlying inputs.

* **Bug Fixes / Reliability**
  * Improved handling of audio recorder and subscribed track lifecycle for cleaner shutdown and interception behavior.

* **Tests**
  * Comprehensive tests covering MultiInputStream behavior, concurrency, edge cases, and error handling.

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents-js/pull/1036" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #1045, opened 2026-02-12
Add AgentTask feature

toubatbrian:


agents/src/stream/deferred_stream.ts
```diff
@@ -118,10 +119,20 @@ export class DeferredReadableStream<T> {
       return;
     }
 
+    const sourceReader = this.sourceReader!;
+    // Clear source first so future setSource() calls can reattach cleanly.
```
lukasIO, line 123: what's the reason for this change and how does it relate to your comment [here](https://github.com/livekit/agents-js/pull/1018#issuecomment-3850284067) ?

agents/src/voice/agent.ts
```diff
@@ -13,26 +13,71 @@ import {
...
 import type { VAD } from '../vad.js';
-import type { AgentActivity } from './agent_activity.js';
+import { type AgentActivity, agentActivityStorage } from './agent_activity.js';
 import type { AgentSession, TurnDetectionMode } from './agent_session.js';
 import type { TimedString } from './io.js';
+import type { SpeechHandle } from './speech_handle.js';
+
+export const functionCallStorage = new AsyncLocalStorage<{ functionCall?: FunctionCall }>();
+export const speechHandleStorage = new AsyncLocalStorage<SpeechHandle>();
+const activityTaskInfoStorage = new WeakMap<Task<any>, _ActivityTaskInfo>();
```
lukasIO, line 41: I haven't used AsyncLocalStorage before, so excuse my naive questions. Can you elaborate a bit on your thinking (and potential tradeoffs) behind the design, e.g. tracking all this in multiple different layered contexts/stores?

agents/src/stream/deferred_stream.ts
toubatbrian: So I made changes here mainly to unblock the current testing flow (as with AgentTask we have to detach the stream when activity paused and reattach it back once resumed). But eventually we would need to switch this to use MultiInputStream once it's merged.

toubatbrian: I will wait for the https://github.com/livekit/agents-js/pull/1036 and do a cleanup here.

agents/src/voice/agent.ts
toubatbrian: Good question. This mirrors how the python uses `contextvars.ContextVar` where each piece of async context has its own independent variable (_SpeechHandleContextVar, _AgentActivityContextVar, _FunctionCallContextVar).

The main reason is they have independent lifecycles: activity context lives for an entire speech task, speech handle is per-pipeline-reply, and function call is scoped to a single tool execution. They nest independently, so bundling them into one object would couple unrelated concerns.

The `WeakMap` for task metadata is intentionally not an `AsyncLocalStorage` because that metadata is tied to a specific Task instance, not to the async execution context. WeakMap is also GC-friendly so we don't leak completed tasks.

agents/src/stream/deferred_stream.ts
toubatbrian: It is ready now

toubatbrian:
@lukasIO can you take another pass, thanks!

PR #1061, opened 2026-02-19
Align agents-js with Python PR #4834 refactor and follow-up fixes

toubatbrian:
## Summary

This PR ports the Python interruption refactor (PR #4834) into agents-js with parity-focused behavior changes, excluding Python-only modeling constructs (dataclass/BaseModel/TypeAdapter internals).

### Ported from Python
- Refactored interruption stream state handling across sentinels/audio flow in interruption_stream.
- Ported follow-up fixes from Python commits:
  - fix shift bug: first-overlap shift logic now mirrors Python behavior.
  - fix buffering bug: continue buffering while agent speech is active (overlap gates inference, not buffering).
  - fix guard: only emit interruption events when overlap is actively tracked.
- Added cache parity helpers in BoundedCache (setOrUpdate, updateValue, popIf) to match Python cache-update/pop patterns used by interruption transports.
- Updated WS/HTTP transport cache update flow to use set-or-update semantics and overlap guards consistent with Python paths.
- Ported overlap sentinel timestamp semantics (startedAt, endedAt) and propagated through voice call-sites (agent_activity, audio_recognition).

### Unit / Time Semantics
- Enforced *InS fields as seconds (totalDurationInS, predictionDurationInS, detectionDelayInS).
- Kept absolute event timestamps in milliseconds (timestamp, overlap start/end sentinel timestamps).
- Hardened WS duration computation by using local request start timing for total-duration calculation while preserving Python-compatible prediction_duration handling (seconds).

### Tests
- Added unit tests for BoundedCache in `agents/src/inference/interruption/utils.test.ts`:
  - eviction, setOrUpdate, updateValue, pop, popIf, empty/no-match paths.

### DIFF
--- a/agents/src/inference/interruption/ws_transport.ts
+++ b/agents/src/inference/interruption/ws_transport.ts
@@ -360,7 +360,6 @@ export function createWsTransport(
    */
   async function reconnect(): Promise<void> {
     close();
-    await ensureConnection();
   }
 
   const transport = new TransformStream<Int16Array | InterruptionEvent, InterruptionEvent>(


agents/src/inference/interruption/interruption_stream.ts
```diff
@@ -350,7 +355,9 @@ export class InterruptionStreamBase {
     this.ensureStreamsNotEnded();
     if (!(frame instanceof AudioFrame)) {
       if (frame.type === 'overlap-speech-started') {
-        this.overlapSpeechStartedAt = Date.now() - frame.speechDurationInS * 1000;
+        // Use the absolute startedAt timestamp from the sentinel directly (computed at call-site).
+        // Ref: python livekit-agents/.../inference/interruption.py _OverlapSpeechStartedSentinel._started_at
+        this.overlapSpeechStartedAt = frame.startedAt;
```
chenghao-mou: I moved away from updating this in this function to avoid race conditions: [commit](https://github.com/livekit/agents/pull/4834/changes/71f940a136748d2c432589a76cb0414470daa3bd) and instead update them during the forward task.

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -37,20 +37,31 @@ export interface WsTransportState {
   cache: BoundedCache<number, InterruptionCacheEntry>;
 }
 
-const wsMessageSchema = z.union([
+const wsMessageSchema = z.discriminatedUnion('type', [
```
lukasIO, line 40: this is a lot nicer!

agents/src/inference/interruption/utils.ts
```diff
@@ -24,6 +24,39 @@ export class BoundedCache<K, V> {
...
+      }
+    }
+    return value;
+  }
+
+  /**
+   * Set a new value with factory when missing; otherwise update in place.
+   * Mirrors python BoundedDict.set_or_update behavior.
+   */
+  setOrUpdate(key: K, factory: () => V, fields: Partial<V>): V {
```
lukasIO, line 49: I left this out on purpose as the only place where it was called was updating all fields, making it redundant to have a separate update method. 

Are there now code paths that only partially update the cache entry?

agents/src/inference/interruption/utils.ts
```diff
@@ -64,6 +97,31 @@ export class BoundedCache<K, V> {
     return undefined;
   }
 
+  /**
+   * Pop a key/value pair if it satisfies the predicate.
+   * Mirrors python BoundedDict.pop_if behavior.
+   */
+  popIf(predicate?: (value: V) => boolean): [K | undefined, V | undefined] {
```
lukasIO: how is this different from the existing pop implementation? I don't think we need both?

agents/src/inference/interruption/utils.ts
toubatbrian: We have 2 places needs this currently. These are mainly to keep consistency to python. It shouldn't hurt if we write in the same way as python I think.

agents/src/inference/interruption/utils.ts
toubatbrian: In python pop_if remove the first / oldest entry if without predicate (FIFO), whereas our logic on pop remove the most recent one. I've removed `popIf` and instead edit the `pop` method

agents/src/inference/interruption/ws_transport.ts
```diff
@@ -360,7 +360,7 @@ export function createWsTransport(
    */
   async function reconnect(): Promise<void> {
     close();
-    // Connection will be re-established on next ensureConnection call
+    await ensureConnection();
```
chenghao-mou: I don't think we need this per codex comment since the reconnection is only for the ws, not the underlying session connection pool.

chenghao-mou (approved):

PR #1065, opened 2026-02-20
fix: handle disconnect during init

Fox32:
## Description

While debugging and edge case I came across this issue: If a participant connects to early (before session init is fully completed) it's later not possible to shutdown the session using `session.shutdown()`.

This seems to be caused by `RoomIO.close` hanging forever - it awaits that the init task is completed after canceling. If the participant disconnects in a way that it happens before the init code is executed, the init task is spawned but never completed. An abort controller is provided, but not every operation inside of it is abortable - `await this.participantAvailableFuture.await` is blocking forever.

I would suggest that this expression should race with the abort controller and just resolves once that is canceled. Hanging in the init task later blocks RoomIO.close() from completing, blocking closing the session.

I could already verify in my project that these changes actually fix the issue.

## Changes Made

- Cancel RoomIO init properly

## Pre-Review Checklist

## Testing

- [x] All tests pass

## Additional Notes

---

### DIFF
--- a/.changeset/moody-ligers-call.md
+++ b/.changeset/moody-ligers-call.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Handle participant disconnect during init


agents/src/voice/room_io/room_io.ts
```diff
@@ -186,7 +183,15 @@ export class RoomIO {
...
-    const participant = await this.participantAvailableFuture.await;
+    const onAbort = new Promise<void>((resolve) => {
+      signal.addEventListener('abort', () => resolve(), { once: true });
+    });
+    const participant = await Promise.race([this.participantAvailableFuture.await, onAbort]);
+
+    if (!participant) {
+      return;
+    }
+
```
toubatbrian: nit: you can actually use `waitForAbort` method exposed from `utils.ts` and race with `this.participantAvailableFuture.await`

agents/src/voice/room_io/room_io.ts
toubatbrian: We might need to apply the same logic here as well

agents/src/voice/room_io/room_io.ts
Fox32: Nice, I needed that helper!

agents/src/voice/room_io/room_io.ts
Fox32: Applied it here too, but I'm not sure how to reproduce this case in my situation, so I could not test it.

PR #1072, opened 2026-02-24
Add TaskGroup feature

toubatbrian:
Add TaskGroup as well as it's dependent features:
- `ChatContext._summarize`
- `agent.updateTools`
- `ToolFlags`

### DIFF
--- a/agents/src/llm/chat_context.ts
+++ b/agents/src/llm/chat_context.ts
@@ -822,6 +822,7 @@ export class ChatContext {
       return this;
     }
 
+    // TODO: refactor this into LLMStream.collect API.
     const promptCtx = new ChatContext();
     promptCtx.addMessage({
       role: 'system',


agents/src/llm/chat_context.ts
```diff
@@ -762,6 +779,111 @@ export class ChatContext {
...
+      role: 'user',
+      content: `Conversation to summarize:\n\n${sourceText}`,
+    });
+
+    const chunks: string[] = [];
+    for await (const chunk of llm.chat({ chatCtx: promptCtx })) {
+      if (chunk.delta?.content) {
+        chunks.push(chunk.delta.content);
+      }
+    }
```
theomonnom, line 844: We don't need it for now, but I also added a better way to call LLMs on Python
https://github.com/livekit/agents/pull/4680

PR #1073, opened 2026-02-25
Remote Session Events

toubatbrian:


agents/src/voice/index.ts
```diff
@@ -5,6 +5,7 @@ export { Agent, StopResponse, type AgentOptions, type ModelSettings } from './ag
 export { AgentSession, type AgentSessionOptions, type VoiceOptions } from './agent_session.js';
 export * from './avatar/index.js';
 export * from './background_audio.js';
+export * from './client_events.js';
```
lukasIO: let's try to be a bit more deliberate about which exports we actually expose to end users here as we'd want to mark all public end user APIs for client_events with an `@experimental` tag for now so that we signal that this API/events are unstable and might change without a major release version

agents/src/voice/client_events.ts
```diff
@@ -0,0 +1,844 @@
...
+    if (this.eventHandlersRegistered) return;
+
+    this.session.on(AgentSessionEventTypes.AgentStateChanged, this.onAgentStateChanged);
+    this.session.on(AgentSessionEventTypes.UserStateChanged, this.onUserStateChanged);
+    this.session.on(AgentSessionEventTypes.ConversationItemAdded, this.onConversationItemAdded);
+    this.session.on(AgentSessionEventTypes.FunctionToolsExecuted, this.onFunctionToolsExecuted);
+    this.session.on(AgentSessionEventTypes.MetricsCollected, this.onMetricsCollected);
+    this.session.on(AgentSessionEventTypes.UserInputTranscribed, this.onUserInputTranscribed);
+    this.session.on(AgentSessionEventTypes.UserInterruptionDetected, this.onUserOverlapSpeech);
+    this.session.on(AgentSessionEventTypes.UserNonInterruptionDetected, this.onUserOverlapSpeech);
```
chenghao-mou, line 328: We merged `InterruptionDetected` and `NonInterruptionDetected` into one `OverlappingSpeechEvent`: [commit 1](https://github.com/livekit/agents/pull/4883/changes/ef7bd81530e053f613ede82904b3c95dca0e8e6b) and [commit 2](https://github.com/livekit/agents/pull/4883/changes/8f114a267e241428348c5537f35de2eb83cce8e0)

toubatbrian: Will start migrating that PR in a different PR.

PR #1093, opened 2026-03-03
Add OpenAI Response WebSocket API

toubatbrian:
Use with
```ts
llm: openai.responses.LLM({ useWebsocket: true })
```

Fully tested with unit tests and e2e tests

### DIFF
--- a/plugins/openai/src/responses/llm.ts
+++ b/plugins/openai/src/responses/llm.ts
@@ -358,6 +358,11 @@ export class LLM extends llm.LLM {
     } else {
       this.#llm = new ResponsesHttpLLM({ ...baseOpts, client });
     }
+
+    // Forward events from the inner delegate so consumers listening on this
+    // wrapper instance (e.g. AgentActivity) receive them.
+    this.#llm.on('metrics_collected', (metrics) => this.emit('metrics_collected', metrics));
+    this.#llm.on('error', (error) => this.emit('error', error));
   }
 
   override label(): string {


examples/src/basic_agent_task.ts
```diff
@@ -117,7 +117,7 @@ export default defineAgent({
     const session = new voice.AgentSession({
       vad: ctx.proc.userData.vad as silero.VAD,
       stt: new inference.STT({ model: 'deepgram/nova-3' }),
-      llm: new openai.ws.LLM(),
+      llm: new openai.responses.LLM({ useWebSocket: true }),
```
tinalenguyen, line 120: nit: `useWebSocket` is `true` by default, should we omit this?

plugins/openai/src/ws/index.ts
```diff
@@ -0,0 +1,6 @@
+// SPDX-FileCopyrightText: 2025 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+export { ResponsesWebSocket } from './llm.js';
+export * from './types.js';
```
lukasIO: do we need to export the schemas of the types as well? 

might be leaner if we only exported the inferred types here?

plugins/openai/src/ws/llm.ts
```diff
@@ -0,0 +1,609 @@
...
+      }
+      this.#outputQueue = [];
+    });
+  }
+
+  /**
+   * Send a response.create event.  Returns a typed `StreamChannel<WsServerEvent>`
+   * that yields validated server events until the response terminates.
+   */
+  sendRequest(payload: WsResponseCreateEvent): stream.StreamChannel<WsServerEvent> {
```
lukasIO, line 106: there's no additional error handling here for cases in which the ws is closed already. 

question:
Should we check for ws connection state and error with a descriptive error?
Should we perform an attempt to auto reconnect the ws connection in cases the close wasn't initiated by the user/agent pipeline?

examples/src/basic_agent_task.ts
toubatbrian: Yes, but since this is example code, I’d like to keep it explicit so users are aware that this feature is available.

PR #1094, opened 2026-03-04
Support Image Input for OpenAI realtime model

toubatbrian:
Repro-ed locally and is working properly


Sohaib-Nisar-QFN:
When will it approve

Issue #1015, opened 2026-02-03
Feature Needed: Inworld Inference Model Options


tommymsz006:
### Your question

I have been using Inworld inference model with LiveKit JS agent, and it works quite well. I wonder if it's possible to make extra model options available. Currently it only supports the basics, namely language, voice and model. I would like to fine-tune speaking rate, sample rate, temperature and text normalization. Per the below, it seems that these are not open:

https://github.com/livekit/agents-js/blob/efa8a4b701a7f115d4246b82bf200a3ece905636/agents/src/inference/tts.ts#L62

Can we make these options available as well?

PR #1102, opened 2026-03-06
Support gateway Inworld model options

toubatbrian:
https://github.com/livekit/agents-js/issues/1015

### DIFF
--- a/.changeset/rude-feet-drum.md
+++ b/.changeset/rude-feet-drum.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Support gateway Inworld model options


agents/src/inference/tts.ts
```diff
@@ -62,7 +62,14 @@ export interface DeepgramTTSOptions {}
...
 export interface RimeOptions {}
 
-export interface InworldOptions {}
+export interface InworldOptions {
+  /** Controls how fast the voice speaks. 1.0 is normal speed, 0.5 is half, 1.5 is 1.5x. Default: 1.0. */
+  speaking_rate?: number;
+  /** Controls randomness in the output. Recommended between 0.6 and 1.1. Default: 1.1. */
+  temperature?: number;
+  /** Controls text normalization. "ON" expands numbers, dates, abbreviations. "OFF" reads text as written. Default: "ON". */
+  text_normalization?: 'ON' | 'OFF';
```
theomonnom, line 71: I don't really like this API. Vs a boolean. I think we should discuss with the gateway team

PR #1104, opened 2026-03-06
Add GPT-5.4 to OpenAI plugin

Topherhindman:
## Summary
- Add `gpt-5.4` to `ChatModels` type in OpenAI plugin
- Add `gpt-5.4` to `supportsReasoningEffort()`
- Backfill missing `gpt-5.1`, `gpt-5.1-chat-latest`, `gpt-5.2`, `gpt-5.2-chat-latest` in `ChatModels`

### DIFF
--- a/agents/src/inference/llm.ts
+++ b/agents/src/inference/llm.ts
@@ -17,7 +17,6 @@ import { type AnyString, createAccessToken } from './utils.js';
 const DEFAULT_BASE_URL = 'https://agent-gateway.livekit.cloud/v1';
 
 export type OpenAIModels =
-  | 'openai/gpt-5.4'
   | 'openai/gpt-5.2'
   | 'openai/gpt-5.2-chat-latest'
   | 'openai/gpt-5.1'


agents/src/inference/llm.ts
```diff
@@ -17,6 +17,7 @@ import { type AnyString, createAccessToken } from './utils.js';
 const DEFAULT_BASE_URL = 'https://agent-gateway.livekit.cloud/v1';
 
 export type OpenAIModels =
+  | 'openai/gpt-5.4'
```
bcherry: this one isn't for the plugin its for inference, which should be done separately

Topherhindman: Reason being we want the agent gateway/inference side to go out first?

bcherry: yeah feels like separate PRs to me. also the title of this one said "plugin"

Topherhindman: yeah i ended up slamming the inference part in when reviewing the agents work.  i think its fair to have that be a separate PR though.

Issue #1108, opened 2026-03-07
IPC safeSend should not throw unhandled errors on channel close during shutdown


KlementMultiverse:
## Problem

In src/worker/ipc.ts, the `safeSend()` function throws unhandled errors when the message channel closes during normal shutdown sequences. This occurs in production when:

1. Agent job completes and initiates graceful shutdown
2. Parent process closes the IPC channel
3. Worker still attempts to send status updates via `safeSend()`
4. Unhandled error crashes the worker process before cleanup completes

Related: #1080, #1089

## Current Behavior

```typescript
// src/worker/ipc.ts - safeSend throws on closed channel
safeSend(message) {
  if (!this.port) throw new Error('IPC port not available');
  this.port.postMessage(message); // throws if port is closed
}
```

When the channel is already closed (expected during shutdown), this throws an unhandled error that terminates the worker.

## Expected Behavior

`safeSend()` should gracefully handle channel-closed conditions:
- Detect closed channels before attempting to send
- Log a debug message (not an error) for expected closures
- Return a status indicator so callers can determine if the message was sent
- Reserve exceptions only for unexpected failures (e.g., serialization errors)

## Why This Matters

The job lifecycle requires clean shutdown: job completion → status update → worker cleanup → exit. When safeSend throws during this sequence, the worker terminates without:
- Flushing logs
- Releasing resources
- Notifying the parent of final state

This creates orphaned worker processes and makes debugging job failures harder.

## Suggested Fix

Wrap `postMessage()` in a try-catch that distinguishes between:
1. **Expected closes** (port.onclose already fired): log at debug level, return false
2. **Unexpected errors** (serialization, state issues): log error, return false or throw
3. **Success**: return true

This allows the job completion flow in src/worker/job.ts to continue gracefully even if the parent has already closed the channel.

## Testing

A reproduction would:
1. Create an agent job
2. Trigger parent disconnection while job is mid-shutdown
3. Verify worker exits cleanly without throwing
4. Confirm final job state is recorded (regardless of whether IPC message reached parent)

---
*Contributed by [Klement Gunndu](https://github.com/KlementMultiverse)*

Issue #1089, opened 2026-03-02
Disconnect during job start causes the agent to hang


Fox32:
### Describe the bug

If I disconnect after `received job request` and before the session is started, the job hangs and is killed with `job is unresponsive`

### Relevant log output

In the end the job is hard killed after a `job is unresponsive` log entry.

```
[15:56:05.303] DEBUG (88816): job exiting
[15:57:04.963] WARN (88816): job is unresponsive
[15:57:04.963] DEBUG (89120): SIGTERM received in job proc
    jobID: "AJ_67dQtGe5Ck2E"
```

### Describe your environment

`1.0.48`

### Minimal reproducible example

I have very simple agent with a greeting inside of `onEnter`:

```ts
class CustomAgent extends voice.Agent {
    constructor() {
        super({
            instructions:
                "You are a helpful assistant, you can hear the user's message and respond to it.",
        })
    }

    async onEnter(): Promise<void> {
        // Note that I have a greeting on enter
        this.session.say('Hello, how can I help you today?', { allowInterruptions: false })
    }
}

// …

const agent = new CustomAgent()

const session = new voice.AgentSession({
    stt: 'assemblyai/universal-streaming:en',
    llm: 'openai/gpt-4.1-mini',
    tts: 'cartesia/sonic-2:9626c31c-bec5-4cca-baa8-f8ba9e84c8bc',
    vad: jobContext.proc.userData.vad! as silero.VAD,
    turnDetection: new turnDetector.MultilingualModel(),
    voiceOptions: {
        preemptiveGeneration: true,
    },
    connOptions: {
        llmConnOptions: {
            maxRetry: 1,
            retryIntervalMs: 2000,
            timeoutMs: 60000,
        },
    },
})

// Inside the entrypoint I first sleep to make it easier to get to this error:
await new Promise((resolve) => setTimeout(resolve, 2000))

await session.start({
    agent,
    room: jobContext.room,
    inputOptions: {
        noiseCancellation: BackgroundVoiceCancellation(),
    },
})
```

I added the timeout to make it easier to reproduce the issue.

### Additional information

I already tried to debug the issue: 

If I connect and disconnect during the 2s timeout, I have the agent hanging, ultimately resulting in `job is unresponsive`.

My idea what the problem is from initial debugging:
1. Inside the `shutdown` code of the agent session at some point `closeImplInner` is called with `drain = false` (default value).
2. `activity.interrupt` is therefore called and awaited (here the shutdown code is hanging on the await!)
[...]

Issue #1080, opened 2026-02-27
IPC safeSend in 1.0.48 throws unhandled error on channel close during normal operation


rinarakaki:
### Describe the bug

In `@livekit/agents@1.0.48`, the new `safeSend` wrapper in `ipc/job_proc_lazy_main.js` throws `new Error("IPC channel closed")` when `process.connected` is `false` during an inference request. This causes the job process to crash with an unhandled exception during normal operational scenarios (e.g., shutdown, reconnection, participant disconnect).

In 1.0.46, `process.send()` would either fail silently or emit a Node `ERR_IPC_CHANNEL_CLOSED` error event, which was less disruptive. The 1.0.48 change (likely intended to address #211) converts this into an explicit `throw`, which is too aggressive for cases where the IPC channel closing is expected.

### Relevant code (1.0.48)

```js
// dist/ipc/job_proc_lazy_main.js
const safeSend = (msg) => {
  if (process.connected && process.send) {
    process.send(msg);
    return true;
  }
  return false;
};

// For inference requests — throws on channel close:
if (!safeSend({ case: "inferenceRequest", value: { requestId, method, data } })) {
  throw new Error("IPC channel closed");  // <-- crashes the job process
}
```

The `safeSend` pattern itself is good, but throwing on inference requests causes unhandled crashes. Other call sites (e.g., `"exiting"`, `"done"`, `"initializeResponse"`) correctly use `safeSend` without throwing, just returning `false`.

### Expected behavior

When the IPC channel is closed during an inference request, the job process should gracefully abort the current operation (e.g., log a warning, reject the pending promise, and clean up) rather than throwing an unhandled error that crashes the process.

### Steps to reproduce

1. Run an agent with `@livekit/agents@1.0.48`
2. Have a participant disconnect or the room close while an inference request is in flight
3. The job process crashes with `Error: IPC channel closed`

### Workaround

Pin `@livekit/agents` to `1.0.46` to avoid the regression.

### Environment

- `@livekit/agents`: 1.0.48 (regression), 1.0.46 (works)
- Node.js: 24.x
- Platform: ECS EC2 (Linux)
- STT: Deepgram
- Using custom agent (not voice pipeline LLM)

### Additional context

Related to #211 (Intermittent Agent Error: Channel Closed). The `safeSend` wrapper was likely introduced to fix #211, but the `throw` on inference requests makes the problem worse in production — agents crash instead of recovering gracefully.

rinarakaki:
[...]

PR #1109, opened 2026-03-07
fix: handle graceful shutdown in safeSend to prevent worker crashes

haroldfabla2-hue:
## Summary

Fixes issue #1108: IPC safeSend should not throw unhandled errors on channel close during shutdown.

## Problem

In `src/ipc/job_proc_lazy_main.ts`, the `safeSend()` function throws unhandled errors when the message channel closes during normal shutdown sequences. This occurs in production when:

1. Agent job completes and initiates graceful shutdown
2. Parent process closes the IPC channel
3. Worker still attempts to send status updates via `safeSend()`
4. Unhandled error crashes the worker process before cleanup completes

## Solution

Wrap `process.send()` in a try-catch block that:
- Returns `false` instead of throwing when channel is closed (expected during shutdown)
- Logs at debug level for expected closures (not error level)
- Only throws for unexpected errors (e.g., serialization errors)

This allows the job completion flow in `src/ipc/job_proc_lazy_main.ts` to continue gracefully even if the parent has already closed the channel.

## Changes

- Modified `safeSend()` function in `agents/src/ipc/job_proc_lazy_main.ts`
- Added try-catch wrapper around `process.send()`
- Added debug logging for channel close errors

## Testing

The fix allows workers to exit cleanly without throwing when the IPC channel is already closed during shutdown sequence. Related issues: #1080, #1089.

### DIFF
--- a/agents/src/ipc/job_proc_lazy_main.ts
+++ b/agents/src/ipc/job_proc_lazy_main.ts
@@ -16,11 +16,19 @@ import type { IPCMessage } from './message.js';
 const ORPHANED_TIMEOUT = 15 * 1000;
 
 const safeSend = (msg: IPCMessage): boolean => {
-  if (process.connected && process.send) {
-    process.send(msg);
-    return true;
+  try {
+    if (process.connected && process.send) {
+      process.send(msg);
+      return true;
+    }
+    // Channel closed or not available - expected during shutdown
+    return false;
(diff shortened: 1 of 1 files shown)


toubatbrian:
I will close this since we already have https://github.com/livekit/agents-js/pull/1110

PR #1132, opened 2026-03-15
fix(IPC): graceful handling when channel closes during inference

### LINKED ISSUE #1080: IPC safeSend in 1.0.48 throws unhandled error on channel close during normal operation

haroldfabla2-hue:
## Summary

When the IPC channel closes during an inference request (which happens during normal shutdown scenarios like participant disconnect, room close, or reconnection), the job process was crashing with an unhandled exception.

## Changes

- Improved error message to be more descriptive and indicate this is an expected scenario during normal operation
- Added debug logging with context (method, requestId) to help diagnose issues in production
- The error now clearly indicates the inference was aborted gracefully rather than crashing the process

## Related Issue

Fixes #1080

## Testing

This fix addresses the regression introduced in 1.0.48 where  was modified to throw on inference requests. The change makes the error handling consistent with other call sites in the codebase that correctly use `safeSend` without throwing.

### DIFF
--- a/.changeset/tender-donuts-collect.md
+++ b/.changeset/tender-donuts-collect.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+fix(IPC): graceful handling when channel closes during inference


toubatbrian:
@haroldfabla2-hue Can you fix the lint error?

PR #1136, opened 2026-03-16
Initial adoption of throws-transformer

lukasIO:
## Changes Made

- 
- 
- 

## Pre-Review Checklist

## Testing

- [ ] Automated tests added/updated (if applicable)
- [ ] All tests pass

## Additional Notes

---

### DIFF
--- a/agents/package.json
+++ b/agents/package.json
@@ -37,6 +37,7 @@
   },
   "devDependencies": {
     "@livekit/rtc-node": "catalog:",
+    "@livekit/throws-transformer": "^0.1.5",
     "@microsoft/api-extractor": "^7.35.0",
     "@types/fluent-ffmpeg": "^2.1.28",
     "@types/json-schema": "^7.0.15",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
   "devDependencies": {
     "@changesets/cli": "^2.30.0",
     "@livekit/changesets-changelog-github": "^0.0.4",
-    "@livekit/throws-transformer": "^0.1.4",
     "@rushstack/heft": "^0.66.0",
     "@trivago/prettier-plugin-sort-imports": "^4.3.0",
     "@types/node": "^22.13.10",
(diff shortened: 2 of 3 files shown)


agents/src/llm/fallback_adapter.ts
```diff
@@ -296,8 +298,9 @@ class FallbackLLMStream extends LLMStream {
 
   /**
    * Main run method - iterates through LLMs with fallback logic.
+   * @throws {APIConnectionError} When all LLM providers have been exhausted
    */
-  protected async run(): Promise<void> {
+  protected async run(): Promise<Throws<undefined, APIConnectionError>> {
```
1egoman: nitpick: Should this be `Throws<void, APIConnectionError>`? I had been tending to use `void` in the client-sdk-js usages of `Throws` over undefined where the return doesn't matter since it is in the return position, for example [here](https://github.com/livekit/client-sdk-js/blob/c5a8d771b81cf8b7b10c12782a0d4cb46845dfd4/src/room/data-track/incoming/IncomingDataTrackManager.ts#L462). Maybe there was a reason you did it like this though?

lukasIO: interesting, the reason I did this way is because `void` would show an error on early returns:

```
Type 'undefined' is not assignable to type 'void & { readonly __throws?: APIConnectionError | undefined; }'
```
you aren't seeing this?

1egoman: I have not seen this, no - interesting....

Just a guess (and this wouldn't explain why you are seeing it here but I am not seeing it in client-sdk-js...) but I wonder if `void` needs to be added here: https://github.com/livekit/throws-transformer-ts/blob/faf66bc28bc5128547533f17f452c95b323f5469/src/throws.ts#L1. There were some similar errors like this I worked through at one point with other types like numbers / booleans / etc like this and that `Primatives` type / the associated `Exclude` in `Throws` was how I worked around it.

1egoman: Actually, I take this back. After migrating to `@livekit/throws-transformer` on the data tracks branch, I do now see it on client-sdk-js! Here's the relevant ci build: https://github.com/livekit/client-sdk-js/actions/runs/23155025072/job/67267267613.

Update: yes, what I suspected was the issue: https://github.com/livekit/throws-transformer-ts/pull/5

PR #1149, opened 2026-03-19
Add retry semantics and error recovery for adaptive interruption detection

toubatbrian:
## Summary
Ports Python's recoverable/unrecoverable retry semantics for adaptive interruption detection (agents#5142) to the JS SDK, and fixes a JS-specific bug where interruption detection becomes permanently dead after a transient transport error.
### Changes
**Transport error classification** (`http_transport.ts`, `ws_transport.ts`)
- Wrap transport errors as `APIError` subclasses (`APIStatusError`, `APIConnectionError`, `APITimeoutError`) instead of generic `Error` — enables the retry loop to classify errors via `isAPIError()` and `.retryable`
- Disable `ofetch`'s internal retry (`retry: 0`) so retry logic is centralized in the stream consumer
- Propagate errors via `controller.error()` instead of swallowing them with `logger.error`
- Remove internal retry loop from WS `ensureConnection()` (retry is now handled at the stream level)
**Retry loop in `createInterruptionTask`** (`audio_recognition.ts`)
- Restructure from a single-stream-per-session design to a `while` loop that creates a fresh `InterruptionStreamBase` on each retry attempt, matching Python's `_main_task` pattern
- Classify caught errors: retryable `APIError` → emit recoverable error + retry with exponential backoff; non-retryable or retries exhausted → emit unrecoverable error + break (triggers VAD fallback)
- Catch `forwardTask` rejections in `finally` to prevent unhandled rejections from breaking the retry loop
- **Re-inject `agent-speech-started` sentinel on retry** — JS creates a new stream per retry (unlike Python where `self._agent_speech_started` persists on the instance), so the new stream's `agentSpeechStarted` closure variable must be re-initialized
**Type and logic alignment** (`interruption.ts`, `agent_activity.ts`, `report.ts`)
- Remove `false` from `InterruptionOptions.mode` union type to match Python's `Literal["adaptive", "vad"]`
- Simplify `resolveInterruptionDetector` conditions accordingly
- Fix `allowInterruptions` computation in session report
## Test plan
- [x] Build passes (`pnpm build --filter agents`)
- [x] Tested transient error scenario: single HTTP 500 → retry → adaptive interruption resumes and successfully detects interruptions
- [x] Tested persistent error scenario: repeated HTTP 500s → retries exhaust → emits unrecoverable error → falls back to VAD → user can still interrupt via VAD
- [x] Verified `onInterruptionError` correctly routes recoverable errors (log only) vs unrecoverable errors (fallback to VAD)
- [x] Error handling is transport-agnostic: both HTTP and WS transports propagate `APIError` subclasses through `controller.error()`, consumed identically by the retry loop

### DIFF
--- a/agents/src/voice/audio_recognition.ts
+++ b/agents/src/voice/audio_recognition.ts
@@ -1115,7 +1115,7 @@ export class AudioRecognition {
               `failed to detect interruption, retrying in ${retryInterval}ms`,
             );
             numRetries++;
-            await delay(retryInterval);
+            await delay(retryInterval, { signal });
           }
         } else {
           const msg = e instanceof Error ? e.message : String(e);


agents/src/voice/turn_config/interruption.ts
```diff
@@ -15,7 +15,7 @@ export interface InterruptionOptions {
    * voice-activity detection. `undefined` means auto-detect.
    * @defaultValue undefined
    */
-  mode: 'adaptive' | 'vad' | false | undefined;
+  mode: 'adaptive' | 'vad' | undefined;
```
toubatbrian, line 18: Remove `false` to keep aligned with python impl

agents/src/voice/audio_recognition.ts
```diff
@@ -1009,77 +1011,127 @@ export class AudioRecognition {
...
+        try {
+          signal.removeEventListener('abort', cleanup);
+          eventReader.releaseLock();
+          await stream.close();
+        } catch (e) {
+          this.logger.debug('createInterruptionTask: error during cleanup:', e);
+        }
+      };
+
+      signal.addEventListener('abort', cleanup, { once: true });
```
lukasIO, line 1031: above this we have a `await stream.push` which is async, so the abort event listener should be registered before and the clean up should also be called in cases where `stream.push` throws

agents/src/voice/audio_recognition.ts
```diff
@@ -1009,77 +1011,127 @@ export class AudioRecognition {
...
+                interruptionDetection.label,
+                true,
+              ),
+            );
+            this.logger.warn(
+              { model: interruptionDetection.label, attempt: numRetries },
+              `failed to detect interruption, retrying in ${retryInterval}ms`,
+            );
+            numRetries++;
+            await delay(retryInterval);
```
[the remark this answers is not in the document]
lukasIO: this is a good point

lukasIO (approved):
apart from my other comment about catching abort before the async push, this looks good to me. 

Generally not a fan of the nested while loop as it introduces a lot of logic flow through multiple `break` levels, but we can address this afterwards also.

PR #1160, opened 2026-03-20
fix: Align inference TTS provider options

toubatbrian:
## Summary

This updates JS inference.TTS provider option typings

The main user-facing fix is for ElevenLabs: modelOptions now accepts flat inference extras like speed, stability, and similarity_boost.


davidzhao (approved):
lgtm!

theomonnom (approved):

PR #1162, opened 2026-03-21
Fix redundant realtime events 

toubatbrian:
## Summary
- Avoid sending `response.cancel` when there is no in-flight realtime response, which removes spurious cancellation errors during tool follow-up
- Avoid sending no-op `session.update` events when `toolChoice` is unchanged.
- Make duplicate `conversation.item.added` handling idempotent when the same user item was already inserted locally, preventing duplicate-ID failures

### DIFF
--- a/plugins/openai/src/realtime/realtime_model.ts
+++ b/plugins/openai/src/realtime/realtime_model.ts
@@ -1257,13 +1257,11 @@ export class RealtimeSession extends llm.RealtimeSession {
     const incomingItem = openAIItemToLivekitItem(event.item);
     const existingItem = this.remoteChatCtx.get(event.item.id);
     const pendingCreateFuture = this.itemCreateFutures[event.item.id];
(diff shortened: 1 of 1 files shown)


plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -1236,13 +1253,26 @@ export class RealtimeSession extends llm.RealtimeSession {
       throw new Error('item.id is not set');
     }
 
+    const serverEventType = event.type as string;
+    const incomingItem = openAIItemToLivekitItem(event.item);
+    const existingItem = this.remoteChatCtx.get(event.item.id);
+    const pendingCreateFuture = this.itemCreateFutures[event.item.id];
+    if (existingItem) {
+      if (serverEventType === 'conversation.item.added' && !pendingCreateFuture) {
```
davidzhao: nit: combine this into the same if?

PR #1173, opened 2026-03-24
fix: handle unhandled 'error' event on FfmpegCommand in audio.ts

enriqueespaillat-gyde:
## Summary

In `audioFramesFromFile()` (`agents/src/audio.ts`), the `FfmpegCommand` instance has no `'error'` listener. fluent-ffmpeg emits `'error'` events on the command object separately from the piped output stream. When the `AgentSession` closes (LLM error, user disconnect, or any teardown), the output stream closes and fluent-ffmpeg's internal timer fires `Error: Output stream closed` on the `FfmpegCommand`. With no listener, Node.js throws an unhandled error and crashes the entire voice-agents worker process:

```
node:events:497
      throw er; // Unhandled 'error' event
Error: Output stream closed
    at Timeout._onTimeout (fluent-ffmpeg/lib/processor.js:489:25)
Emitted 'error' event on FfmpegCommand instance at:
    at emitEnd (fluent-ffmpeg/lib/processor.js:422:16)
    at Timeout._onTimeout (fluent-ffmpeg/lib/processor.js:489:17)
```

## Root Cause

Error handlers are attached to `outputStream` (the result of `command.pipe()`) but not to `command` itself. fluent-ffmpeg emits `'error'` events on both objects independently — the piped stream listener doesn't catch errors emitted on the command.

## Fix

Added `command.on('error', ...)` before `command.pipe()`, using the same suppression pattern already established in `recorder_io.ts` (lines 204–217). Known teardown errors (`Output stream closed`, `received signal 2`, `SIGKILL`, `SIGINT`) are suppressed with a debug log; unexpected errors are logged at error level.

```diff
+  command.on('error', (err: Error) => {
+    if (
+      err.message?.includes('Output stream closed') ||
+      err.message?.includes('received signal 2') ||
+      err.message?.includes('SIGKILL') ||
+      err.message?.includes('SIGINT')
+    ) {
+      // Expected during teardown — not an error
+      logger.debug('FFmpeg command ended during shutdown');
+    } else {
+      logger.error(err, 'FFmpeg command error');
+    }
+    commandRunning = false;
+    onClose();
+  });
+
   const outputStream = command.pipe();
```

## Why `recorder_io.ts` doesn't have this bug

`recorder_io.ts` (lines 192–218) correctly chains `.on('error', ...)` directly on the `FfmpegCommand` and suppresses the same set of expected shutdown errors. This fix brings `audio.ts` in line with that existing pattern.

## Reproduction

1. Start a voice agent with background audio enabled (e.g., `OFFICE_AMBIENCE`)
2. Trigger an `AgentSession` close while audio is playing (easiest: cause an LLM connection failure that exhausts retries)
3. The process crashes with `Emitted 'error' event on FfmpegCommand instance`

## Pre-Review Checklist
- [x] **Build passes**: Change is additive error handling only
- [x] **Changes explained**: All changes documented above
- [x] **Scope appropriate**: Single-file fix, one logical change

### DIFF
--- a/agents/src/utils.ts
+++ b/agents/src/utils.ts
@@ -753,12 +753,7 @@ export function isStreamClosedError(error: unknown): boolean {
 }
 
 /** FFmpeg error messages expected during normal teardown/shutdown. */
-const FFMPEG_TEARDOWN_ERRORS = [
-  'Output stream closed',
-  'received signal 2',
-  'SIGKILL',
-  'SIGINT',
-];
+const FFMPEG_TEARDOWN_ERRORS = ['Output stream closed', 'received signal 2', 'SIGKILL', 'SIGINT'];
 
 /**
  * Check if an error is an expected FFmpeg teardown error that can be safely ignored during cleanup.
@@ -768,8 +763,7 @@ const FFMPEG_TEARDOWN_ERRORS = [
  */
 export function isFfmpegTeardownError(error: unknown): boolean {
   return (
-    error instanceof Error &&
-    FFMPEG_TEARDOWN_ERRORS.some((msg) => error.message?.includes(msg))
+    error instanceof Error && FFMPEG_TEARDOWN_ERRORS.some((msg) => error.message?.includes(msg))
   );
 }
 


agents/src/audio.ts
```diff
@@ -143,6 +143,22 @@ export function audioFramesFromFile(
     }
   };
 
+  command.on('error', (err: Error) => {
+    if (
+      err.message?.includes('Output stream closed') ||
+      err.message?.includes('received signal 2') ||
+      err.message?.includes('SIGKILL') ||
+      err.message?.includes('SIGINT')
```
toubatbrian: nit: can you extract those list of error messages in an array and check via `FFMPEG_ERRORS.some(...)`?

enriqueespaillat-gyde: Moved to a shared isFfmpegTeardownError() helper in utils.ts, alongside the existing isStreamClosedError() pattern. Let me know if that looks good, happy to just use a constant.

toubatbrian (approved):

PR #1174, opened 2026-03-25
fix: add idle timeouts to TTS stream reads to prevent agent stuck in speaking state

toubatbrian:
## Summary
- Add per-message idle timeout (10s, matching Python's ws.receive(timeout=...)) to the inference TTS WebSocket recv loop, so a gateway that never sends done no longer blocks the pipeline indefinitely
- Add defense-in-depth idle timeouts (10s) in _performTTSInferenceImpl and forwardAudio read loops that activate after the first frame, covering all TTS providers
- Extract a reusable waitUntilTimeout utility with Throws<T, E> branded return type for encoding thrown errors at the type level

## Context
Traced from customer session `RM_8p6KE6emDfkd` where the Inference Gateway TTS WebSocket stopped sending messages after producing initial audio frames. The tts_node span never ended, causing the pipeline to block at `waitIfNotInterrupted(tasks)` for 66 seconds with the agent stuck in speaking state and interruption detection disabled. Python's inference TTS client already had a per-message `ws.receive(timeout=10)` guard; the JS client had none.

## Test plan

- [x] Unit tests: mock TTS streams that produce frames then stall — verify timeout kicks in and tasks complete
- [x] Unit tests: normal TTS stream closure still works without hitting timeout
- [x] Manual test with inference TTS + multi-tool-call flow to exercise the tool response continuation path

### DIFF
--- a/agents/src/voice/generation.ts
+++ b/agents/src/voice/generation.ts
@@ -640,9 +640,10 @@ export function performTTSInference(
           break;
         }
 
-        const { done, value: frame } = firstByteReceived
-          ? await waitUntilTimeout(ttsStreamReader.read(), TTS_READ_IDLE_TIMEOUT_MS)
-          : await ttsStreamReader.read();
+        const { done, value: frame } = await waitUntilTimeout(
+          ttsStreamReader.read(),
+          TTS_READ_IDLE_TIMEOUT_MS,
+        );
         if (done) {
           break;
         }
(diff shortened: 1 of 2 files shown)


agents/src/voice/generation.ts
```diff
@@ -791,9 +809,13 @@ async function forwardAudio(
         break;
       }
 
-      const { done, value: frame } = await reader.read();
+      const { done, value: frame } = hasReceivedFrame
```
davidzhao: do we not want to wait also on the first frame? what if TTS doesn't return any frames back

PR #1177, opened 2026-03-27
Reuse STT Pipeline Across Agent Handoff

toubatbrian:
Changes' been verified


agents/src/voice/audio_recognition.ts
```diff
@@ -69,6 +70,49 @@ export interface RecognitionHooks {
   retrieveChatCtx: () => ChatContext;
 }
 
+export class STTPipeline {
```
tinalenguyen, line 73: should we rename to `_STTPipeline`?

toubatbrian: It should be fine cuz we are not exposing STTPipeline to end user.

tinalenguyen (approved):

PR #1213, opened 2026-04-07
Resolve 4 Detail bugs (maxToolSteps, JWT caching, fallback concurrency, mergeFrames perf)

toubatbrian:
### Summary
- Fix `SpeechHandle._numSteps` initialization so `maxToolSteps` limit is actually enforced in realtime tool-call loops
- Add JWT expiry tracking to telemetry exporters to prevent permanent log/trace export failure after 6 hours
- Cap `pendingLogs` in `PinoCloudExporter` at 10k to prevent OOM on persistent export failure
- Remove shared `LLM` error listener in `FallbackLLMStream` that caused concurrent request cross-contamination
- Rewrite `mergeFrames` from O(N^2) array spread to O(N) pre-allocated `TypedArray.set()` 
A fifth bug (Queue dropping falsy values) was already fixed on the current branch.

### Changes
- **`agents/src/voice/speech_handle.ts`**
`_numSteps` was hardcoded to `1`, ignoring the `_stepIndex` constructor arg. Now initialized from `_stepIndex` so the `maxToolSteps` check in `agent_activity.ts` sees the correct accumulated step count.

- **`agents/src/telemetry/otel_http_exporter.ts`** and **`agents/src/telemetry/pino_otel_transport.ts`**
`ensureJwt()` cached the token forever (`if (this.jwt) return`). Now tracks `jwtExpiresAt` and refreshes 1 hour before the 6-hour server TTL. Also caps `pendingLogs` at 10k entries in the flush error handler to prevent unbounded memory growth.

- **`agents/src/llm/fallback_adapter.ts`**
`FallbackLLMStream` attached an error handler to the shared `LLM` instance, capturing errors from unrelated concurrent requests. Removed the `llm.on('error')` listener, `streamError` variable, and post-loop `if (streamError) throw` check. Errors are already caught via the `try/catch` around `for await (const chunk of stream)`.

- **`agents/src/utils.ts`**
`mergeFrames` used `new Int16Array([...data, ...frame.data])` in a loop (O(N^2)). Rewritten to pre-compute total size, allocate once, and copy with `TypedArray.set()` (O(N)). Also normalized `Queue.get()` check from `typeof item === 'undefined'` to `item === undefined` (no behavior change).

### Test plan
- [x] `pnpm build` passes
- [ ] `pnpm vitest run agents/src/utils.test.ts` — 31 tests pass (covers Queue and mergeFrames)
- [ ] Verify `restaurant_agent.ts` works in Agent Playground (exercises speech handles + telemetry)

### DIFF
--- a/.changeset/famous-gifts-wave.md
+++ b/.changeset/famous-gifts-wave.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+Resolve 4 Detail bugs (maxToolSteps, JWT caching, fallback concurrency, mergeFrames perf)
--- a/agents/src/llm/fallback_adapter.ts
+++ b/agents/src/llm/fallback_adapter.ts
@@ -209,6 +209,11 @@ class FallbackLLMStream extends LLMStream {
       extraKwargs: this.extraKwargs,
     });
 
+    // Suppress unhandled 'error' events from the child LLM — error detection
+    // is done via stream._runError to avoid cross-request contamination.
+    const noop = () => {};
+    llm.on('error', noop);
+
     try {
       let shouldSetCurrent = !checkRecovery;
       for await (const chunk of stream) {
@@ -217,6 +222,10 @@ class FallbackLLMStream extends LLMStream {
           this._currentStream = stream;
         }
         yield chunk;
+      }
+
+      if (stream._runError) {
+        throw stream._runError;
       }
     } catch (error) {
       if (error instanceof APIError) {
@@ -245,6 +254,8 @@ class FallbackLLMStream extends LLMStream {
         this._log.error({ llm: llm.label(), error }, 'unexpected error, switching to next LLM');
       }
       throw error;
+    } finally {
+      llm.off('error', noop);
     }
   }
 
--- a/agents/src/llm/fallback_adapter.ts
+++ b/agents/src/llm/fallback_adapter.ts
@@ -209,13 +209,6 @@ class FallbackLLMStream extends LLMStream {
       extraKwargs: this.extraKwargs,
     });
 
-    // Listen for error events - child LLMs emit errors via their LLM instance, not the stream
-    let streamError: Error | undefined;
-    const errorHandler = (ev: { error: Error }) => {
-      streamError = ev.error;
-    };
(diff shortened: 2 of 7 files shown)


u9g:
This behavior may be shocking for people moving to JS:

In [py](https://github.com/livekit/agents/blob/b4f721345b17179ad441e48d6ffafc3078117d61/livekit-agents/livekit/agents/voice/speech_handle.py#L50), we set `self._num_steps = 1` in `fun __init__` (the first step)

In [js](https://github.com/livekit/agents-js/blob/5c04e0b24111db5fd104d09031c365e3224802f4/agents/src/voice/speech_handle.ts#L78), we `stepIndex = 0` (the first step)

In [py](https://github.com/livekit/agents/blob/b4f721345b17179ad441e48d6ffafc3078117d61/livekit-agents/livekit/agents/voice/agent_activity.py#L2698-L2707), we check `speech_handle.num_steps >= self._session.options.max_tool_steps + 1`.

In [js](https://github.com/livekit/agents-js/blob/5c04e0b24111db5fd104d09031c365e3224802f4/agents/src/voice/agent_activity.ts#L2123-L2130) We check `speechHandle.numSteps >= maxToolSteps`

JS has 0-based stepIndex, while PY has 1-based. We should probably converge on the 1-based.

---

This behavior will also be shocking for people moving to JS:

In [py](https://github.com/livekit/agents/blob/b4f721345b17179ad441e48d6ffafc3078117d61/livekit-agents/livekit/agents/voice/agent_activity.py#L3344-L3358), we pass the same speech handle in again to be reused.

In [js](https://github.com/livekit/agents-js/blob/5c04e0b24111db5fd104d09031c365e3224802f4/agents/src/voice/agent_activity.ts#L2653-L2680), we recreate the speech handle on each turn, hence the changes in this PR.

WIth this too, I think we should converge on the python's way of reusing the speech handle.

This allows us to simply increment num_steps on the handle. [py](https://github.com/livekit/agents/blob/b4f721345b17179ad441e48d6ffafc3078117d61/livekit-agents/livekit/agents/voice/agent_activity.py#L3271-L3272)

u9g (changes requested):
described in my comment

PR #1215, opened 2026-04-09
feat(voice): add answering machine detection helper

chenghao-mou:
## Summary

Adds a JS/TS answering machine detection helper for voice sessions, following the Python AMD work in livekit/agents#4906 and the follow-up OTEL + authorization parity from #5376.

## Changes

- add voice.AMD to classify early call transcripts as human, machine-ivr, machine-vm, machine-unavailable, or uncertain
- pause reply authorization while AMD is running so queued speech does not play before detection completes
- resume authorization in a finally path so the session cannot remain stuck if detection throws
- add a dedicated answering_machine_detection OTEL span with AMD-specific attributes and participant/session context
- optionally force-interrupt queued speech when a machine outcome is detected
- export the new helper from voice/index.ts
- add a dedicated telephony_amd example that mirrors the Python PR flow for human, IVR, voicemail, and unavailable mailbox handling

## Testing

- pnpm test -- agents/src/voice/amd.test.ts agents/src/voice/agent_activity.test.ts
- pnpm exec tsc -p agents/tsconfig.json --noEmit
- pnpm exec eslint agents/src/voice/amd.ts agents/src/voice/amd.test.ts examples/src/telephony_amd.ts
- pnpm exec prettier --check agents/src/voice/amd.ts agents/src/voice/amd.test.ts examples/src/telephony_amd.ts

## Notes

This is the JS counterpart to the Python AMD helper, adapted to the current JS voice/session architecture, with category names aligned to the Python implementation.


agents/src/voice/amd.ts
```diff
@@ -0,0 +1,288 @@
...
+    const parsed = this.parseDetection(rawResponse);
+    return {
+      ...parsed,
+      transcript,
+      rawResponse,
+      isMachine:
+        parsed.category === AMDCategory.MACHINE_IVR ||
+        parsed.category === AMDCategory.MACHINE_VM ||
+        parsed.category === AMDCategory.MACHINE_UNAVAILABLE,
+    };
```
toubatbrian, line 424: why do we not use the tool call here like what was done in python?

agents/src/voice/amd.ts
```diff
@@ -0,0 +1,288 @@
...
+        category: this.normalizeCategory(parsed.category),
+        reason: parsed.reason?.trim() || 'No reason provided.',
+      };
+    } catch {
+      return {
+        category: AMDCategory.UNCERTAIN,
+        reason: normalized || 'Failed to parse AMD model response.',
+      };
+    }
+  }
```
toubatbrian, line 448: ^

toubatbrian:
@chenghao-mou I made some changes to your PR, mostly style improvement + aligning the timeout handling to python. Regarding the tool calling v.s generating raw JSON. Based on my testing it works 100 out of 100, so probably it's fine to leave it as it is right now.

Let me know your thoughts as well!

toubatbrian (approved):

PR #1219, opened 2026-04-09
feat(assemblyai): new STT plugin for Universal-Streaming (v3)

chasef07:
## Summary

Adds a new `@livekit/agents-plugin-assemblyai` STT plugin for AssemblyAI's Universal-Streaming v3 WebSocket API. Ported line-by-line from the Python [`livekit-plugins-assemblyai`](https://github.com/livekit/agents/tree/main/livekit-plugins/livekit-plugins-assemblyai) plugin; every non-trivial block has a `// Ref: python .../stt.py - N-M lines` comment pointing back to the source so future syncs are straightforward.

Architecture mirrors the existing Deepgram STT plugin: retry loop around a single WebSocket lifetime, concurrent `sendTask` / `listenTask` / `configTask` inside `#runWS`, graceful close via `{"type":"Terminate"}`.

## Features

- **All AssemblyAI v3 options exposed**: `speechModel` (`universal-streaming-english`, `universal-streaming-multilingual`, `u3-rt-pro`), `sampleRate`, `bufferSizeMs`, `min/maxTurnSilence`, `formatTurns`, `keytermsPrompt`, `prompt` (u3-rt-pro only), `vadThreshold`, `endOfTurnConfidenceThreshold`, `speakerLabels`, `maxSpeakers`, `domain`, custom `baseUrl` (e.g. for EU region).
- **Live config updates** via `UpdateConfiguration` messages over the active socket — no reconnect needed, unlike Deepgram's `updateOptions`. Backed by a small pending-queue + `Future` wake pattern.
- **`forceEndpoint()`** to finalize the current turn on demand.
- **`sessionId` / `expiresAt` getters** for debugging; `sessionId` is also propagated as `requestId` on `RECOGNITION_USAGE` events so metrics correlate back to a specific AssemblyAI connection.
- **Turn-event mapping**: `Turn` message → `INTERIM_TRANSCRIPT` (cumulative words), `PREFLIGHT_TRANSCRIPT` (chunk-based from `utterance`), `FINAL_TRANSCRIPT` + `END_OF_SPEECH` on `end_of_turn` (respecting `turn_is_formatted` when `formatTurns` is set).
- **Word timestamps** correctly converted from AssemblyAI's ms to `createTimedString`'s seconds.
- **`u3-pro` → `u3-rt-pro` rename** handled with a deprecation warning, matching Python.

## Known framework gap (not blocking)

`stt.SpeechData` in `@livekit/agents` doesn't yet have a `speakerId` field, so per-word speaker labels from AssemblyAI's diarization are not surfaced on emitted events. The `speakerLabels` option is exposed and sent to the server; a follow-up PR can wire through speaker ids once the base type is extended. See the TSDoc on `speakerLabels` in `plugins/assemblyai/src/stt.ts` for the exact forwarding point.

## Test plan

- [x] `pnpm --filter @livekit/agents-plugin-assemblyai... build` — clean
- [x] `pnpm --filter @livekit/agents-plugin-assemblyai lint` — clean
- [x] `pnpm -w format:check` — clean
- [x] `pnpm lint` (whole monorepo) — 24/24 green, no regressions
- [x] `pnpm build` (whole monorepo) — 24/24 green
- [x] `pnpm vitest run plugins/assemblyai --testTimeout 30000` against the real AssemblyAI API — green, transcription matches reference transcript within the 20% Levenshtein threshold
- [x] Swap into a real voice-agent example (e.g. `examples/src/basic_agent.ts`) and verify turn detection behaves naturally

## Notes for reviewers

- The shared `plugins/test/src/stt.ts` helper has a default 5000ms vitest timeout; AssemblyAI's streaming happy path lands at 5.5-6.2s end-to-end, which is occasionally flaky near the boundary. Running with `--testTimeout 30000` is reliable. This affects the shared helper, not the plugin itself — a follow-up could bump the helper's default.
- Version is pinned to `1.2.4` to match the monorepo's fixed-versioning group (`.changeset/config.json` has `"fixed": [["@livekit/agents", "@livekit/agents-plugin-*", ...]]`). Changeset is `patch` to match recent precedent — a maintainer can bump to `minor` at release time if preferred.
- `turbo.json` gains one line: `ASSEMBLYAI_API_KEY` in `globalEnv`, required so the new test file's env-var check doesn't trip `turbo/no-undeclared-env-vars`.

### DIFF
--- a/.changeset/assemblyai-stt-plugin.md
+++ b/.changeset/assemblyai-stt-plugin.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents-plugin-assemblyai": patch
+---
+
+feat(assemblyai): new STT plugin for AssemblyAI Universal-Streaming (v3) API. Ported from the Python livekit-plugins-assemblyai plugin.
--- a/plugins/assemblyai/README.md
+++ b/plugins/assemblyai/README.md
@@ -1,0 +1,18 @@
+<!--
+SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+# AssemblyAI plugin for LiveKit Agents
+
+The Agents Framework is designed for building realtime, programmable
+participants that run on servers. Use it to create conversational, multi-modal
+voice agents that can see, hear, and understand.
+
+This package contains the AssemblyAI plugin, which allows for speech recognition
+via AssemblyAI's Universal-Streaming (v3) API. Refer to the
+[documentation](https://docs.livekit.io/agents/overview/) for information on how
+to use it, or browse the [API
+reference](https://docs.livekit.io/agents-js/modules/plugins_agents_plugin_assemblyai.html).
+See the [repository](https://github.com/livekit/agents-js) for more information
+about the framework as a whole.
--- a/plugins/assemblyai/api-extractor.json
+++ b/plugins/assemblyai/api-extractor.json
@@ -1,0 +1,20 @@
+/**
+ * Config file for API Extractor.  For more info, please visit: https://api-extractor.com
+ */
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+
+  /**
+   * Optionally specifies another JSON config file that this file extends from.  This provides a way for
+   * standard settings to be shared across multiple projects.
+   *
+   * If the path starts with "./" or "../", the path is resolved relative to the folder of the file that contains
+   * the "extends" field.  Otherwise, the first path segment is interpreted as an NPM package name, and will be
+   * resolved using NodeJS require().
+   *
+   * SUPPORTED TOKENS: none
+   * DEFAULT VALUE: ""
+   */
+  "extends": "../../api-extractor-shared.json",
+  "mainEntryPointFilePath": "./dist/index.d.ts"
+}
--- a/plugins/assemblyai/package.json
+++ b/plugins/assemblyai/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@livekit/agents-plugin-assemblyai",
+  "version": "1.2.4",
+  "description": "AssemblyAI plugin for LiveKit Agents for Node.js",
+  "main": "dist/index.js",
+  "require": "dist/index.cjs",
+  "types": "dist/index.d.ts",
+  "exports": {
+    "import": {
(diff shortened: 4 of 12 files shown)


toubatbrian:
@codex review and check the plugin package format against others. Review for web-socket patterns compared with other STT plugins and check for alignment or discrepancies

chasef07:
Addressed the two review items in `2436513c`.

Changes:
- Reset `retries` after a successful `#runWS()` session so expected reconnects do not accumulate toward the fatal retry cap.
- Tightened the AssemblyAI PCM guard to require both `data.sampleRate === this.#opts.sampleRate` and `data.channels === 1` before writing into the mono `AudioByteStream`, which rejects stereo frames instead of misinterpreting them as mono.

Validation:

chasef07:
Addressed the follow-up reconnect issue in `0925809b`.

Changes:
- Added a session-scoped `AbortController` inside `#runWS()`.
- Updated `sendTask` to race on both the stream abort and the current-session abort.
- Abort the session controller in the outer `finally` block so an old `sendTask` exits immediately when that WebSocket session tears down.

This prevents an orphaned `sendTask` from surviving a reconnect and competing for `this.input` with the next session.

Validation:

chasef07:
I checked the current AssemblyAI API docs before changing this, and I do not think the new silence-unit suggestion is correct.

Both current docs describe `min_turn_silence` / `max_turn_silence` as integer values in milliseconds, not seconds:
- Universal-Streaming: `max_turn_silence` defaults to `1280` and `min_turn_silence` defaults to `400`, both described as milliseconds.
- Universal-3 Pro Streaming: `max_turn_silence` defaults to `1000` and `min_turn_silence` defaults to `100`, both described as milliseconds.

So the current JS implementation keeping these values in ms and sending them through unchanged matches the published API. Dividing by `1000` here would likely under-shoot turn detection rather than fix it.

The earlier fixes in `2436513c` and `0925809b` were still valid:
- reject non-mono PCM before writing into the mono `AudioByteStream`
- reset reconnect retries after a healthy session
- abort the old `sendTask` on WS teardown so reconnects do not leave concurrent readers on `this.input`

Docs I checked:
- https://www.assemblyai.com/docs/api-reference/streaming-api/universal-streaming/universal-streaming
- https://www.assemblyai.com/docs/api-reference/streaming-api/universal-3-pro-streaming/universal-3-pro-streaming

toubatbrian (approved):

PR #1220, opened 2026-04-09
(phonic): add logic for maintaining a single ws conn and remove say()

tinalenguyen:


agents/src/voice/agent_activity.ts
```diff
@@ -408,27 +408,39 @@ export class AgentActivity implements RecognitionHooks {
       // skip the update if the session is reused and no mid-session update is supported
       // this means the content is the same as the previous session
       const capabilities = this.llm.capabilities;
-      if (!rtReused || capabilities.midSessionInstructionsUpdate) {
+      if (!rtReused && this.realtimeSession?.realtimeModel.provider == 'phonic') {
```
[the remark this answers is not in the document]
tinalenguyen: it will be reachable soon

agents/src/voice/agent_activity.ts
```diff
@@ -408,27 +408,39 @@ export class AgentActivity implements RecognitionHooks {
...
         try {
-          await this.realtimeSession!.updateInstructions(this.agent.instructions);
+          await (this.realtimeSession as any)._updateSession(
+            this.agent.instructions,
+            this.agent.chatCtx,
+            this.tools,
+          );
```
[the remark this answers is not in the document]
tinalenguyen: it will be reachable soon

agents/src/voice/agent_activity.ts
qionghuang6: Should this be `if (this.realtimeSession?.realtimeModel.provider == 'phonic'`?
In the Python implementation, it looks like `update_session` is called whether or not `rt_reused` is true. 
Image

tinalenguyen: ah, i think it should be if rtReused is true then we call _updateSession, but it's also dependent on whether or not your _updateSession can support both the initial and reset config. in that case we can call _updateSession unconditionally

agents/src/voice/agent_activity.ts
```diff
@@ -408,27 +408,41 @@ export class AgentActivity implements RecognitionHooks {
       // skip the update if the session is reused and no mid-session update is supported
       // this means the content is the same as the previous session
       const capabilities = this.llm.capabilities;
-      if (!rtReused || capabilities.midSessionInstructionsUpdate) {
+      if (rtReused && this.realtimeSession?.realtimeModel.provider == 'phonic') {
```
toubatbrian, line 412: Any reason we’re doing a provider-specific check here?

Since `agent_activity.ts` is part of the core runtime, it feels like this should stay framework-agnostic. Wondering are there any blockers pushing this to provider layer?

tinalenguyen: On the Python side at least, we are holding off on adding `_updateSession` as a method for realtime models as this is a specialized method for Phonic. They send their instructions, chat context, and tools in one config. If more providers need this functionality then we can add that method

toubatbrian: SG!

toubatbrian (approved):

PR #1232, opened 2026-04-11
fix: prevent AgentTask inside tool from being killed by premature RunState resolution

julien-lottie:
## Problem

When an `AgentTask` is awaited inside a tool's `execute` function (e.g. a transfer-acknowledgement task inside a `transfer_to_care_home` tool), the SDK can kill the task prematurely:

```
Error: The Task finished before TransferAckTask was completed.
```

There are two bugs at play:

### Bug 1 — Premature RunState resolution kills the AgentTask

The call chain is:

1. Tool `execute` runs → calls `agentTask.run()`
2. `AgentTask.run()` calls `session._updateActivity()`, which pauses the current activity and starts the AgentTask's own activity (including `onEnter`)
3. If `onEnter` generates a reply, a **new speech handle** is created and watched by the `RunResult` (`_globalRunState._watchHandle(newHandle)`)
4. Back in `AgentTask.run()`, the parent speech handle is unwatched (since `watchedHandleCount > 1`), then **`runState._markDoneIfNeeded()` is called explicitly**
5. If the new handle's pipeline already completed (common with fast/mocked TTS), `_markDoneIfNeeded()` sees all remaining watched handles as done → **resolves the RunState**
6. The RunState resolution cascades: the parent Task finishes, its done callback fires, and since the AgentTask's future hasn't resolved yet, it force-completes with an error
7. The tool's catch block runs, but the SDK has already finalized the tool as failed — `performCallForward` (the actual Twilio transfer) is never reached

**Root cause:** The explicit `_markDoneIfNeeded()` call in `AgentTask.run()` is premature. It checks handle states at a moment when the AgentTask's new handles may have already completed their pipelines, but the parent tool execution is still in progress.

### Bug 2 — FunctionCallOutput not committed to chatCtx on interruption

In `_pipelineReplyTaskImpl`, when a speech is interrupted during tool execution:

1. `onToolExecutionStarted` adds the `FunctionCall` to `agent._chatCtx` ✅
2. `onToolExecutionCompleted` adds the `FunctionCallOutput` to `speechHandle._itemAdded` only — **not** to `agent._chatCtx` ❌
3. The interruption path returns without committing the output
4. `removeInvalidToolCalls()` later strips the orphaned `FunctionCall` (no matching output)
5. The LLM has no record that the tool was ever invoked

### Fix 1 — Remove premature `_markDoneIfNeeded()` call (`agent.ts`)

Removed the explicit `runState._markDoneIfNeeded()` call after `_updateActivity()` in `AgentTask.run()`. Handles already register their own done callbacks via `_watchHandle()` → `handle.addDoneCallback(() => this._markDoneIfNeeded(handle))`, so the RunState resolves naturally when handles truly complete. The unwatch guard (`watchedHandleCount > 1`) is preserved — it correctly prevents unwatching the last handle.

### Fix 2 — Commit completed FunctionCallOutputs on interruption (`agent_activity.ts`)

After `executeToolsTask.cancelAndWait()` in the interruption path, any completed `FunctionCallOutput` items are now committed to `agent._chatCtx` and emitted via `_toolItemsAdded()`. This mirrors the non-interrupted path (lines 2132-2137) and ensures orphaned `FunctionCall` entries are properly paired with their outputs.

## Test plan

- [x] All 627 existing tests pass (`npx vitest run` — 41 test files, 0 failures)
- [x] Build + typecheck pass (`pnpm --filter @livekit/agents build`)
- [x] Lint passes on changed files
- [ ] Manual validation with Eliza TransferAckTask (tool that calls `agentTask.run()` inside `transfer_to_care_home`)

Co-Authored-By: Ticketmaster <https://github.com/LottieHQ/ticketmaster>

### DIFF
--- a/agents/src/voice/agent.ts
+++ b/agents/src/voice/agent.ts
@@ -1,10 +1,9 @@
 // SPDX-FileCopyrightText: 2025 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-
+import type { AudioFrame } from '@livekit/rtc-node';
 import { AsyncLocalStorage } from 'node:async_hooks';
 import { ReadableStream } from 'node:stream/web';
-import type { AudioFrame } from '@livekit/rtc-node';
 import {
   LLM as InferenceLLM,
   STT as InferenceSTT,
@@ -24,7 +23,7 @@ import {
   type ToolContext,
 } from '../llm/index.js';
 import { log } from '../log.js';
-import type { SpeechEvent, STT } from '../stt/index.js';
+import type { STT, SpeechEvent } from '../stt/index.js';
 import { StreamAdapter as STTStreamAdapter } from '../stt/index.js';
 import { SentenceTokenizer as BasicSentenceTokenizer } from '../tokenize/basic/index.js';
 import type { TTS } from '../tts/index.js';
--- a/agents/src/voice/agent.ts
+++ b/agents/src/voice/agent.ts
@@ -1,9 +1,10 @@
 // SPDX-FileCopyrightText: 2025 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
-import type { AudioFrame } from '@livekit/rtc-node';
+
 import { AsyncLocalStorage } from 'node:async_hooks';
 import { ReadableStream } from 'node:stream/web';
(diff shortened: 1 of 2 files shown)


julien-lottie:
premature

PR #1244, opened 2026-04-14
fix(voice): propagate caller's OTel context through AgentSession

### LINKED ISSUE #924: feat: Add trace context propagation API for Node.js SDK (parity with Python)
## Problem

When building production agents with the Node.js SDK, I need to correlate my application's OpenTelemetry spans with LiveKit's internal spans (`agent_session`, `agent_turn`, `user_turn_completed`).

Currently, the Python SDK exposes `set_tracer_provider()` for trace context configuration, but the Node.js SDK has no equivalent API. This means:

1. LiveKit's spans and my application's spans have **separate trace IDs**
2. I cannot visualize the full call flow in a single trace tree
3. Debugging requires manually correlating spans by `room_sid` across two separate traces

## Current Behavior

- `@livekit/rtc-node` registers an OpenTelemetry TracerProvider at import time
- My observability code can add span processors to this provider (so spans export to the same backend)
- But there's no way to **share trace context** - each creates its own root span

## Desired Behavior

An API similar to Python's `set_tracer_provider()` that allows:
1. Setting a custom TracerProvider, OR
2. Propagating a parent span context so LiveKit's spans become children of my trace

## Use Case

I have an agent with an 8-phase initialization pipeline (`init:entry`, `init:connect_room`, `init:prepare_session`, etc.). I want all of these phases PLUS LiveKit's internal spans to appear in a single trace tree for end-to-end observability.

## Environment

- @livekit/agents: v1.0.x
- Node.js: v22.x
- OpenTelemetry SDK: v2.x

mrniket:
## Summary
- AgentSession now honours the active OpenTelemetry context when creating
  its `agent_session` span, so spans the caller started before `session.start()`
  become proper parents. Fixes #924.
- Added a comment to `setupCloudTracer` clarifying that `tracerProvider.register()`
  uses set-once semantics in the OTel API — it won't clobber a user's existing
  context manager or tracer provider if they've already called `NodeSDK.start()`.

## Test plan
- New unit test `agents/src/telemetry/traces.test.ts` verifying that
  `DynamicTracer.startSpan` inherits the active OTel context when no
  explicit context is passed — the contract the fix depends on.
- Manually verified on LiveKit Cloud with a downstream Langfuse integration.
- `pnpm test`, `pnpm lint`, `pnpm format:check` all pass.

### DIFF
--- a/agents/src/telemetry/traces.ts
+++ b/agents/src/telemetry/traces.ts
@@ -259,8 +259,7 @@ export async function setupCloudTracer(options: {
       resource,
       spanProcessors: [new MetadataSpanProcessor(metadata), new BatchSpanProcessor(spanExporter)],
     });
-    tracerProvider.register();
-
+    // Don't call tracerProvider.register() — it would replace the global context manager.
     setTracerProvider(tracerProvider);
 
     // Initialize standalone Pino cloud exporter (no OTEL SDK dependency)
--- a/agents/src/telemetry/traces.ts
+++ b/agents/src/telemetry/traces.ts
@@ -259,7 +259,10 @@ export async function setupCloudTracer(options: {
       resource,
       spanProcessors: [new MetadataSpanProcessor(metadata), new BatchSpanProcessor(spanExporter)],
     });
-    // Don't call tracerProvider.register() — it would replace the global context manager.
+    // register() installs an AsyncLocalStorageContextManager (needed for span nesting)
+    // and sets the global tracer provider. Both use set-once semantics in the OTel API,
+    // so if the user already called NodeSDK.start(), these are safe no-ops.
+    tracerProvider.register();
     setTracerProvider(tracerProvider);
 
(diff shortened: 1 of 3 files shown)


agents/src/telemetry/traces.ts
```diff
@@ -258,8 +258,7 @@ export async function setupCloudTracer(options: {
       resource,
       spanProcessors: [new MetadataSpanProcessor(metadata), new BatchSpanProcessor(spanExporter)],
     });
-    tracerProvider.register();
```
toubatbrian, line 262: I don't think we should delete this line; otherwise the cloud tracing will no longer work.

mrniket: Good catch — restored in 7c336dd. Added a test (389a050) covering the user-registered-provider case. Also verified locally against a real app.

theomonnom (approved):

PR #1266, opened 2026-04-16
fix(voice): allow awaiting speech handles inside function tools

u9g:
## Description

Fixes a deadlock/throw-on-valid-use in the voice pipeline: before this change, `SpeechHandle.waitForPlayout()` threw a "circular wait" error whenever any function tool was on the async stack — even when the awaited handle was a *different* one scheduled from inside the tool. That blocked the common "announce-then-work" pattern:

```ts
execute: async ({ location }, { ctx }) => {
  await ctx.session.say(`Checking ${location}...`).waitForPlayout();
  // ... do the work ...
},
```

The throw was caught defensively on purpose — but the check was too broad. Python (`livekit-agents`) narrows it to the SpeechHandle that **owns** the active tool, which correctly rejects the true circular wait while allowing nested handles. This PR ports that narrower check to JS and additionally makes `SpeechHandle` directly awaitable so `await session.generateReply()` (and `session.say()`) match Python's idiom.

## Changes Made

- `speech_handle.ts`: narrow the `waitForPlayout()` throw to the owning SpeechHandle only (port of `livekit-agents/livekit/agents/voice/speech_handle.py:156-182`).
- `agent.ts` / `generation.ts`: extend `functionCallStorage` to carry the owning `SpeechHandle` so the owner check can discriminate.
- `speech_handle.ts`: add `.then()` so `SpeechHandle` is awaitable (port of Python's `__await__`). Uses a shadow-and-restore trick on `.then` to let Promise assimilation fulfill with `this` without infinite recursion; adds a `ResolvedSpeechHandle = Omit<SpeechHandle, 'then'>` alias to keep TS's `Awaited<T>` unwrap from hitting TS1062.
- `agent_activity.ts`: add `_backgroundSpeeches` tracking for handles whose generation has finished but whose tool execution is still running; propagate `interrupt()` to them (port of Python's `_background_speeches`; closes `TODO(AJS-273)`).
- Regression tests in `speech_handle.test.ts` covering the owner check, the awaitable protocol, `.then` restoration, and a tool-context scenario that used to deadlock.

## Pre-Review Checklist

## Testing

- [x] Automated tests added/updated (if applicable) — new `speech_handle.test.ts` (9 tests) covering the 4 behaviors
- [x] All tests pass (the `speech_handle.test.ts` suite passes; other failures seen locally were unrelated env issues — missing API keys in plugin tests)

Additional manual verification: linked the local `@livekit/agents` into `agent-starter-node` and exercised a tool that calls `await ctx.session.say(...).waitForPlayout()` — previously threw, now plays the announcement, waits, and returns normally.

## Additional Notes

The Python reference for each structural change is called out inline via `// Ref: python ...` comments. The `.then` implementation comments describe the shadow-vs-delete mechanism used to avoid `.then` recursion during Promise assimilation.

Breaking changes: none at the API level. `await session.generateReply()` previously returned the handle immediately without waiting; callers that depended on that fire-and-forget behavior will now block until playout completes. Easy workaround: drop the `await`. This matches Python's behavior.

---

### DIFF
--- a/agents/src/voice/speech_handle.ts
+++ b/agents/src/voice/speech_handle.ts
@@ -152,13 +152,19 @@ export class SpeechHandle {
    * has entirely played out, including any tool calls and response follow-ups.
    */
   async waitForPlayout(): Promise<void> {
+    // Ref: python livekit-agents/livekit/agents/voice/speech_handle.py - 156-182 lines
+    // Only throw when the running function tool is owned by *this* SpeechHandle —
+    // that's the true circular wait. Waiting on a different handle scheduled from
+    // inside the tool (e.g. session.generateReply().waitForPlayout()) is safe,
+    // because the main speech-queue loop frees the owning handle's generation slot
+    // via _markGenerationDone() before awaiting tool execution.
     const store = functionCallStorage.getStore();
-    if (store && store?.functionCall) {
+    if (store?.functionCall && store.speechHandle === this) {
       throw new Error(
-        `Cannot call 'SpeechHandle.waitForPlayout()' from inside the function tool '${store.functionCall.name}'. ` +
+        `Cannot call 'SpeechHandle.waitForPlayout()' from inside the function tool '${store.functionCall.name}' that owns this SpeechHandle. ` +
           'This creates a circular wait: the speech handle is waiting for the function tool to complete, ' +
           'while the function tool is simultaneously waiting for the speech handle.\n' +
-          "To wait for the assistant's spoken response prior to running this tool, use RunContext.wait_for_playout() instead.",
+          "To wait for the assistant's spoken response prior to running this tool, use RunContext.waitForPlayout() instead.",
       );
     }
     await this.doneFut.await;
--- a/agents/src/voice/speech_handle.ts
+++ b/agents/src/voice/speech_handle.ts
@@ -5,7 +5,7 @@ import { ThrowsPromise } from '@livekit/throws-transformer/throws';
 import type { Context } from '@opentelemetry/api';
 import type { ChatItem } from '../llm/index.js';
 import type { Task } from '../utils.js';
-import { Event, Future, shortuuid } from '../utils.js';
+import { Event, Future, dedent, shortuuid } from '../utils.js';
 import { functionCallStorage } from './agent.js';
 
 /** Symbol used to identify SpeechHandle instances */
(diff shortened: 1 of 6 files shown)


agents/src/voice/speech_handle.ts
```diff
@@ -152,18 +166,66 @@ export class SpeechHandle {
...
+   * gets unwrapped, calling `.then()` again, forever). We side-step this by
+   * shadowing `.then` with `undefined` on the instance for the duration of
+   * the synchronous `Resolve(this)` call. The spec-level IsCallable check
+   * reads `undefined`, fulfills the outer promise with `this` as a plain
+   * value, and we restore the prototype method immediately after.
+   */
+  then<R1 = ResolvedSpeechHandle, R2 = never>(
+    onFulfilled?: ((value: ResolvedSpeechHandle) => R1 | PromiseLike<R1>) | null,
+    onRejected?: ((reason: unknown) => R2 | PromiseLike<R2>) | null,
+  ): Promise<R1 | R2> {
```
toubatbrian, line 218: Can we use the `ThrowsPromise` here for error type safety

agents/src/voice/speech_handle.ts
```diff
@@ -152,18 +166,66 @@ export class SpeechHandle {
...
       );
     }
     await this.doneFut.await;
   }
 
+  /**
+   * Makes the SpeechHandle awaitable: `await handle` resolves to the handle
+   * itself once its playout has finished.
+   *
+   * Ref: python livekit-agents/livekit/agents/voice/speech_handle.py - 184-189 lines
```
toubatbrian: Let's remove the `// Ref: ` comments before merging the PR

agents/src/voice/speech_handle.ts
```diff
@@ -23,6 +23,20 @@ export function isSpeechHandle(value: unknown): value is SpeechHandle {
...
+ * Structurally identical to SpeechHandle at runtime — this alias only exists
+ * to hide the `then` key from the static view. Without it, TypeScript's
+ * `Awaited<T>` unwrap recurses through `SpeechHandle`'s own `.then` callback
+ * parameter forever, emitting TS1062 ("Type is referenced directly or
+ * indirectly in the fulfillment callback of its own 'then' method").
+ * Omitting `then` terminates the unwrap because the pattern
+ * `object & { then(...) }` no longer matches. In practice, calling `.then`
+ * on an already-awaited handle has no meaningful use.
+ */
+export type ResolvedSpeechHandle = Omit<SpeechHandle, 'then'>;
```
toubatbrian, line 38: Should we expose `SpeechHandle` as `Omit<_SpeechHandle, 'then'>` and make below class to be `_SpeechHandle` instead of introducing another `ResolvedSpeechHandle`? This way the user facing API will stay consistent type of `SpeechHandle` when calling `then`

toubatbrian (commented):
Added some comments, otherwise LGTM!

agents/src/voice/speech_handle.ts
u9g: I don't think so. Then everyone that has `const x: SpeechHandle = generateReply()`'s code is broken. The way the changes in the PR are now is that the ResolvedSpeechHandle type is only introduced when `await` is used, which is also new to this PR.

TLDR: ResolvedSpeechHandle only comes from then'ing, which was also introduced in this pr. If we instead did your suggestion, we break existing type annotations on generateReply. If you think we should though it's fine, but it's a breaking change.

agents/src/voice/speech_handle.ts
u9g: done here: https://github.com/livekit/agents-js/pull/1266/changes/1d463d54ecc728d616741b6fc5fe264cb381b947

agents/src/voice/speech_handle.ts
u9g: done here, but in a different place in the code that makes more sense, this then function is more just machinery that calls out to waitForPlayout, where I added the ThrowsPromise to: https://github.com/livekit/agents-js/pull/1266/changes/1d463d54ecc728d616741b6fc5fe264cb381b947

toubatbrian (approved):
Nice work!

PR #1280, opened 2026-04-20
feat(voice): port AvatarSession base class and transcript sync asymmetric detach warning

toubatbrian:
## Summary

Automated port of livekit/agents#5499 (`feat(avatar): add AvatarSession base class, warn on sync mis-wire`) into agents-js.

cc @toubatbrian @livekit/agent-devs — please review.

> This PR was created by an automated Claude Code Routine (currently in experimentation stage). It mirrors portable core + plugin changes from the Python repo; Python-only plugin diffs (e.g. `avatario`, `avatartalk`, `bithuman`, `did`, `keyframe`, `liveavatar`, `simli`, `tavus`) that have no JS counterpart were skipped.

### 1. `voice.AvatarSession` base class (new)

New file: `agents/src/voice/avatar/avatar_session.ts` — re-exported from `voice/avatar/index.ts` so it lands on `voice.AvatarSession`.

Behavior mirrors the Python base at `livekit-agents/livekit/agents/voice/avatar/_types.py`:

- `start(agentSession, room)`:
  - Calls `getJobContext(false)` and, if present, registers `() => this.aclose()` via `jobCtx.addShutdownCallback(...)`. If no job context is active, logs a debug message telling the caller to `aclose()` manually — same message as Python.
  - Reads `agentSession.output.audio` and, when the session has already been started, logs a warning that the existing audio output will be replaced. In JS there is no `.label` on `AudioOutput`, so the warning attaches `audioOutput.constructor.name` instead of `label`.
- `aclose()`: default no-op, overridable by subclasses.

### 2. `AgentSession._started` internal getter

`AgentSession` exposes an internal `_started` getter that reads the private `started` field. This is the JS analog of Python's `agent_session._started` attribute access inside the base `AvatarSession.start`. Marked `@internal` with a `// Ref:` comment pointing at the Python source.

### 3. `TranscriptionSynchronizer` asymmetric-detach warning

In `agents/src/voice/transcription/synchronizer.ts`:

- Added `_audioAttached`, `_textAttached`, and `_warnedAsymmetricDetach` internal flags to `TranscriptionSynchronizer` (defaults `true`, `true`, `false` — matches Python).
- Added `_onAttachmentChanged({ audioAttached?, textAttached? })` which updates the attached flags and calls `this.enabled = this._audioAttached && this._textAttached`.
- `set enabled()` now resets `_warnedAsymmetricDetach = false` on the disabled → enabled transition (matches Python `set_enabled`).
- `SyncedAudioOutput` and `SyncedTextOutput` now override `onAttached()` / `onDetached()` to forward the event into `_onAttachmentChanged`. This is the JS equivalent of Python's `on_attached` / `on_detached` overrides in `_SyncedAudioOutput` / `_SyncedTextOutput`.
- In `SyncedAudioOutput.captureFrame` (when the synchronizer is disabled) and `SyncedTextOutput.captureText` (same), we now emit a one-shot warning when the synchronizer finds itself with an asymmetric (audio attached / text detached, or vice versa) state. Warning strings match Python verbatim.

### 4. Avatar plugin updates

Updated plugins to extend `voice.AvatarSession` and call `super.start(agentSession, room)` at the top of `start()`:

- `plugins/anam/src/avatar.ts`
- `plugins/bey/src/avatar.ts`
- `plugins/lemonslice/src/avatar.ts`
- `plugins/trugen/src/avatar.ts`

Each call site carries a `// Ref: python ...` comment pointing at the line in the corresponding Python plugin.

`plugins/hedra` is already deprecated (the ctor throws), so it was intentionally left unchanged. `plugins/runway` is TTS-only in agents-js and has no avatar, so it was skipped.

## Implementation Notes (language-level differences)

- **Return type of `AvatarSession.start`**: Python uses `None`, and some plugins (e.g. lemonslice) return a session id. Python's type system lets them override with `# type: ignore[override]`. TypeScript rejects widening `Promise<void>` to `Promise<string>` in an override, so the base class's return type is `Promise<unknown>` instead of `Promise<void>`. Subclasses are free to return `void` or any other value; the base implementation explicitly `return undefined`.
- **No `label` on `AudioOutput`**: The Python `AudioOutput` base has a `label` property that's surfaced in the "audio output will be replaced" warning. `AudioOutput` in agents-js does not expose `label`, so the JS warning uses `audioOutput.constructor.name` instead. Same signal, slightly different shape.
- **Attachment wiring**: Python wires `on_attached` / `on_detached` on `_SyncedAudioOutput` / `_SyncedTextOutput` via the base `io.AudioOutput` / `io.TextOutput` hooks, which are already called on `session.output.audio = ...` / `session.output.transcription = ...` assignment in `AgentOutput`. The JS `AudioOutput` / `TextOutput` base classes already have `onAttached()` / `onDetached()` with the same call-site wiring in `AgentOutput` (see `agents/src/voice/io.ts`), so the JS port simply overrides these methods in the synced subclasses — no new plumbing was required.
- **Bithuman local-mode `aclose`**: The Python PR also migrates bithuman's local-mode runtime cleanup from an inline `job_ctx.add_shutdown_callback` into `AvatarSession.aclose()`. agents-js has no bithuman plugin, so this piece is not ported.

## Tests

- `pnpm build` passes (all packages).
- `pnpm test agents/src/voice/transcription` passes (19/19 existing synchronizer tests unaffected).
- `pnpm test plugins/lemonslice` passes (2/2).
- `pnpm lint` passes (only pre-existing warnings in unrelated files).

## Test plan

- [ ] Verify `AvatarSession.start()` registers `aclose` on the job shutdown callback (anam, bey, lemonslice, trugen).
- [ ] Verify the "started after AgentSession.start()" warning fires when a developer wires the avatar after `session.start({ agent, room })`.
- [ ] Verify the asymmetric-detach warning fires exactly once per enabled cycle when `session.output.audio = null` or `session.output.transcription = null` while the other side is still attached.
- [ ] Smoke test at least one avatar plugin end-to-end against LiveKit.

https://claude.ai/code/session_01HwutyuE78oUGoYwUV69pF8

### DIFF
--- a/plugins/bey/src/avatar.test.ts
+++ b/plugins/bey/src/avatar.test.ts
@@ -1,0 +1,28 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { voice } from '@livekit/agents';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { AvatarSession } from './avatar.js';
+
+describe('Bey AvatarSession', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('calls base AvatarSession.start first', async () => {
+    const sentinel = new Error('super-start-called');
+    const superStartSpy = vi
+      .spyOn(voice.AvatarSession.prototype, 'start')
+      .mockRejectedValue(sentinel);
+
+    const avatar = new AvatarSession({
+      apiKey: 'bey-test-key',
+    });
+
+    await expect(
+      avatar.start({ _started: false, output: { audio: null } } as any, {} as any),
+    ).rejects.toThrow('super-start-called');
+    expect(superStartSpy).toHaveBeenCalledTimes(1);
+  });
+});
--- a/.changeset/avatar-session-base-port.md
+++ b/.changeset/avatar-session-base-port.md
@@ -1,0 +1,9 @@
+---
+'@livekit/agents': patch
+'@livekit/agents-plugin-anam': patch
+'@livekit/agents-plugin-bey': patch
+'@livekit/agents-plugin-lemonslice': patch
+'@livekit/agents-plugin-trugen': patch
+---
+
+Add `voice.AvatarSession` base class and port the asymmetric-detach warning from the Python `TranscriptSynchronizer`. The new base class registers `aclose` as a job shutdown callback and warns when an avatar session is started after `AgentSession.start()` has already wired an audio output. The transcript synchronizer now tracks `_audioAttached` / `_textAttached` via `onAttached` / `onDetached` and logs a one-shot warning when audio or text is detached asymmetrically (covering external avatars and manual `session.output.audio` / `.transcription` replacement). Existing avatar plugins (anam, bey, lemonslice, trugen) now inherit from `voice.AvatarSession` and call `super.start(agentSession, room)` first.
--- a/agents/src/voice/agent_session.ts
+++ b/agents/src/voice/agent_session.ts
@@ -267,6 +267,12 @@ export class AgentSession<
   /** @internal - Timestamp when the session started (milliseconds) */
   _startedAt?: number;
 
+  // Ref: python livekit-agents/livekit/agents/voice/avatar/_types.py - 62-71 lines
+  /** @internal - Whether `start()` has been called and completed. */
+  get _started(): boolean {
+    return this.started;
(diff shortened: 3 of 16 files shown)


plugins/bey/src/avatar.test.ts
theomonnom, line 1: all the tests look similar, i'm not sure if they're really testing smtgh useful

theomonnom (approved):

PR #1303, opened 2026-04-23
fix(openai): forward session.update on RealtimeSession.updateOptions

toubatbrian:
## Summary

Ports [livekit/agents#5531](https://github.com/livekit/agents/pull/5531) ("fix(openai): forward session.update on RealtimeModel.update_options") from the Python repo.

### Upstream bug

`RealtimeSession.updateOptions()` compared the incoming options against `RealtimeModel._options` and then mutated that same object before forwarding the update. Because both sides of the diff read from the same mutable object, the diff always saw "no change" and no `session.update` event was ever sent to OpenAI — so the realtime session never actually applied the new options.

The shared-state problem is additionally harmful with multiple sessions attached to the same `RealtimeModel`: one session's `updateOptions` would silently overwrite another session's options.

### Fix

Give each `RealtimeSession` its own shallow copy of `RealtimeOptions`, populated from the model's options at construction time. All reads/writes inside `RealtimeSession` now go through `this._options` instead of `this.oaiRealtimeModel._options`, so:

- `updateOptions` compares against — and mutates — per-session state, so its diff is meaningful and the `session.update` event is emitted.
- Sessions no longer stomp on each other when they share a model.

## Ported changes

- `plugins/openai/src/realtime/realtime_model.ts`
  - New private field `RealtimeSession._options: RealtimeOptions`.
  - Constructor initializes it via a shallow spread: `this._options = { ...realtimeModel._options }`.
  - Mechanically rewrote every `this.oaiRealtimeModel._options` reference inside `RealtimeSession` to `this._options` (connection headers, URL building, Azure branching, session-update construction, tool-update event, `updateOptions`, max-session-duration timeout, modalities resolution, etc.).
  - All references inside `RealtimeModel` itself are untouched — only `RealtimeSession` now reads from a local copy.
- `.changeset/openai-realtime-session-opts-copy.md` (patch bump for `@livekit/agents-plugin-openai`).

Per-line `// Ref: python …` comments are added at the two structural change sites (the new field and the constructor copy), matching the repo convention in `CLAUDE.md` for Python ports.

## Implementation nuances / notes on parity

- **Shallow copy choice.** Python uses `dataclasses.replace(realtime_model._opts)`, which produces a new dataclass instance with the same field values (shallow). In TS, `{ ...realtimeModel._options }` is the direct equivalent — we copy the top-level field references. Nested objects (`connOptions`, `turnDetection`, `inputAudioTranscription`, etc.) are still shared by reference, which matches Python's `dataclasses.replace` semantics exactly. This is deliberate: `updateOptions` replaces whole option fields rather than mutating nested ones, so shared nested references don't reintroduce the bug.
- **Scope of `updateOptions` differs from Python.** The JS `updateOptions` currently only accepts `toolChoice` (see the existing `TODO(brian): add other options here`), whereas the Python version diffs many more options (`voice`, `turn_detection`, `input_audio_transcription`, `input_audio_noise_reduction`, `speed`, `tracing`, `truncation`, `max_response_output_tokens`). We ported the **root-cause structural fix** (per-session `_options` copy), so when additional options are added to JS's `updateOptions` in the future, they will automatically inherit the correct diff semantics. Porting the additional option branches themselves is left out of this PR to keep the change minimal and 1:1 with the upstream fix's intent.
- **`RealtimeModel._options` unchanged.** The model-level options still reflect constructor inputs and are what new sessions copy from. This matches Python: `RealtimeModel._opts` is the template, each `RealtimeSession` gets an independent snapshot.
- **No behavior change at the model level.** Callers who only use one session per model get identical observable behavior plus the `session.update` fix. Callers who reuse a model across sessions get the additional correctness improvement of isolated per-session state.

## Test plan

- [x] `pnpm build:agents` — passes.
- [x] `pnpm --filter=@livekit/agents-plugin-openai build` — passes (types + tsup).
- [x] `pnpm test -- plugins/openai` — all 12 `realtime_model.test.ts` tests pass. (Unrelated `ws/llm.test.ts` failure requires `OPENAI_API_KEY` and predates this change.)
- [x] `pnpm --filter=@livekit/agents-plugin-openai lint` — no new errors/warnings introduced by this change (remaining warnings are pre-existing `no-explicit-any` / tsdoc items).
- [ ] Manual verification in Agent Playground: call `session.updateOptions({ toolChoice: ... })` mid-session and confirm a `session.update` frame is emitted.

---

🤖 This PR was generated by an automated Claude Code routine that ports merged `livekit/agents` plugin / core runtime PRs into `livekit/agents-js`. Experimental — please review the ported logic and the nuances noted above carefully.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UMabbgaksL9ErAXjqyQfaY)_

### DIFF
--- a/.changeset/openai-realtime-session-opts-copy.md
+++ b/.changeset/openai-realtime-session-opts-copy.md
@@ -1,0 +1,11 @@
+---
+"@livekit/agents-plugin-openai": patch
+---
+
+fix(openai): forward session.update on RealtimeSession.updateOptions
+
+`RealtimeSession.updateOptions()` compared against the shared `RealtimeModel._options`, but the same call mutated that shared object before forwarding to sessions. The diff always saw "no change" and no `session.update` was sent to OpenAI.
+
+Give each `RealtimeSession` its own `_options` copy so the per-session diff is independent of the model-level state and of any other sessions sharing the same model.
+
+Ports [livekit/agents#5531](https://github.com/livekit/agents/pull/5531).
--- a/plugins/openai/src/realtime/realtime_model.test.ts
+++ b/plugins/openai/src/realtime/realtime_model.test.ts
@@ -2,9 +2,9 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import { llm } from '@livekit/agents';
-import { describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import type * as api_proto from './api_proto.js';
-import { livekitItemToOpenAIItem } from './realtime_model.js';
+import { RealtimeModel, livekitItemToOpenAIItem } from './realtime_model.js';
 
 describe('livekitItemToOpenAIItem', () => {
   describe('message items', () => {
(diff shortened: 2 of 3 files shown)


toubatbrian:
Requesting review from @toubatbrian and @livekit/agent-devs.

Note: GitHub doesn't let the PR author be added as a formal "Reviewer" since @toubatbrian is the author of this automated PR — tagging here instead.

---

tinalenguyen (approved):
lgtm!

PR #1310, opened 2026-04-24
docs(voice): scaffold resume_false_interruption port (tracks livekit/agents#5535)

### LINKED ISSUE #5509: Agent transitions `thinking → speaking` after user has already transitioned to `speaking` (stale reply plays over new user turn)
### Summary

We are reliably reproducing a race on `livekit-agents ~= 1.5.2`: the agent enters `speaking` and begins TTS playout even though the `UserStateChangedEvent` for `listening → speaking` has already fired milliseconds earlier. The stale reply gets played over the user's new turn, causing both sides to talk at once and typically resulting in a bad interruption/EOT resolution.

We shipped a workaround on our side that works, but we believe the SDK should guard this internally — once the session has observed the user transitioning into `speaking`, it shouldn't let a reply that is still in `thinking` (no audio emitted yet) promote to `speaking`.

The bug is visible when turn detection is driven by both VAD and STT.

### Reproduction — real call logs

In this call the LLM generation for the previous user turn had just started when the user began a new turn. The SDK still flipped the agent to `speaking`:

```
02:43:58.700  Agent state changed: listening → thinking
02:43:59.094  VAD speech detected (triggering_probability=0.90)
02:43:59.095  User state changed: away → speaking
02:43:59.144  VAD waiting for EOS (probability=0.92, min_silence_required=1.10s)
02:43:59.264  Agent state changed: thinking → speaking   ← should not happen
02:43:59.299  ElevenLabs TTS request completed
```

Between the user going `speaking` (02:43:59.095) and the agent going `speaking` (02:43:59.264) there is ~170 ms — plenty of time for the SDK to observe the user state transition and cancel the pending generation.

A full chat history + traces + logs bundle from one of these calls is available on request — we didn't want to paste customer data in a public issue.

### Our workaround

In our `user_state_changed` handler, when the new state is `speaking` and `session.agent_state == "thinking"`, we call `session.interrupt()` to cancel the in-flight generation before it can promote to `speaking`. The committed user turn's `on_end_of_turn` path then regenerates naturally against the new transcript.

Guardrails we had to add:

- Skip when `current_speech.num_steps > 1` — otherwise we cancel the follow-up generation that runs after a tool call (e.g. agent handoff).
- Skip when `current_speech.allow_interruptions` is `False` — `SpeechHandle.interrupt(force=False)` raises `RuntimeError` for non-interruptible speeches (our initial greeting, mandatory compliance messages, etc.).
- Wrap `session.interrupt()` in `try/except RuntimeError` as defence in depth for non-interruptible speeches sitting in the scheduling queue.

### How often we see it

We added a metric (`agent.thinking_interrupted_by_user`) that increments every time our workaround fires. Across **72 completed calls** with **~51s average duration**, the metric fired **127 times** — so this is not a rare edge case, it fires multiple times per call in some conversations. Since shipping the workaround we haven't seen the bug in call reviews.

### Why we think this belongs in the SDK

- Every agent hitting this race has to reimplement the same `user_state_changed` listener with the same guardrails (tool-call follow-ups, non-interruptible speeches). That logic is easy to get wrong.
- The SDK already knows the user's state and the agent's state — it is in a much better position than application code to break the tie atomically.
- Semantically, if the user is already in `speaking` when a reply is still in `thinking` (nothing audible yet), promoting it to `speaking` is never the right call.

### Suggested fix direction

Inside the session scheduler, before allowing a speech to transition from `thinking → speaking`, check whether the user is currently in `speaking` state. If yes, cancel the pending speech (same semantics as `session.interrupt()` does today) rather than starting TTS playout. Respect `allow_interruptions=False` and the tool-call follow-up case the same way we had to externally.

If maintainers agree with this direction, happy to open a PR.

### Versions

```
livekit-agents ~= 1.5.2 (with deepgram/soniox/silero/turn-detector extras)
livekit-api >= 1.1.0, < 2
```

toubatbrian:
## Summary

Automated Claude Code port of Python PR **livekit/agents#5535** — *"fix(voice): pause output when user starts speaking during thinking"*.

> **Heads-up:** this is a minimal **scaffolding** port, not a feature-equivalent port. The full fix cannot be applied yet because the underlying `resume_false_interruption` pause/resume pipeline is **not yet ported to agents-js** (see "Implementation nuances" below).

## What the Python PR does (context for reviewers)

Fixes [livekit/agents#5509](https://github.com/livekit/agents/issues/5509). When a user starts a new turn while the agent is still `thinking` (LLM generating, TTS audio not yet flowing), the stale reply could still reach `speaking` and play over the new user turn.

Key Python changes in [agent_activity.py](https://github.com/livekit/agents/pull/5535/files#diff-a6d390dc78ef14e3):

1. **New `_PausedSpeechInfo` dataclass** — carries `handle + agent_state + timeout`. The captured `agent_state` is restored on resume (instead of always `speaking`), so a pause that began during `thinking` resumes to `thinking`.
2. **New `on_start_of_speech` pause path** — when `agent_state != "speaking"` and pause is enabled, pause the output with `timeout=0`. A brief false-positive VAD then resumes immediately on VAD EOS; `_interrupt_by_audio_activity` upgrades the timeout to the real `false_interruption_timeout` once VAD confirms active speech.
3. **New helpers** — `_update_paused_speech(handle, timeout)` and `_pause_enabled()`.
4. **`_paused_speech: SpeechHandle | None` → `_PausedSpeechInfo | None`** — all call sites updated (`_start_false_interruption_timer`, `_cancel_speech_pause`, `on_end_of_speech`, `on_interim_transcript`, `on_final_transcript`).
5. **Test harness updates** — `FakeAudioOutput(can_pause=True)` supports pause; `FakeUserSpeech` with empty `transcript` fires VAD SOS/EOS only (no STT events) to simulate sub-`min_duration` noise. Two new regression tests: `test_interrupt_before_speaking_with_pausable_audio` and `test_false_interruption_before_speaking_resumes`.

## What this JS PR ports

Because the base feature isn't in TS yet, this PR is **structural only** — no behavioral change:

- **`PausedSpeechInfo` interface** added in `agent_activity.ts` with full doc and a Python ref comment pointing to the Python dataclass.
- **`_pausedSpeech: PausedSpeechInfo | null = null` field** added on `AgentActivity` so future wiring is a field-reuse rather than an API introduction.
- **`AgentState` type import** from `./events.js` (matches the Python `AgentState` import added in the fix).
- **Python Ref markers** added at the four TS TODO sites that map to the Python fix, each pointing to the exact Python line ranges:
  - `onStartOfSpeech` → Python lines 1665–1683 (the new pause path, this PR's core change)
  - `interruptByAudioActivity` → Python lines 1615–1645 (pauseEnabled / updatePausedSpeech / AgentFalseInterruptionEvent / state transition)
  - `onEndOfSpeech` → Python lines 1707–1708 (timer start using `pausedSpeech.timeout`)
  - `onFinalTranscript` → Python lines 1800–1811 (resume-timer + cancel-speech-pause task)
- **Changeset** (`patch`) describing the scaffolding.

## Implementation nuances — why no full-feature port

The JS repo already defines the option types (`turn_config/interruption.ts`):

```ts
falseInterruptionTimeout: number;      // defaultValue 2000
resumeFalseInterruption: boolean;      // defaultValue true
```

but greps show they are **never read in `agents/src/voice/**`** — the implementation is missing:

- No `_pausedSpeech`, `_falseInterruptionTimer`, or `_cancelSpeechPauseTask` state on `AgentActivity`.
- No `_startFalseInterruptionTimer`, `_cancelSpeechPause`, `_pauseEnabled`, or `_updatePausedSpeech` methods.
- `interruptByAudioActivity` hard-interrupts via `this._currentSpeech.interrupt()` instead of pausing.
- No `AgentFalseInterruptionEvent` emitter (not defined in `events.ts`).
- `FakeAudioOutput` (JS testing mock) doesn't currently support `canPause=true`.

Context: `livekit/agents-js#843` ("OTEL logging integration & System-wise Traces") explicitly lists `resume_agent_activity` / `pause_agent_activity` as "Pause/resume not in TS" under Python-specific features.

Porting all of the above alongside PR #5535's fix in a single automated pass would materially expand the change surface (multi-hundred-line addition across `agent_activity.ts`, `events.ts`, `agent_session.ts`, fake I/O, tests) and was judged out-of-scope for an automated routine. Instead this PR:

1. Adds the **minimal structural anchors** (interface + field + refs) so a subsequent human-driven port lands as a set of focused edits rather than introducing new symbols.
2. Makes the Python↔TS mapping **discoverable by grep** for anyone picking up the follow-up work — search for `TODO(port-resume-false-interruption)` or `livekit/agents#5535` to find every site.

## Follow-up (suggested scope for the full port)

A future PR should:
- Define `AgentFalseInterruptionEvent` in `events.ts` and wire it through `AgentSession`.
- Add pause/resume plumbing on `AgentActivity`: timer handles, `pauseEnabled()`, `updatePausedSpeech()`, `startFalseInterruptionTimer()`, `cancelSpeechPause()`.
- Swap the hard-interrupt in `interruptByAudioActivity` for the pause path when `pauseEnabled()` is true.
- Implement the new `onStartOfSpeech` pause path (the fix introduced by #5535).
- Extend `FakeAudioOutput` with `canPause`/pause bookkeeping mirroring `tests/fake_io.py`.
- Port the two regression tests (`test_interrupt_before_speaking_with_pausable_audio`, `test_false_interruption_before_speaking_resumes`).

## Verification

- `pnpm --filter @livekit/agents build` — passes.
- `pnpm --filter @livekit/agents lint` — no new errors (pre-existing warnings only).
- `pnpm format:check` — clean.
- `pnpm --filter @livekit/agents exec vitest run src/voice/agent_activity.test.ts` — 8/8 passed.
- `pnpm api:check` — pre-existing api-extractor failure on main (`export * as ___ not supported`), unrelated to this change.

## Provenance

- Python source: https://github.com/livekit/agents/pull/5535 (merged 2026-04-24, created 2026-04-23 — qualifies per the post-Apr-15 2026 cutoff).
- This PR was opened automatically by a Claude Code routine cc @toubatbrian — currently in experimentation stage.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KZ27WAL3p48gvyt7gUNt77)_

### DIFF
--- a/.changeset/wet-eels-refuse.md
+++ b/.changeset/wet-eels-refuse.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+feat: Resume false interruption feature


toubatbrian:
cc @claude rebased on top of resume false interruption branch. Adjust your change accordingly

PR #1314, opened 2026-04-24
fix(tts): unsubscribe StreamAdapter listeners on close

u9g:
## Summary
- `StreamAdapter` wired anonymous `metrics_collected`/`error` handlers on the wrapped TTS in the constructor and never removed them, leaking subscriptions for the lifetime of the inner instance.
- Hoist the handlers into named fields (`#forwardMetrics`, `#forwardError`) and add a `close()` override that `.off()`s them and chains `super.close()`. The wrapped TTS is caller-owned, so `close()` does not close it.

## Test plan
- [x] `pnpm --filter @livekit/agents exec tsc --noEmit`
- [x] `pnpm test -- --run agents/src/tts`

### DIFF
--- a/agents/src/tts/stream_adapter.ts
+++ b/agents/src/tts/stream_adapter.ts
@@ -2,18 +2,27 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import { ThrowsPromise } from '@livekit/throws-transformer/throws';
+import type { TTSMetrics } from '../metrics/base.js';
 import type { SentenceStream, SentenceTokenizer } from '../tokenize/index.js';
 import type { APIConnectOptions } from '../types.js';
 import { USERDATA_TIMED_TRANSCRIPT } from '../types.js';
 import { Task } from '../utils.js';
 import { createTimedString } from '../voice/io.js';
-import type { ChunkedStream } from './tts.js';
+import type { ChunkedStream, TTSError } from './tts.js';
 import { SynthesizeStream, TTS } from './tts.js';
 
 export class StreamAdapter extends TTS {
   #tts: TTS;
   #sentenceTokenizer: SentenceTokenizer;
   label: string;
+
+  #forwardMetrics = (metrics: TTSMetrics) => {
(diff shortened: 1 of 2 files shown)


agents/src/tts/stream_adapter.ts
```diff
@@ -2,32 +2,43 @@
...
-      this.emit('error', error);
-    });
+    this.#tts.on('metrics_collected', this.#forwardMetrics);
+    this.#tts.on('error', this.#forwardError);
+  }
+
+  async close(): Promise<void> {
+    this.#tts.off('metrics_collected', this.#forwardMetrics);
+    this.#tts.off('error', this.#forwardError);
+    await super.close();
```
toubatbrian, line 41: In `agent.ts` `ttsNode` the StreamAdapter never gets closed once the turn finished. So this `close` method never gets called. 

```ts
    async ttsNode(
      agent: Agent,
      text: ReadableStream<string>,
      _modelSettings: ModelSettings,
    ): Promise<ReadableStream<AudioFrame> | null> {
      const activity = agent.getActivityOrThrow();
      if (!activity.tts) {
        throw new Error('ttsNode called but no TTS node is available');
      }

      let wrappedTts = activity.tts;

      if (!activity.tts.capabilities.streaming) {
        wrappedTts = new TTSStreamAdapter(wrappedTts, new BasicSentenceTokenizer());
      }

      ...

      let cleaned = false;
      const cleanup = () => {
        if (cleaned) return;
        cleaned = true;
        stream.close();
      };
```
Can we add `wrapedTts.close()` conditionally in `cleanup` so it properly remove the extra callback handlers?

u9g: solved.

u9g: https://github.com/livekit/agents-js/pull/1314/changes/e91715b459be99d959c77456707a30381a1d80ba

theomonnom (approved):

PR #1321, opened 2026-04-27
chore(openai): remove STT.withGroq constructor

toubatbrian:
## Summary

Port of [livekit/agents#5555](https://github.com/livekit/agents/pull/5555) — *chore(openai): remove STT.with_groq constructor*.

Groq has its own dedicated plugin (`livekit-plugins-groq` on Python; the equivalent `@livekit/agents-plugin-groq` should be the JS migration target for users). The OpenAI plugin's Groq STT shortcut is therefore being removed.

> ⚠️ This is an automated Claude Code Routine PR created by @toubatbrian. Right now it is in experimentation stage.

## Ported changes

- **Remove `STT.withGroq` static constructor** from `plugins/openai/src/stt.ts`. This mirrors the removal of `STT.with_groq` in the Python OpenAI plugin (`livekit/plugins/openai/stt.py`).
- **Drop the now-unused `GroqAudioModels` type** from `plugins/openai/src/models.ts`. This mirrors removal of the `GroqAudioModels` literal in `livekit/plugins/openai/models.py`.
- **Drop the `GroqAudioModels` import** from `stt.ts` (the only consumer).

## Implementation nuances / divergence from the Python PR

The Python PR also removed the `GroqChatModels` literal from `models.py`, which was safe in Python because `LLM.with_groq` had already been deleted in a prior change. **In `agents-js`, `LLM.withGroq` still exists in `plugins/openai/src/llm.ts` and references `GroqChatModels`.** Therefore, this port intentionally keeps `GroqChatModels` and `LLM.withGroq` in place — removing them would be an out-of-scope breaking change for JS users and is not part of upstream PR #5555.

A follow-up port (mirroring whichever earlier Python PR removed `LLM.with_groq`) will be needed to fully reach parity — when that lands, `GroqChatModels` and `LLM.withGroq` can be removed together.

The Python PR also touches `uv.lock`, which is Python-specific and intentionally not ported.

## Breaking change

Calls to `STT.withGroq({ ... })` in user code will need to be migrated to the dedicated Groq plugin.

## Test plan

- [ ] `pnpm -w build` succeeds (no remaining references to `GroqAudioModels` or `STT.withGroq`).
- [ ] `pnpm -w test` passes.
- [ ] Verify no example or doc in this repo still imports `STT.withGroq` or `GroqAudioModels`.

## Reviewers

---

🤖 Generated by Claude Code automation. Source PR: livekit/agents#5555.

---
_Generated by [Claude Code](https://claude.ai/code/session_011UeAZ2NQi4fT6XhfvB1MF3)_

### DIFF
--- a/plugins/openai/src/models.ts
+++ b/plugins/openai/src/models.ts
@@ -112,11 +112,6 @@ export type GroqChatModels =
   | 'gemma-7b-it'
   | 'gemma2-9b-it';
 
-export type GroqAudioModels =
-  | 'whisper-large-v3'
-  | 'distil-whisper-large-v3-en'
-  | 'whisper-large-v3-turbo';
-
 export type DeepSeekChatModels = 'deepseek-coder' | 'deepseek-chat';
 
 export type TogetherChatModels =
--- a/plugins/openai/src/stt.ts
+++ b/plugins/openai/src/stt.ts
@@ -4,7 +4,7 @@
 import { type AudioBuffer, mergeFrames, normalizeLanguage, stt } from '@livekit/agents';
 import type { AudioFrame } from '@livekit/rtc-node';
 import { OpenAI } from 'openai';
-import type { GroqAudioModels, WhisperModels } from './models.js';
+import type { WhisperModels } from './models.js';
 
 export interface STTOptions {
   apiKey?: string;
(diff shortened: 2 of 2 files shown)


lukasIO (approved):
lgtm, but similar question to @theomonnom on the python side about how we want to deal with the fact that this is a breaking change. 

Should we deprecate it instead for now?

theomonnom:
It was deleted on Python as well, so it lg to me

PR #1323, opened 2026-04-27
feat(metrics): port `playbackLatency` metric from python (#5524)

toubatbrian:
## Summary

Ports [livekit/agents#5524](https://github.com/livekit/agents/pull/5524) ("feat(metrics): add playback_start_latency metric") to agents-js.

Adds a new `playbackLatency` field on `MetricsReport` (and therefore on assistant `ChatMessage.metrics`) measuring the delay between the first audio frame being forwarded to the `AudioOutput` and the output reporting that playback actually started.

This metric is typically near-zero in the default pipeline (the room output self-reports `playbackStarted` when it pushes the first frame to the track and so does not account for network delivery to the client), and becomes significant when a remote avatar worker is in the chain — in that scenario the avatar reports playback via the `lk.playback_started` RPC and the latency reflects the avatar/network leg.

This is an automated port created by the Claude Code routine triggered by @toubatbrian. cc @toubatbrian @livekit/agent-devs as reviewers.

### `agents/src/llm/chat_context.ts`

- Adds `playbackLatency?: number` (seconds) to `MetricsReport`. Documented with a tsdoc block matching the Python field's docstring.

### `agents/src/voice/generation.ts`

- Adds `startedForwardingAt?: number` (ms — `Date.now()`) to the `_AudioOut` interface.
- In `forwardAudio`, sets `out.startedForwardingAt = Date.now()` the first time a frame is appended to `out.audio`. Mirrors the Python `_audio_forwarding_task` change (`out.started_forwarding_at = time.time()` at the same point in the loop).

### `agents/src/voice/agent_activity.ts`

Two replyKind paths populate `assistantMetrics`; both are updated:

1. **`ttsTask` (the `say()` path)** — `onFirstFrame` now also captures `replyStartedForwardingAt` from `audioOut.startedForwardingAt` (or falls back to `replyStartedSpeakingAt` for the text-only / no-audio case). `replyAssistantMetrics.playbackLatency = (replyStartedSpeakingAt - replyStartedForwardingAt) / 1000` is populated alongside `startedSpeakingAt` / `stoppedSpeakingAt`.

2. **`_pipelineReplyTaskImpl` (LLM → TTS pipeline)** — same treatment via `agentStartedForwardingAt` and a `playbackLatency` entry on `assistantMetrics`.

Each touched location has an inline `// Ref: python livekit-agents/livekit/agents/voice/<file>.py - <range> lines` comment as required by `CLAUDE.md`.

## Implementation nuances (where exact code-level parity is not possible)

1. **Time units.** Python tracks all timestamps as seconds-floats (`time.time()`); JS in this repo uses milliseconds (`Date.now()`) by convention (per `CLAUDE.md`). `_AudioOut.startedForwardingAt` is therefore stored in **ms**, and we divide the `(startedSpeakingAt - startedForwardingAt)` difference by `1000` when writing it into `MetricsReport.playbackLatency` (which, per the existing `startedSpeakingAt` / `stoppedSpeakingAt` / `e2eLatency` fields, is in **seconds**). End result: `playbackLatency` is in seconds, matching Python's `playback_latency`.

2. **`_on_first_frame` receives `audio_out`.** Python upgrades the realtime path's callback to take the `audio_out` via `functools.partial(_on_first_frame, audio_out=audio_out)`. JS already creates `onFirstFrame` as a closure inside the relevant scope, so the equivalent change is just a parameter (`audioOut: _AudioOut | null`) — no `partial`-equivalent needed. The text-only branch passes `null`.

3. **Realtime generation path is intentionally not touched.** In Python, `_realtime_generation_task_impl` already populates `assistant_metrics["started_speaking_at"]` etc. and the diff just adds `playback_latency` to that block. In JS, the realtime generation task (`_realtime_generation_task_impl` / `realtimeReplyTaskImpl`) does **not** currently track `started_speaking_at`/`stopped_speaking_at` on the assistant `ChatMessage` at all (the `ChatMessage`s are created without a `metrics` field — see `agent_activity.ts:2754` and `:2781`). Adding `playbackLatency` there in isolation would be misleading without the surrounding metrics. This pre-existing gap is unrelated to PR #5524 and is left as a separate follow-up.

4. **Example port skipped.** The Python diff updates `examples/avatar_agents/audio_wave/agent_worker.py` to log the metric. agents-js doesn't carry an equivalent avatar example in this repo, and the change in Python is illustrative (`@session.on("conversation_item_added")` already exists in JS, so users can observe the new field with the same pattern).

## Test plan

- [x] `pnpm build:agents` — clean
- [x] `pnpm lint --filter=@livekit/agents` — only pre-existing warnings, no new ones
- [x] `pnpm test agents/src/llm/chat_context` — 39/39 pass
- [x] `pnpm test agents/src/voice` — 136/136 pass
- [x] Changeset added (`patch` bump on `@livekit/agents`)
- [ ] Manual: run `restaurant_agent.ts` / `realtime_agent.ts` against the Agent Playground and verify `metrics.playbackLatency` appears on assistant messages — recommend a reviewer with playground access do this; my port runs in an isolated environment without LiveKit credentials.

## Reviewers

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gnho2cWKHrCUqdDYWyyiuV)_

### DIFF
--- a/agents/src/llm/chat_context.ts
+++ b/agents/src/llm/chat_context.ts
@@ -89,6 +89,17 @@ export interface MetricsReport {
   onUserTurnCompletedDelay?: number;
   llmNodeTtft?: number;
   ttsNodeTtfb?: number;
+  /**
+   * Delay (in seconds) between forwarding the first audio frame and the `AudioOutput`
+   * reporting playback started. Near-zero for the default room output (self-reported
+   * when the frame is pushed to the track, so it doesn't account for network delivery
+   * to the client); meaningful when a remote avatar worker is in the chain and reports
+   * playback via the `lk.playback_started` RPC.
+   *
+   * Assistant `ChatMessage` only.
+   */
+  // Ref: python livekit-agents/livekit/agents/llm/chat_context.py - 294-301 lines
+  playbackLatency?: number;
   e2eLatency?: number;
 }
 
--- a/agents/src/llm/chat_context.ts
+++ b/agents/src/llm/chat_context.ts
@@ -98,7 +98,6 @@ export interface MetricsReport {
    *
    * Assistant `ChatMessage` only.
    */
-  // Ref: python livekit-agents/livekit/agents/llm/chat_context.py - 294-301 lines
   playbackLatency?: number;
   e2eLatency?: number;
 }
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -1832,7 +1832,6 @@ export class AgentActivity implements RecognitionHooks {
     let replyStartedForwardingAt: number | undefined;
     let replyTtsGenData: _TTSGenerationData | null = null;
 
-    // Ref: python livekit-agents/livekit/agents/voice/agent_activity.py - 2195-2208 lines
     const onFirstFrame = (audioOut: _AudioOut | null, startedSpeakingAt?: number) => {
       replyStartedSpeakingAt = startedSpeakingAt ?? Date.now();
       replyStartedForwardingAt = audioOut?.startedForwardingAt ?? replyStartedSpeakingAt;
@@ -1915,7 +1914,6 @@ export class AgentActivity implements RecognitionHooks {
         replyAssistantMetrics.startedSpeakingAt = replyStartedSpeakingAt / 1000; // ms -> seconds
         replyAssistantMetrics.stoppedSpeakingAt = replyStoppedSpeakingAt / 1000; // ms -> seconds
 
-        // Ref: python livekit-agents/livekit/agents/voice/agent_activity.py - 2320-2323 lines
         if (replyStartedForwardingAt !== undefined) {
           replyAssistantMetrics.playbackLatency =
             (replyStartedSpeakingAt - replyStartedForwardingAt) / 1000; // ms -> seconds
@@ -2118,7 +2116,6 @@ export class AgentActivity implements RecognitionHooks {
 
     let agentStartedSpeakingAt: number | undefined;
     let agentStartedForwardingAt: number | undefined;
-    // Ref: python livekit-agents/livekit/agents/voice/agent_activity.py - 2526-2548 lines
     const onFirstFrame = (audioOutRef: _AudioOut | null, startedSpeakingAt?: number) => {
       agentStartedSpeakingAt = startedSpeakingAt ?? Date.now();
       agentStartedForwardingAt = audioOutRef?.startedForwardingAt ?? agentStartedSpeakingAt;
@@ -2199,7 +2196,6 @@ export class AgentActivity implements RecognitionHooks {
       assistantMetrics.startedSpeakingAt = agentStartedSpeakingAt / 1000; // ms -> seconds
       assistantMetrics.stoppedSpeakingAt = agentStoppedSpeakingAt / 1000; // ms -> seconds
 
-      // Ref: python livekit-agents/livekit/agents/voice/agent_activity.py - 2645-2647 lines
       if (agentStartedForwardingAt !== undefined) {
         assistantMetrics.playbackLatency =
           (agentStartedSpeakingAt - agentStartedForwardingAt) / 1000; // ms -> seconds
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -1829,10 +1829,13 @@ export class AgentActivity implements RecognitionHooks {
     }
 
     let replyStartedSpeakingAt: number | undefined;
+    let replyStartedForwardingAt: number | undefined;
     let replyTtsGenData: _TTSGenerationData | null = null;
 
-    const onFirstFrame = (startedSpeakingAt?: number) => {
+    // Ref: python livekit-agents/livekit/agents/voice/agent_activity.py - 2195-2208 lines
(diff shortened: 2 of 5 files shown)


agents/src/voice/agent_activity.ts
```diff
@@ -1910,6 +1914,12 @@ export class AgentActivity implements RecognitionHooks {
       if (replyStartedSpeakingAt !== undefined) {
         replyAssistantMetrics.startedSpeakingAt = replyStartedSpeakingAt / 1000; // ms -> seconds
         replyAssistantMetrics.stoppedSpeakingAt = replyStoppedSpeakingAt / 1000; // ms -> seconds
+
+        // Ref: python livekit-agents/livekit/agents/voice/agent_activity.py - 2320-2323 lines
```
lukasIO: not sure how useful/confusing these line numbers will become without a commit hash attached

lukasIO (approved):
lgtm

agents/src/voice/agent_activity.ts
toubatbrian: Fair point. The `// Ref: python <path> - <range> lines` format (no commit hash) is what `CLAUDE.md` currently prescribes, so I followed that convention here for consistency with the rest of the codebase. Happy to either:

1. Add the merge SHA from livekit/agents#5524 (`f910969`) to each Ref comment in this PR, or
2. Update `CLAUDE.md` separately so the convention applies to all future ports too,

whichever you'd prefer. Let me know and I'll push the change.

---

agents/src/llm/chat_context.ts
```diff
@@ -89,6 +89,17 @@ export interface MetricsReport {
...
+  /**
+   * Delay (in seconds) between forwarding the first audio frame and the `AudioOutput`
+   * reporting playback started. Near-zero for the default room output (self-reported
+   * when the frame is pushed to the track, so it doesn't account for network delivery
+   * to the client); meaningful when a remote avatar worker is in the chain and reports
+   * playback via the `lk.playback_started` RPC.
+   *
+   * Assistant `ChatMessage` only.
+   */
+  // Ref: python livekit-agents/livekit/agents/llm/chat_context.py - 294-301 lines
```
toubatbrian: remove // Ref: comments like this in this PR @claude

agents/src/voice/agent_activity.ts
toubatbrian: Done — removed all 7 `// Ref:` comments in 81de3d7. Build still passes.

---

toubatbrian: @lukasIO I asked claude to add lines when they do PR migration from python. Adding these to make sure the piece of code it generates in JS are less likely to hallucinate by having some context anchor. Those should be removed once we feel confident about PR and before it merge to main

agents/src/voice/agent_activity.ts
```diff
@@ -1829,10 +1829,12 @@ export class AgentActivity implements RecognitionHooks {
     }
 
     let replyStartedSpeakingAt: number | undefined;
+    let replyStartedForwardingAt: number | undefined;
     let replyTtsGenData: _TTSGenerationData | null = null;
 
-    const onFirstFrame = (startedSpeakingAt?: number) => {
+    const onFirstFrame = (audioOut: _AudioOut | null, startedSpeakingAt?: number) => {
       replyStartedSpeakingAt = startedSpeakingAt ?? Date.now();
+      replyStartedForwardingAt = audioOut?.startedForwardingAt ?? replyStartedSpeakingAt;
```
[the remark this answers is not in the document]
toubatbrian: Holding on this one — the fallback faithfully mirrors the upstream Python source (livekit/agents#5524 in `_pipeline_reply_task` and the `_tts_say` path):

```python
started_forwarding_at = (
    audio_out.started_forwarding_at
    if audio_out and audio_out.started_forwarding_at is not None
    else started_speaking_at
)
```

Python deliberately falls back to `started_speaking_at` for the text-only path, which produces `playback_latency = 0` there. So Devin's suggestion is a *semantic improvement* (don't emit the metric at all when there's no audio) but a *divergence* from Python parity.

I'd rather not silently flip behavior. @toubatbrian / @lukasIO — would you like the JS port to:
- **(a) Match Python** (current behavior, `playbackLatency = 0` on text-only assistant messages), or
- **(b) Diverge** per Devin's suggestion and omit the field on text-only?

Same question applies to the parallel comment at line 2121.

---

agents/src/voice/agent_activity.ts
```diff
@@ -2107,8 +2115,10 @@ export class AgentActivity implements RecognitionHooks {
     }
 
     let agentStartedSpeakingAt: number | undefined;
-    const onFirstFrame = (startedSpeakingAt?: number) => {
+    let agentStartedForwardingAt: number | undefined;
+    const onFirstFrame = (audioOutRef: _AudioOut | null, startedSpeakingAt?: number) => {
       agentStartedSpeakingAt = startedSpeakingAt ?? Date.now();
+      agentStartedForwardingAt = audioOutRef?.startedForwardingAt ?? agentStartedSpeakingAt;
```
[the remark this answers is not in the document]
toubatbrian: Same as the line 1837 thread — flagging for reviewer choice between Python parity (current) and Devin's semantic improvement. Will follow whatever @toubatbrian / @lukasIO prefer.

---

toubatbrian:
Verified locally and on observability dashboard.

PR #1324, opened 2026-04-27
feat(liveavatar): port plugin from python with video_quality param

toubatbrian:
## Summary

Automated port of livekit/agents#5552 (`(liveavatar): add video_quality param`) into agents-js, plus the **first-time port of the entire `liveavatar` plugin** because the Python diff lands on a plugin that previously had no JS counterpart (it was explicitly skipped in livekit/agents-js#1280 alongside the other Python-only avatar plugins).

cc @toubatbrian @livekit/agent-devs — please review.

> This PR was created by an automated Claude Code Routine (currently in experimentation stage). It mirrors the Python `livekit-plugins-liveavatar` plugin into a new `@livekit/agents-plugin-liveavatar` package and pulls in the `videoQuality` parameter introduced by livekit/agents#5552.

### 1. New `@livekit/agents-plugin-liveavatar` package

Layout matches existing avatar plugins (`bey`, `lemonslice`, `trugen`, `anam`):

- `plugins/liveavatar/package.json`
- `plugins/liveavatar/tsconfig.json`
- `plugins/liveavatar/tsup.config.ts`
- `plugins/liveavatar/README.md`
- `plugins/liveavatar/src/index.ts` — `Plugin` registration + re-exports
- `plugins/liveavatar/src/log.ts`
- `plugins/liveavatar/src/api.ts` — `LiveAvatarAPI` HTTP client (mirrors `livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/api.py`)
- `plugins/liveavatar/src/avatar.ts` — `AvatarSession` + queue-based audio forwarder (mirrors `livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/avatar.py`)

Each ported method/field carries an inline `// Ref: python <path> - <line-range> lines` comment pointing back to the Python source per the `CLAUDE.md` porting convention.

### 2. `videoQuality` parameter (livekit/agents#5552)

- New TS type alias `VideoQuality = 'very_high' | 'high' | 'medium' | 'low'` mirroring the Python `Literal` (api.ts).
- `AvatarSession` constructor accepts `videoQuality?: VideoQuality`; when set, `LiveAvatarAPI.createStreamingSession` includes it on the `/token` payload (omitted when unset, matching the Python `if video_quality is not None` guard).

### 3. Queue-based audio forwarding

Python uses `livekit.agents.voice.avatar.QueueAudioOutput` (a built-in core helper) plus an `AudioSegmentEnd` sentinel pushed onto the queue when `flush()` is called. agents-js doesn't expose `QueueAudioOutput` today, so the plugin defines a small private subclass of `voice.AudioOutput` that:

- Buffers `AudioFrame` items into a `streamNs.StreamChannel<AudioFrame | AudioSegmentEnd>`.
- Pushes a private `AudioSegmentEnd` sentinel value when `flush()` is called (instead of a typed-queue tag).
- Re-emits an internal `'clear_buffer'` event when `clearBuffer()` is invoked, which the `AvatarSession` driver wires into the same interrupt handling Python uses.

The forwarding loop reads from this stream, resamples to 24 kHz mono via `AudioResampler` (`@livekit/rtc-node`), buffers up to ~600 ms (first chunk) / ~1 s (subsequent) of audio, base64-encodes it, and ships an `agent.speak` event over the LiveAvatar websocket. On `AudioSegmentEnd` it flushes the buffer and emits `agent.speak_end` + `agent.start_listening`. Inbound `session.state_updated` / `agent.speak_started` / `agent.speak_ended` / `agent.speak_interrupted` events drive `onPlaybackStarted` / `onPlaybackFinished` notifications and an `agent.interrupt` send when the user barges in.

### 4. Core export: `voice.AudioOutput`

Adds `AudioOutput`, `AudioOutputCapabilities`, `PlaybackFinishedEvent`, `PlaybackStartedEvent` to the public surface of `@livekit/agents` (`agents/src/voice/index.ts`). The abstract class already existed in `agents/src/voice/io.ts` but was not exported through `voice/index.ts`, so plugin authors had no supported way to write a custom audio sink. This is the minimum change required to enable the queue-based subclass above without reaching into deep paths.

### 5. `turbo.json` env vars

Adds `LIVEAVATAR_API_KEY`, `LIVEAVATAR_API_URL`, `LIVEAVATAR_AVATAR_ID` to `globalEnv` so `eslint-plugin-turbo`'s `no-undeclared-env-vars` is satisfied (mirrors how `LEMONSLICE_*`, `BEY_*`, `TRUGEN_*` are tracked).

## Implementation Notes (language-level differences)

- **No `QueueAudioOutput` in JS core.** Python imports `from livekit.agents.voice.avatar import QueueAudioOutput, AudioSegmentEnd` — both first-party utilities. agents-js doesn't ship them, so the plugin inlines a minimal queue-backed `voice.AudioOutput` subclass + `AudioSegmentEnd` sentinel locally rather than landing them in core for a single consumer. If a second plugin ever wants the same primitive, lifting these into `agents/src/voice/avatar/` is the natural follow-up.
- **Sample-rate-only `AudioResampler` overload.** Python's `rtc.AudioResampler(input_rate=..., output_rate=..., num_channels=1)` accepts a `num_channels` kwarg. The JS binding from `@livekit/rtc-node` exposes the same parameter as the third positional arg (`new AudioResampler(inputRate, outputRate, 1)`), so the resample step is line-for-line equivalent.
- **`asyncio.Event` → `Future`.** The `_session_connected` / `_chunk_interrupted` Python `asyncio.Event` instances become a `Future<void>` (one-shot connect signal) and a simple boolean flag (the chunk-interrupt signal is read once per loop iteration), respectively. Closing semantics are unchanged.
- **`utils.aio.Chan` → `createStreamChannel`.** The websocket send queue is a `streamNs.StreamChannel<Record<string, unknown>>`. `closed` and `close()` semantics map 1-to-1; the `Close` `AgentSessionEvent` listener closes it the same way the Python `on_agent_session_close` handler does.
- **`utils.aio.interval(60).tick()` → `setTimeout` loop.** A `setTimeout`-based loop replaces Python's tickable interval. The forward path additionally resets the timer whenever a real event is sent (matches Python's `ping_interval.reset()` after a successful `ws_conn.send_json`).
- **`get_job_context()` shutdown callback.** Python's `await super().start(...)` (`AvatarSession` base) registers `aclose` on the job context's shutdown callback list. The JS base class isn't merged yet (livekit/agents-js#1280 is still open), so this plugin registers its own `jobCtx.addShutdownCallback(() => this.aclose())` directly inside `start()`. Once #1280 lands, this can be folded into a `super.start(...)` call.
- **`AccessToken` shape.** Python uses `livekit.api.AccessToken` with `with_kind('agent')`, `with_grants(VideoGrants(room_join=True, room=...))`, `with_attributes({ATTRIBUTE_PUBLISH_ON_BEHALF: ...})`. JS uses `livekit-server-sdk`'s `AccessToken` with the equivalent property accessors (`at.kind = 'agent'`, `at.addGrant(...)`, `at.attributes = ...`) — same wire format.
- **`ATTRIBUTE_PUBLISH_ON_BEHALF` is hardcoded.** `agents/src/constants.ts` exports it but `@livekit/agents`'s public barrel does not, so this plugin hardcodes the literal `'lk.publish_on_behalf'` string the same way `lemonslice`, `trugen`, and `runway` do.
- **`videoQuality` typing.** Python's `VideoQuality` lives in `avatar.py` and is imported into `api.py` via `if TYPE_CHECKING:` to avoid a circular import. TypeScript has no equivalent friction, so `VideoQuality` is defined and exported from `api.ts` and re-imported by `avatar.ts` directly.

## Tests

- `pnpm build` passes (28/28 packages including the new plugin).
- `pnpm --filter @livekit/agents-plugin-liveavatar lint` is clean.
- `pnpm format:check` is clean across the repo.
- No new unit tests added in this port (Python source has none either; lemonslice's test pattern is API-key-gated and not portable without a live LiveAvatar credential).

## Test plan

- [ ] Run the new plugin against a real LiveAvatar account end-to-end: start an `AgentSession`, attach `AvatarSession`, verify the avatar joins the room as `liveavatar-avatar-agent` and speaks the agent's audio.
- [ ] Verify each `videoQuality` value (`very_high` / `high` / `medium` / `low`) is honored on the `/token` payload and reflected in the avatar's video.
- [ ] Confirm interrupt → `agent.interrupt` is sent on `clearBuffer` while the avatar is mid-speech, and that `playbackPosition` is reported with `interrupted: true` to the agent session.
- [ ] Confirm clean shutdown: `aclose()` (or `Close` `AgentSessionEvent`) closes the message channel, drains audio, calls `/stop`, and tears down the websocket without leaking timers.
- [ ] Sandbox mode: with `isSandbox: true`, confirm the 1-minute disconnect is logged as a warning instead of raising `APIConnectionError`.
- [ ] Smoke test against an avatar using a non-24 kHz / non-mono TTS source to exercise the `AudioResampler` swap path.

https://claude.ai/code/session_01DE5pBrf3y1bFgLTK8NDTkB

---
_Generated by [Claude Code](https://claude.ai/code/session_01DE5pBrf3y1bFgLTK8NDTkB)_


plugins/liveavatar/README.md
```diff
@@ -0,0 +1,81 @@
...
+SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# LiveAvatar plugin for LiveKit Agents
+
+Support for [LiveAvatar](https://www.liveavatar.com) interactive avatars.
+
+This is the JS/TS port of the Python `livekit-plugins-liveavatar` plugin. See [https://docs.livekit.io/agents/integrations/avatar/](https://docs.livekit.io/agents/integrations/avatar/) for more information.
```
tinalenguyen:
```suggestion
This is the JS/TS port of the Python `livekit-plugins-liveavatar` plugin. See [https://docs.livekit.io/agents/models/avatar/plugins/liveavatar/](https://docs.livekit.io/agents/models/avatar/plugins/liveavatar/) for more information.
```

plugins/liveavatar/src/avatar.ts
```diff
@@ -0,0 +1,661 @@
...
+type AudioQueueItem = AudioFrame | AudioSegmentEnd;
+
+/**
+ * AudioOutput that captures agent speech frames into a queue for the LiveAvatar
+ * websocket forwarder to consume. Mirrors Python's `QueueAudioOutput` in spirit,
+ * but with a single queue + segment-end sentinel rather than a typed queue.
+ *
+ * Ref: python livekit-agents/livekit/agents/voice/avatar/_queue_io.py
+ */
+class QueueAudioOutput extends voice.AudioOutput {
```
tinalenguyen: should we move this to `agents/src/voice/avatar` as `queue_io.ts`?

.changeset/liveavatar-plugin.md
```diff
@@ -0,0 +1,10 @@
+---
+'@livekit/agents-plugin-liveavatar': minor
```
tinalenguyen:
```suggestion
'@livekit/agents-plugin-liveavatar': patch
```

plugins/liveavatar/src/avatar.ts
toubatbrian: cc @claude

.changeset/liveavatar-plugin.md
```diff
@@ -0,0 +1,10 @@
+---
+'@livekit/agents-plugin-liveavatar': patch
+'@livekit/agents': minor
```
tinalenguyen:
```suggestion
'@livekit/agents': patch
```

plugins/liveavatar/src/avatar.ts
```diff
@@ -0,0 +1,661 @@
...
+   */
+  private async mainTask(): Promise<void> {
+    if (!this.wsUrl) {
+      throw new LiveAvatarException('ws_url not set');
+    }
+    const ws = new WebSocket(this.wsUrl);
+    await new Promise<void>((resolve, reject) => {
+      ws.once('open', resolve);
+      ws.once('error', reject);
+    });
```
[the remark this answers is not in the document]
toubatbrian: Done in af8b8ed. Moved the `new WebSocket(...)` + open handshake `await` inside `mainTask`'s outer `try/catch/finally`, so a DNS / TLS / connect failure is now routed through the same cleanup path as runtime errors (close ws, release readers, drain audio buffer, log and return). Also attached a `.catch` at the spawn site (`this.mainTaskPromise = this.mainTask().catch(...)`) as a defense-in-depth net so a thrown error never escapes as an unhandled rejection.

---

plugins/liveavatar/src/api.ts
```diff
@@ -0,0 +1,202 @@
...
+   * Ref: python livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/api.py - 109-138 lines
+   */
+  private async post(
+    endpoint: string,
+    payload: Record<string, unknown>,
+    headers: Record<string, string>,
+  ): Promise<unknown> {
+    const url = this.apiUrl + endpoint;
+    const maxRetry = this.connOptions.maxRetry;
+    for (let i = 0; i < maxRetry; i++) {
```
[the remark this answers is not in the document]
toubatbrian: Done in af8b8ed. Loop now uses `i <= maxRetry` (i.e. `maxRetry + 1` total attempts), so `connOptions: { maxRetry: 0 }` correctly fires one initial request with no retries. This also matches the convention used by the other agents-js avatar plugins (lemonslice, runway). The terminal `APIConnectionError` message is now `"Failed to call LiveAvatar API after ${maxRetry + 1} attempts"` to be unambiguous about how many calls were actually made.

Note this intentionally diverges from Python's `for i in range(self._conn_options.max_retry)` (which would skip the initial attempt at `max_retry=0`) — the JS convention here is more consistent with how other plugins in the repo interpret the option.

---

plugins/liveavatar/src/avatar.ts
```diff
@@ -0,0 +1,661 @@
...
+    } catch (e) {
+      this.#logger.warn({ error: String(e) }, 'LiveAvatar main task error');
+    } finally {
+      this.closing = true;
+      closingResolver.resolve();
+
+      try {
+        if (this.sessionId && this.sessionToken) {
+          const data = await this.api.stopStreamingSession(this.sessionId, this.sessionToken);
+          if (data.code <= 200) {
```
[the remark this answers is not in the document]
toubatbrian: Keeping as-is (intentionally). This mirrors the Python plugin verbatim — `livekit-plugins/livekit-plugins-liveavatar/.../avatar.py` line 304: `if data["code"] <= 200:`. Since this is a port, I'd rather preserve identical semantics with the Python source than introduce a JS-only divergence in the success log path. The check is logging-only — the stop-session call already completed by the time we hit it, and any underlying failure is caught by the surrounding `try/catch` and surfaced via the `"Failed to stop LiveAvatar session"` warn log. I added a `// Mirrors python ... avatar.py - 304 line` comment in af8b8ed so the parity is explicit.

If we want to tighten the check, that's better tracked as a parallel change against the Python plugin first so the two stay in sync.

---

plugins/liveavatar/src/avatar.ts
toubatbrian: Done in af8b8ed. Lifted to `agents/src/voice/avatar/queue_io.ts` and exported from `voice/avatar/index.ts` so it's now reachable as `voice.QueueAudioOutput` / `voice.AudioSegmentEnd`. Layout matches Python's `livekit.agents.voice.avatar` exactly. The plugin imports them via `voice.QueueAudioOutput` and the inline classes are gone.

Also picked up the helper methods (`stream()`, `aclose()`, `notifyPlaybackStarted()`, `notifyPlaybackFinished()`, `closed` getter) into the public API so any other avatar plugin that needs a queue-backed audio sink + segment-end sentinel can reuse them without reaching into private state.

---

.changeset/liveavatar-plugin.md
toubatbrian: Applied in af8b8ed.

---

plugins/liveavatar/src/avatar.ts
```diff
@@ -0,0 +1,606 @@
...
+    this.audioPlaying = false;
+
+    if (wasPlaying && this.audioBuffer) {
+      this.audioBuffer.notifyPlaybackFinished(this.playbackPosition, true);
+      if (this.avatarSpeaking) {
+        this.sendEvent({ type: 'agent.interrupt', event_id: shortuuid() });
+      }
+      this.playbackPosition = 0;
+    }
+  }
```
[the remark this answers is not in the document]
toubatbrian: Good catch — fixed in 651e3de.

`QueueAudioOutput.clearBuffer` now emits the `'clear_buffer'` event with a `{ wasCapturing }` payload (typed as `voice.QueueAudioOutputClearEvent`) that's set synchronously from `startedSegment` *inside* the same call frame as the `clearBuffer()` invocation. The plugin's `onClearBuffer(ev)` gates `notifyPlaybackFinished(playbackPosition, true)` on `ev.wasCapturing` instead of the asynchronously-set `audioPlaying` field, so the segment-count bookkeeping in the base class always stays balanced — even when the interrupt lands in the window between `super.captureFrame()` (which increments `playbackSegmentsCount`) and the channel reader picking up the frame.

This makes the race window go away rather than papering over it: `wasCapturing` is read on the same call stack that `captureFrame` set `startedSegment = true` on, so there's no asynchronous gap to lose.

---

plugins/liveavatar/src/avatar.ts
```diff
@@ -0,0 +1,606 @@
...
+
+        const reader = this.audioBuffer.stream().getReader();
+        try {
+          while (true) {
+            const { done, value } = await reader.read();
+            if (done) break;
+            if (this.chunkInterrupted) {
+              this.chunkInterrupted = false;
+              discardChunk();
+            }
```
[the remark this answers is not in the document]
toubatbrian: Fixed in 651e3de — went a little further than just `continue;`.

The suggested `continue;` only handles the *first* stale frame. If multiple frames were already queued from the interrupted segment when `clearBuffer()` fired, the second-and-onwards would still fall through and accumulate into the next segment's chunk buffer.

The fix introduces an `interruptDraining` flag in the `forwardAudio` loop: set it whenever `chunkInterrupted` is consumed, drop every audio frame while it's true, and clear it on the next `AudioSegmentEnd` (at which point the interrupted segment has fully drained and the producer side is on a fresh segment). The `AudioSegmentEnd` itself still fires `agent.speak_end` + `agent.start_listening` so the avatar service sees a clean segment boundary even when the segment was empty.

This intentionally diverges from the Python plugin (which has the same leak); will track a parallel fix against the Python source so the two stay aligned.

---

agents/src/voice/avatar/queue_io.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  override clearBuffer(): void {
+    // Capture the in-flight state synchronously so consumers can race-free decide
+    // whether to fire `notifyPlaybackFinished` (and avoid leaking the base class's
+    // `playbackSegmentsCount > playbackFinishedCount` bookkeeping).
+    const wasCapturing = this.startedSegment;
+    this.startedSegment = false;
+    this.emit('clear_buffer', { wasCapturing } satisfies QueueAudioOutputClearEvent);
+  }
```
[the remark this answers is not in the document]
toubatbrian: Excellent catch — fixed in f518785. Both halves of your suggestion landed:

1. **`QueueAudioOutput.clearBuffer()` now writes the `AudioSegmentEnd` sentinel** when `wasCapturing` is true, so every in-flight segment gets a closing boundary on the channel regardless of whether the producer's `flush()` follows up. The subsequent `flush()` is a no-op for the sentinel write since `startedSegment` is already false. This fixes the primary failure mode where an interrupt followed by `flush()` would leave the avatar's `forwardAudio` loop stuck in `interruptDraining=true` and silently drop the entire next reply.

2. **Avatar's `onClearBuffer` now early-returns when `wasCapturing` is false**, so the timing-dependent variant you flagged in the "Also consider" section (Path 1 interrupt flow where `clearBuffer` arrives after `flush` already wrote the sentinel) can't leak `chunkInterrupted` into the next segment's first frame either.

Verified the full repro chain from your detailed trace is closed: under the agent_activity.ts:2226 interrupt path, step 5 now writes `AudioSegmentEnd` from `clearBuffer()` itself, step 6 still detects `chunkInterrupted` for the in-flight chunk_buf, and the avatar loop sees the sentinel before any of the *next* segment's frames arrive — `interruptDraining` resets, `agent.speak_end` + `agent.start_listening` fire, and the next segment is processed normally.

---

theomonnom (approved):

PR #1326, opened 2026-04-27
feat(mistral): add full node.js support for `mistralai` plugin

jeanprbt:
## Description

Rewrite the Mistral AI plugin (`plugins/mistral` -> `plugins/mistralai`) with full Node.js support, mirroring the [Python Mistral plugin](https://github.com/livekit/agents/tree/main/livekit-plugins/livekit-plugins-mistralai): LLM, STT (batch  + realtime), and TTS.                                                                                                      
- The LLM now uses the [Conversations API](https://docs.mistral.ai/studio-api/agents/agents-api#conversations) (`startStream` only), matching the Python plugin
- Add Mistral-specific provider format (`agents/src/llm/provider_format/mistralai.ts`) for converting `ChatContext` to Conversations API entries

## Changes Made

**Core framework (`agents/`)**
- Add `mistralai` provider format with `toChatCtx()` mapping system/developer messages to `instructions`, user messages to `message.input`, assistant messages to `message.output`, function calls to `function.call`, and tool outputs to `function.result`
- Register `'mistralai'` in the `ProviderFormat` union and dispatch

**Plugin (`plugins/mistralai/`)**
- **LLM**: Conversations API, retryable error handling matching the Python plugin pattern
- **STT**: Batch transcription via `audio.transcriptions.complete` and realtime streaming via `RealtimeTranscription` WebSocket with client-side VAD for endpointing
- **TTS**: Streaming PCM synthesis via `audio.speech.complete` with f32le-to-s16le conversion, voice presets and reference audio (voice cloning) support
- **Models**: Full model type definitions for chat, STT, TTS, and voice presets

## Pre-Review Checklist

## Testing

- [x] All tests pass


tinalenguyen (commented):
thank you for the PR! a few notes:
- could you add a `patch` changeset?
- it seems that this would require a package rename mistral -> mistralai on npm, @toubatbrian would this be possible?
- we have not added support for provider tools yet in JS, perhaps we can port that over before merging this PR

jeanprbt:
> could you add a patch changeset?

I added a `patch` changeset, including package renaming & STT/LLM/TTS support from Mistral.

> we have not added support for provider tools yet in JS, perhaps we can port that over before merging this PR

That would be great! However if you prefer I can remove the provider tools support from this PR now so as to make merging easier, and I'll add it later on when you will port that over.

tinalenguyen:
@jeanprbt thanks for iterating! it seems that some files are still renamed to mistralai and in the changeset, could you resolve those and the merge conflicts?

and yes, i think we should remove the provider tools file for now. i'd be happy to add them back once we support them so you wouldn't have to do anything on your end! :)

plugins/mistralai/src/llm.ts
```diff
@@ -0,0 +1,508 @@
...
+    });
+  }
+
+  /** @internal */
+  _setConversationState(
+    conversationId: string | null,
+    prevChatCtx: llm.ChatContext,
+    pendingToolCalls: Set<string>,
+  ): void {
+    this.#conversationId = conversationId;
```
tinalenguyen: in the python plugin we also reverted using the previous conversation id chaining logic to prevent interruption errors, i think we can simplify it here as well

jeanprbt: You're right, I just changed implementation to match python one, where we start new conversation at each turn instead of using previous conversation id!

jeanprbt:
> it seems that some files are still renamed to mistralai and in the changeset, could you resolve those and the merge conflicts?

I did not understand you wanted to keep `mistral` as the package name, I switched all files to `mistralai` since I'd like to match python plugin naming. Is this an issue? 

> and yes, i think we should remove the provider tools file for now. 

I removed them, you're right! Thanks 🙌

plugins/mistral/src/models.ts
```diff
@@ -1,15 +0,0 @@
-// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
-//
-// SPDX-License-Identifier: Apache-2.0
-
-export type MistralChatModels =
-  | 'mistral-large-latest'
-  | 'mistral-medium-latest'
-  | 'mistral-small-latest'
-  | 'mistral-tiny-latest'
```
toubatbrian, line 9: Let's re-export the types, but from the new package, this way we can maintain backward compatibility as well.

plugins/mistral/src/llm.ts
```diff
@@ -1,475 +0,0 @@
...
-
-export interface LLMOptions {
-  model: string | MistralChatModels;
-  apiKey?: string;
-  baseURL?: string;
-  client?: Mistral;
-  temperature?: number;
-  maxTokens?: number;
-  parallelToolCalls?: boolean;
-  toolChoice?: llm.ToolChoice;
```
toubatbrian, line 22: Let's re-export the llm, but from the new package, this way we can maintain backward compatibility as well.

pnpm-lock.yaml
```diff
@@ -278,9 +278,9 @@ importers:
       '@livekit/agents-plugin-minimax':
         specifier: workspace:*
         version: link:../plugins/minimax
-      '@livekit/agents-plugin-mistral':
+      '@livekit/agents-plugin-mistralai':
```
toubatbrian, line 278: Let's also keep both package "alive" for now, and we can gradually deprecate the old `agents-plugin-mistral` package going forward

plugins/mistralai/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@livekit/agents-plugin-mistral",
+  "name": "@livekit/agents-plugin-mistralai",
```
toubatbrian, line 2: Let's keep those **.json files in old `mistral` folder for now to maintain backward competibility

plugins/mistral/src/models.ts
jeanprbt: Done! I referenced new package's models from `mistral/src/index.ts`.

plugins/mistral/src/llm.ts
jeanprbt: Done! I referenced new package's `llm` from `mistral/src/index.ts`.

pnpm-lock.yaml
jeanprbt: I modified the `pnpm-lock.yaml` file accordingly.

plugins/mistralai/package.json
jeanprbt: Done!

plugins/mistralai/src/llm.ts
```diff
@@ -0,0 +1,381 @@
...
+    const data = ev.data;
+    const chunks: llm.ChatChunk[] = [];
+
+    if ((data as ResponseStartedEvent).type === 'conversation.response.started') {
+      return chunks;
+    }
+
+    if ((data as FunctionCallEvent).type === 'function.call.delta') {
+      const fncData = data as FunctionCallEvent;
+      const callId = fncData.toolCallId || shortuuid('tool_call_');
```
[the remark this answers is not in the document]
jeanprbt: Done!

jeanprbt:
I maintained backward compatibility by adding back former `mistral` package, with a single `index.ts` file which wraps the new package's `models.ts` & `llm.ts` files. I also added a deprecation warning & change former README to account for deprecation. 🙏

plugins/mistralai/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export * from './llm.js';
+export * from './models.js';
+export * from './stt.js';
+export * from './tts.js';
+
+class MistralPlugin extends Plugin {
+  constructor() {
+    super({
+      title: 'mistral',
```
toubatbrian: Maybe change here to `misrralai` to avoid naming collision to old package

toubatbrian (approved):
Could you also fix the linting in CI; otherwise LGTM!

plugins/mistralai/src/index.ts
jeanprbt: Of course!

jeanprbt:
Linting & package names are fixed 🙌

PR #1359, opened 2026-05-01
fix(inference): omit content from streamed tool call chunks

### LINKED ISSUE #1339: inference.LLM leaks tool-call text to TTS — Node equivalent of livekit/agents#4999 missing
### Describe the bug

When inference.LLM (or any OpenAI-compatible provider) streams a chat-completion delta that contains both
  content and tool_calls in the same chunk, the text is passed through to the TTS pipeline and the agent
  vocalizes the tool call (function name, JSON arguments, etc.). Python upstream fixed this for the Google
  plugin in livekit/agents#4999 by setting content = None when a function_call is present; that PR's
  description notes the rule already exists in the Python OpenAI plugin. The Node SDK has not shipped the
  equivalent.

  Affected code

  agents/src/inference/llm.ts → LLMStream.createRunningToolCallChunk:

  return {
    id,
    delta: {
      role: 'assistant',
      content: delta.content || undefined,   // ← passes text through alongside tool call
      extra: deltaExtra,
      toolCalls: [llm.FunctionCall.create({ ... })],
    },
  };

  Reproduction

  - @livekit/agents 1.2.x, Node 22, new inference.LLM({ model: 'openai/gpt-5.4' }), default voice pipeline
  (Deepgram STT, ElevenLabs TTS).
  - Long-running multi-turn sessions where the LLM is prompted to speak and call a tool in the same turn.
  - Intermittent — the model sometimes emits a single delta with both content: "saveAnswer({...})" and
  tool_calls: [{...}]. The text is spoken aloud.

  Expected

  When delta.tool_calls is non-empty, delta.content should not be forwarded to downstream nodes (matching
  the Python OpenAI/Google behavior).

  Suggested fix

  In createRunningToolCallChunk (and any sibling code paths that yield a ChatChunk with toolCalls
  populated), set content: undefined whenever toolCalls.length > 0. The pure-content path in parseChoice
  should remain unchanged.

  Workaround

  Override llmNode per-agent and wrap the upstream stream with a transform that strips delta.content from
  any chunk where delta.toolCalls?.length is truthy. Verified to eliminate the leak. Should be deletable
  once the upstream fix lands.

  References

  - Python fix: https://github.com/livekit/agents/pull/4999
  - Affected file (current main): agents/src/inference/llm.ts → createRunningToolCallChunk

### Relevant log output

_No response_

### Describe your environment

I experienced this on livekit cloud with the following agent versions
 npmPackages:
    @livekit/agents: ^1.2.0 => 1.2.1
    @livekit/agents-plugin-livekit: ^1.2.0 => 1.2.1
    @livekit/agents-plugin-silero: ^1.2.0 => 1.2.1
    @livekit/noise-cancellation-node: ^0.1.9 => 0.1.9

_No response_

### Additional information

_No response_

Genmin:
Fixes #1339.

## Summary

- Drop assistant `content` from streamed `ChatChunk`s that contain `toolCalls`.
- Preserve normal content-only chunks and existing tool call parsing/extra metadata behavior.
- Add a regression test for an OpenAI-compatible streaming delta that carries both text and `tool_calls` in the same choice.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm exec vitest run agents/src/inference/llm.test.ts`
- `pnpm exec prettier --check agents/src/inference/llm.ts agents/src/inference/llm.test.ts .changeset/clean-dancers-wait.md`
- `pnpm --filter @livekit/agents lint` (passes; existing repo warnings remain outside this patch)
- `git diff --check`

### DIFF
--- a/.changeset/clean-dancers-wait.md
+++ b/.changeset/clean-dancers-wait.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+fix(inference): drop streamed assistant text from tool call chunks
--- a/agents/src/inference/llm.test.ts
+++ b/agents/src/inference/llm.test.ts
@@ -11,6 +11,7 @@ beforeAll(() => {
 });
 
 type CapturedHeaders = Record<string, string>;
+type CompletionChunk = Record<string, unknown>;
 
 /**
  * Build an LLM, stub its OpenAI client's chat.completions.create, start a chat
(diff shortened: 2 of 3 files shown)


toubatbrian (approved):

ubaniabalogun:
thanks for the rapid turnaround here @Genmin! Any idea when this will be included in a release? The issue is actively affecting a usecase my team is building on Livekit

PR #1382, opened 2026-05-04
feat: add fish audio plugin

davidzhao:
also added support for autoloading .env.local in dev mode

### DIFF
--- a/.changeset/cli-envlocal.md
+++ b/.changeset/cli-envlocal.md
@@ -1,7 +1,0 @@
----
-"@livekit/agents": minor
----
-
-Auto-load `.env.local` from the current working directory when running the
-agent CLI in `dev` or `connect` mode. Existing `process.env` values take
-precedence, so shell exports always win.
--- a/.changeset/cli-envlocal.md
+++ b/.changeset/cli-envlocal.md
@@ -1,0 +1,7 @@
+---
+"@livekit/agents": minor
+---
+
+Auto-load `.env.local` from the current working directory when running the
+agent CLI in `dev` or `connect` mode. Existing `process.env` values take
+precedence, so shell exports always win.
--- a/.changeset/fishaudio-plugin.md
+++ b/.changeset/fishaudio-plugin.md
@@ -1,5 +1,5 @@
 ---
-"@livekit/agents-plugin-fishaudio": minor
+"@livekit/agents-plugin-fishaudio": patch
 ---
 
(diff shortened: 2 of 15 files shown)


agents/src/cli.ts
```diff
@@ -3,11 +3,57 @@
...
+    let value = line.slice(eq + 1).trim();
+    if (
+      (value.startsWith('"') && value.endsWith('"')) ||
+      (value.startsWith("'") && value.endsWith("'"))
+    ) {
+      value = value.slice(1, -1);
+    } else {
+      // strip trailing inline comment when value is unquoted
+      const hash = value.indexOf(' #');
+      if (hash !== -1) value = value.slice(0, hash).trimEnd();
```
toubatbrian: Should we use the library instead of building the toolkit ourselves? Below are code used by agent-starter-node
```ts
import dotenv from 'dotenv';

dotenv.config({ path: '.env.local' });
```

.changeset/cli-envlocal.md
```diff
@@ -0,0 +1,7 @@
+---
+"@livekit/agents": minor
```
toubatbrian: This should be patch

.changeset/fishaudio-plugin.md
```diff
@@ -0,0 +1,8 @@
+---
+"@livekit/agents-plugin-fishaudio": minor
```
toubatbrian: ^

agents/src/cli.ts
davidzhao: fair.. maybe the basic agent examples should just include dot env.. it was annoying enough for me to figure out how to load it in.. seems like it was simple enough to make automatic.. but we can revisit that holistically

.changeset/fishaudio-plugin.md
davidzhao: good call!

plugins/fishaudio/src/tts.ts
```diff
@@ -0,0 +1,536 @@
...
+            if (!doneFut.done) {
+              doneFut.reject(
+                new APIStatusError({
+                  message: `Fish Audio TTS request failed: ${body}`,
+                  options: { statusCode: status, body: { raw: body } },
+                }),
+              );
+            }
+          });
+          return;
```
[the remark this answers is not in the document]
toubatbrian: This is a valid suggestion

toubatbrian:
Verified locally

toubatbrian (approved):

PR #1392, opened 2026-05-05
refactor(agents): replace uuid with crypto.randomUUID

benasher44:
## Summary

- Remove the `uuid` npm dependency from `@livekit/agents`.
- Implement `shortuuid` using `randomUUID()` from `node:crypto` (same RFC 4122 string shape as before and available since node 14.17, so `.slice(0, 12)` behavior is unchanged).
- Refresh `pnpm-lock.yaml` and drop the Renovate minimum-release-age exclude entry for `uuid`.

## Verification

- `pnpm test -- agents/src/utils.test.ts`

### DIFF
--- a/.changeset/popular-coats-march.md
+++ b/.changeset/popular-coats-march.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+refactor(agents): replace uuid with crypto.randomUUID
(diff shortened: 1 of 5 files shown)


toubatbrian (approved):

toubatbrian:
@benasher44 could you fix the linting error?

benasher44:
Done!

PR #1402, opened 2026-05-06
fix(amd): avoid negative zero delay

u9g:
## Summary
- Adds AMD `speechDurationMs` and `delayMs` result fields, using an explicit zero delay when no speech end timestamp exists.
- Ports the source AMD short-greeting silence timer behavior so transcripts arriving during the human-silence window replace the pre-baked HUMAN timer without extending the machine deadline.
- Adds trace attributes and parity/regression coverage for the ported source AMD classifier tests.

## Tests
- `pnpm test -- agents/src/voice/amd.test.ts`
- `pnpm test -- agents/src/telemetry/traces.test.ts`
- `pnpm lint` (passes with existing warnings in unrelated files)

## Notes
- `pnpm --filter @livekit/agents api:check` fails before API comparison on existing API Extractor unsupported `export * as ...` syntax in `agents/dist/index.d.ts`.

### DIFF
--- a/agents/src/voice/amd.test.ts
+++ b/agents/src/voice/amd.test.ts
@@ -11,7 +11,7 @@ import type { ToolChoice, ToolContext } from '../llm/tool_context.js';
 import type { APIConnectOptions } from '../types.js';
 import type { AgentSession } from './agent_session.js';
 import { AMD, AMDCategory } from './amd.js';
-import { AgentSessionEventTypes } from './events.js';
+import { AgentSessionEventTypes, type UserState } from './events.js';
 
 class StaticLLM extends LLM {
   constructor(private readonly response: string | Error) {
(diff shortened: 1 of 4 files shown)


chenghao-mou:
This is a duplicate of https://github.com/livekit/agents-js/pull/1394

agents/src/voice/amd.test.ts
```diff
@@ -50,6 +50,44 @@ class StaticLLM extends LLM {
   }
 }
 
+class ToolCallLLM extends LLM {
+  constructor(private readonly category: AMDCategory) {
```
u9g, line 54: why does this take an AMDCategory if it's not stored or used.

agents/src/voice/amd.ts
```diff
@@ -483,10 +521,23 @@ export class AMD {
...
+
+  private delayMs(): number {
+    return this.speechEndedAt !== undefined ? Date.now() - this.speechEndedAt : 0;
+  }
+
   private setSpanAttributes(result: AMDResult): void {
     this.span?.setAttribute(traceTypes.ATTR_AMD_CATEGORY, result.category);
     this.span?.setAttribute(traceTypes.ATTR_AMD_REASON, result.reason);
     this.span?.setAttribute(traceTypes.ATTR_AMD_IS_MACHINE, result.isMachine);
+    this.span?.setAttribute(traceTypes.ATTR_AMD_SPEECH_DURATION, result.speechDurationMs);
```
u9g, line 539: why are there so many wider reaching changes that weren't part of the original diff?

agents/src/voice/amd.test.ts
u9g: it actually is, this is ts shorthand for property declaration

agents/src/voice/amd.ts
u9g: this was more like a port of other AMD behavior in a trench coat

PR #1405, opened 2026-05-06
Stream Rime TTS response chunks

chasef07:
## Summary
- stream Rime PCM response chunks from `response.body` instead of waiting for the full `response.arrayBuffer()`
- keep the existing one-frame deferral behavior so only the last emitted frame is marked `final: true`
- flush a non-empty trailing PCM remainder after the HTTP stream ends
- add a regression test that proves the Rime plugin emits audio before the mocked HTTP body closes

## Context
The Rime JS TTS plugin was buffering the whole HTTP response before writing any audio into the LiveKit audio queue. Rime can flush PCM as synthesis progresses, but `await response.arrayBuffer()` means the agent sees no audio until the provider response has completely finished. For longer utterances, app-observed TTS TTFB therefore grows with total synthesis duration instead of reflecting the provider's first available audio chunk.

This brings the JS plugin behavior in line with the intended streaming path: consume the response body reader incrementally, pass each received PCM chunk through `AudioByteStream`, and enqueue complete 100ms `AudioFrame`s as soon as they are available.

## Implementation notes
- `response.body` is checked before reading so a missing body fails explicitly.
- `ReadableStream.getReader()` is released in `finally`, and the queue is closed in the same cleanup path.
- `AudioByteStream.write()` already accepts `ArrayBufferView` and honors `byteOffset` / `byteLength`, so the streamed `Uint8Array` chunks can be passed directly without copying into a sliced `ArrayBuffer`.
- `AudioByteStream.flush()` is still called after EOF to drain a valid partial-frame remainder; empty flush frames are skipped.

## Validation
- `pnpm test -- plugins/rime/src/tts.test.ts`
- `pnpm --filter @livekit/agents-plugin-rime lint`
- `pnpm format:check`
- `pnpm turbo run build --filter=@livekit/agents-plugin-rime`
- `git diff --check`

### DIFF
--- a/.changeset/rime-streaming-response.md
+++ b/.changeset/rime-streaming-response.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-rime': patch
+---
+
+Stream Rime TTS response chunks as they arrive instead of waiting for the full response body.
--- a/plugins/rime/src/tts.test.ts
+++ b/plugins/rime/src/tts.test.ts
@@ -1,12 +1,87 @@
 // SPDX-FileCopyrightText: 2024 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
+import { initializeLogger } from '@livekit/agents';
 import { STT } from '@livekit/agents-plugin-openai';
 import { tts } from '@livekit/agents-plugins-test';
-import { describe, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import { TTS } from './tts.js';
 
+initializeLogger({ pretty: false, level: 'silent' });
+
 const hasRimeConfig = Boolean(process.env.RIME_API_KEY && process.env.OPENAI_API_KEY);
+
+function pcmChunk(byteLength: number): Uint8Array {
+  const chunk = new Uint8Array(byteLength);
+  for (let i = 0; i < chunk.length; i += 2) {
+    chunk[i] = 1;
+  }
+  return chunk;
+}
+
+async function withTimeout<T>(promise: Promise<T>, ms: number): Promise<T | 'timeout'> {
+  let timer: ReturnType<typeof setTimeout> | undefined;
+  const timeout = new Promise<'timeout'>((resolve) => {
+    timer = setTimeout(() => resolve('timeout'), ms);
+  });
+
+  return Promise.race([promise, timeout]).finally(() => {
+    if (timer) clearTimeout(timer);
+  });
+}
+
+describe('Rime TTS streaming', () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('emits audio before the Rime response body closes', async () => {
+    let bodyController!: ReadableStreamDefaultController<Uint8Array>;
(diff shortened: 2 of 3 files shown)


adrian-cowham (commented):
## PR Review Summary

**Reviewed by:** code-reviewer, silent-failure-hunter, pr-test-analyzer, code-simplifier

### Critical Issues
None.

### Important Issues
None.

### Suggestions

**[pr-test-analyzer] Cross-chunk frame accumulation not exercised** (severity 7/10) — `plugins/rime/src/tts.test.ts`
The test sends two chunks of exactly `bytesPerFrame` (3200 bytes each), so each independently yields one frame. In production, HTTP chunk boundaries are arbitrary and will rarely align. A test with misaligned chunks (e.g., `pcmChunk(1600)`, `pcmChunk(1600)`, `pcmChunk(3200)`) would exercise byte accumulation across reads — the real behavioral change this PR introduces.

**[pr-test-analyzer] `audioByteStream.flush()` trailing-frame path not exercised** (severity 6/10) — `plugins/rime/src/tts.test.ts`
With aligned chunks the internal buffer is empty at end-of-stream, so `flush()` returns a zero-sample frame that gets skipped. Sending `pcmChunk(3200 + 1600)` followed by `bodyController.close()` would force the flush path to emit a trailing partial frame — a common scenario when Rime's server sends a final unaligned chunk.

**[pr-test-analyzer] Reader release on mid-stream error not tested** (severity 5/10) — `plugins/rime/src/tts.test.ts`
A test calling `bodyController.error(new Error('network failure'))` after enqueuing one chunk would verify that the `finally` cleanup (reader lock release + queue close) fires correctly on network errors mid-stream.

**[code-simplifier] `pcmChunk` helper fill loop is unnecessary** (minor) — `plugins/rime/src/tts.test.ts`
The even-byte fill produces non-silent PCM but no assertion inspects sample content. `return new Uint8Array(byteLength)` (silence) exercises the same code paths.

### Strengths
- **Correct streaming conversion** — the `lastFrame` deferred-emit pattern preserves `final: false` / `final: true` / `done: true` sequencing identically to the pre-PR behavior
- **Solid resource cleanup** — `reader.releaseLock()` + `queue.close()` in `finally` ensures cleanup on both success and error paths; the intentional lack of a `catch` block lets errors propagate to the base class retry/emit machinery
- **Proper null-body guard** — `!response.body` early-throw is consistent with baseten, inworld, and minimax plugins
- **Well-designed regression test** — uses a real `ReadableStream` with manual controller (not over-mocked), validates audio arrives before body closes via `withTimeout`, and runs unconditionally in CI without external credentials
- **Consistent with codebase patterns** — the streaming loop mirrors the existing baseten TTS plugin almost exactly
- **`flush()` zero-sample guard is load-bearing** — `AudioByteStream.flush()` does not short-circuit when the buffer is empty, so the `samplesPerChannel === 0` filter is necessary

Looks good overall. The suggestions above are optional hardening for test coverage — none are blockers.

🤖 Reviewed with [Claude Code](https://claude.ai/code)

toubatbrian (approved):
Verified locally, LG!

PR #1429, opened 2026-05-08
Support OpenAI Realtime Whisper STT

toubatbrian:
## Summary

This PR adds realtime OpenAI STT support for `gpt-realtime-whisper` in `openai.STT`.

`openai.STT` now defaults to `gpt-realtime-whisper` with `useRealtime: true`. Because this model does not support OpenAI server-side `turn_detection`, callers must provide a VAD instance so the plugin can commit the audio buffer at end-of-speech.

## Usage

```ts
const vad = await silero.VAD.load();

  vad,
  stt: new openai.STT({
    model: 'gpt-realtime-whisper',
    vad,
  }),
  // llm, tts, ...
});
```

If `turnDetection` is provided with `gpt-realtime-whisper`, the plugin warns and ignores it by normalizing to `null`.

## Notes

- `gpt-realtime-whisper` uses the OpenAI Realtime transcription WebSocket path.
- WebSocket server events are parsed with Zod for type safety.
- `useRealtime: false` still uses the previous batch `whisper-1` path.
- Tests cover realtime defaults, batch opt-out, required VAD behavior, and model-specific `turnDetection` handling.


plugins/openai/src/stt.ts
```diff
@@ -1,10 +1,167 @@
...
 import type { AudioFrame } from '@livekit/rtc-node';
 import { OpenAI } from 'openai';
+import { type MessageEvent, WebSocket } from 'ws';
+import { z } from 'zod';
 import type { GroqAudioModels, WhisperModels } from './models.js';
+import type * as api_proto from './realtime/api_proto.js';
+
+const REALTIME_SAMPLE_RATE = 24000;
+const REALTIME_NUM_CHANNELS = 1;
+const DEFAULT_REALTIME_MODEL = 'gpt-realtime-whisper';
```
tinalenguyen, line 26: should we match python's default of `gpt-4o-mini-transcribe` or standardize across both?

toubatbrian: Maybe let's update the default on both side since 'gpt-realtime-whisper' just came out recently and should be better than previous model?

toubatbrian: One thing to note is that 'gpt-realtime-whisper' does not support server-side VAD like previous model does. So I have to add a VAD option just like the way we did to mistral STT. Happy to make the same PR on python side once we merged this.

tinalenguyen: oh i didn't know it was new, yeah we should make that the default on both sides if it's better, sounds good!

tinalenguyen (approved):
tested it and it lgtm! one note is that in the mistral stt plugin, if a vad is not passed then one will be created by default, i think we could do that too. wdyt?

toubatbrian:
> tested it and it lgtm! one note is that in the mistral stt plugin, if a vad is not passed then one will be created by default, i think we could do that too. wdyt?

Sounds good! I'll merge this and create a follow up PR, just to make things separate and cleaner

PR #1467, opened 2026-05-12
fix(openai): include ?model= on Responses-API and realtime STT WebSocket URLs

enriqueespaillat-gyde:
## Summary

Two of the OpenAI plugin's three WebSocket endpoints open the connection without a `?model=` query parameter:

- **Responses-API LLM** (`plugins/openai/src/ws/llm.ts`) — opens `wss://.../responses` with no model in the URL.
- **Realtime STT** (`plugins/openai/src/stt.ts`) — opens `wss://.../realtime?intent=transcription` with no model in the URL.

The third endpoint — the conversational realtime API in `plugins/openai/src/realtime/realtime_model.ts` — [already includes the model](https://github.com/livekit/agents-js/blob/main/plugins/openai/src/realtime/realtime_model.ts#L391):

```ts
// Standard OpenAI: /realtime?model=<model>
queryParams['model'] = model;
```

This PR extends the same convention to the other two endpoints.

## Why

OpenAI-compatible proxies route WebSocket connections at the HTTP upgrade — they don't (and generally can't) wait for and parse the first WS frame to discover which model the client wants. LiteLLM is the most concrete example: its [realtime routing docs](https://docs.litellm.ai/docs/realtime) require `?model=` on the upgrade URL (e.g., `ws://0.0.0.0:4000/v1/realtime?model=openai-gpt-4o-realtime-audio`) and reject the connection without it. Other OpenAI-compatible gateways (Cloudflare AI Gateway, Helicone, Portkey) have the same shape of constraint when they expose realtime endpoints.

OpenAI's own native endpoints accept and ignore the parameter — which is what makes the existing `realtime_model.ts` behavior safe today, and what makes this change a no-op for users hitting `api.openai.com` directly.

## Changes

- `ws/llm.ts`: extract `buildResponsesWsUrl(baseURL, model)`, set `?model=` on the URL.
- `stt.ts`: extract `buildRealtimeSttUrl(baseURL, model)`, set `?model=` on the URL.
- Unit tests for both URL builders covering OpenAI-direct, custom baseURL, trailing-slash and `/v1`-suffix handling.

Both helpers are `@internal` in JSDoc; they're exported only so the URL construction can be unit-tested without instantiating the full LLM/STT class. Happy to keep them module-private (matching `processBaseURL` in `realtime_model.ts`) if preferred — let me know.

## Test plan

- [x] `pnpm test plugins/openai/src/ws/url.test.ts plugins/openai/src/stt.test.ts` — 12 passed, 1 skipped (the pre-existing integration test that requires `OPENAI_API_KEY`)
- [x] `pnpm --filter @livekit/agents-plugin-openai build` — clean
- [x] `pnpm --filter @livekit/agents-plugin-openai lint` — no new warnings
- [x] `pnpm format:write` — no changes outside touched files

### DIFF
--- a/.changeset/openai-ws-url-model-param.md
+++ b/.changeset/openai-ws-url-model-param.md
@@ -1,0 +1,14 @@
+---
+'@livekit/agents-plugin-openai': patch
+---
+
+fix(openai): include `?model=` on Responses-API and realtime STT WebSocket URLs
+
+OpenAI-compatible gateways (LiteLLM, Cloudflare AI Gateway, Helicone,
+Portkey, etc.) can only see the URL at the WebSocket upgrade — they
+cannot read the subsequent `session.update` / `response.create` frame to
+determine which backend to dial. The `realtime/realtime_model.ts`
+conversational endpoint already includes the model in the upgrade URL;
+this aligns the Responses-API LLM and realtime STT paths with the same
+convention. OpenAI's native endpoints accept and ignore the parameter,
+so this is a no-op for direct connections.
--- a/plugins/openai/src/stt.test.ts
+++ b/plugins/openai/src/stt.test.ts
@@ -6,7 +6,7 @@ import { VAD } from '@livekit/agents-plugin-silero';
 import { stt } from '@livekit/agents-plugins-test';
 import { describe, expect, it, vi } from 'vitest';
 import type { SpeechStream } from './stt.js';
-import { STT } from './stt.js';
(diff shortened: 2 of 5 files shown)


enriqueespaillat-gyde:
@toubatbrian Any chance this can be included? We have a patch for this but wondering if this is potentially useful for others + would like to make sure we don't divert.

toubatbrian (approved):

toubatbrian:
Thanks for bumping on this, the PR looks good to me

PR #1505, opened 2026-05-14
refactor(elevenlabs): abort recv channel on ws close/error

u9g:
## Description
Replace the side-channel `Future<Error>` in the TTS recv loop with a `messageChannel.abort(err)` call from the ws event handlers, so the pending `reader.read()` rejects inline. Mirrors Python's `_recv_loop`, which raises directly from `await ws.receive()`.

## Changes Made
- `onClose`/`onError` now call `messageChannel.abort(err)` instead of resolving a separate `Future`
- Drop the post-loop `if (errorFuture.done) throw` check — the outer `try/catch` handles it
- Drop the now-unused `Future` import

## Pre-Review Checklist

## Testing
- [x] All elevenlabs tests pass (one pre-existing unrelated failure in shared TTS scaffold around `openai/STT` VAD)

### DIFF
--- a/plugins/elevenlabs/src/tts.ts
+++ b/plugins/elevenlabs/src/tts.ts
@@ -8,7 +8,6 @@ import {
   APIStatusError,
   APITimeoutError,
   AudioByteStream,
-  Future,
   type TimedString,
   asError,
   createTimedString,
@@ -431,7 +430,6 @@ class Connection {
   async #recvLoop(): Promise<void> {
     try {
       const messageChannel = stream.createStreamChannel<Record<string, unknown>>();
-      const errorFuture = new Future<Error>();
 
       const onMessage = (rawData: Buffer) => {
         try {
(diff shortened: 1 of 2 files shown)


plugins/elevenlabs/src/tts.ts
```diff
@@ -596,11 +594,6 @@ class Connection {
             }
           }
         }
-
-        // Throw any error that occurred
-        if (errorFuture.done) {
-          await errorFuture.await;
-        }
       } finally {
```
toubatbrian: I wonder if we doing in this approach then reader.read() could potentially throw an error. We would need to add a `catch` in the inner try catch as well.

plugins/elevenlabs/src/tts.ts
toubatbrian, line 596: Since we also handle error here, maybe let's flatten nested try catch into just one?

u9g: done

plugins/elevenlabs/src/tts.ts
u9g: we no longer have an inner try catch after latest commit

toubatbrian (approved):

PR #1511, opened 2026-05-15
feat: livekit-agents download-files script to optimize Docker layers

davidzhao:
enables plugin dependencies to be installed before app code is copied over

for example:

```dockerfile
COPY package.json pnpm-lock.yaml ./
RUN pnpm install --frozen-lockfile
RUN npx livekit-agents download-files
COPY src/ ./src
RUN pnpm build
```


agents/src/cli.ts
```diff
@@ -218,6 +203,12 @@ export const runApp = (opts: ServerOptions) => {
       initializeLogger({ pretty: true, level: commandOptions.logLevel });
       const logger = log();
 
+      logger.warn(
+        '`download-files` via cli.runApp() is deprecated. ' +
+          'Use `npx livekit-agents download-files` instead — it discovers installed ' +
```
toubatbrian, line 208: `npx livekit-agents download-files` seems quite verbose, maybe `npx lk download-files`?

toubatbrian (approved):

agents/src/cli.ts
davidzhao: that would confuse with our cli lk imo

examples/src/Dockerfile-example
```diff
@@ -42,26 +55,6 @@ COPY . .
...
-    --disabled-password \
-    --gecos "" \
-    --home "/app" \
-    --shell "/sbin/nologin" \
-    --uid "${UID}" \
-    appuser
-
-# Set proper permissions
-RUN chown -R appuser:appuser /app
-USER appuser
```
davidzhao, line 58: we cannot remove this. I'll open a new PR

PR #1515, opened 2026-05-15
fix(stt): reflect active child in FallbackAdapter model/provider

julien-lottie:
## Description

The STT `FallbackAdapter` exposes static `model` (`'FallbackAdapter'`) and `provider` (`'livekit'`) getters. But `audio_recognition.refreshUserTurnSttAttributes` reads `this.sttInstance.model` / `this.sttInstance.provider` on every STT event to stamp the OTel `gen_ai.request.model` and `gen_ai.provider.name` attributes on the `user_turn` span. Its own docstring describes the intended behaviour:

> Set/refresh the STT model and provider attributes on `user_turn`. Called from `ensureUserTurnSpan` and again on each STT event so providers that switch mid-turn (e.g. `FallbackAdapter`) update the span instead of being frozen at construction-time labels.

Because the getters are static, every `user_turn` span reports `"FallbackAdapter"` / `"livekit"` regardless of which child actually transcribed. A mid-turn fallover is invisible in traces, and the per-provider performance/error comparison that motivates running a `FallbackAdapter` in the first place is lost.

This PR makes the getters reflect the elected child.

## Changes Made

- Add a private `_activeStt: STT | undefined` field on `FallbackAdapter`.
- Override `model` and `provider` to return `this._activeStt?.{model,provider} ?? 'FallbackAdapter'|'livekit'`.
- Expose `_setActiveStt(stt: STT)` (`@internal`) so the co-located `FallbackSpeechStream` can record the elected child without the field becoming public.
- Set `_activeStt` from both paths:
  - `_recognize()`: immediately after a successful `stt.recognize()` returns.
  - Streaming: inside the inner `for await (const ev of child)` loop in `FallbackSpeechStream.run()`, matching the cadence at which `refreshUserTurnSttAttributes` is called.

## Pre-Review Checklist

## Testing

- [x] Automated tests added/updated: 4 new cases in a new `FallbackAdapter dynamic model/provider getters` describe block in `agents/src/stt/fallback_adapter.test.ts` covering (1) wrapper defaults before any STT is active, (2) post-`recognize()` reflection, (3) fallback child reflection after recognize fall-through, and (4) the streaming path. They use a small test-local `IdentifiedFakeSTT` subclass of the existing `FakeSTT` harness that overrides `model` / `provider` (the harness doesn't yet expose these as constructor opts).
- [x] All tests pass: `pnpm --filter '@livekit/agents' exec vitest run` → 822 passed, 0 failed in the agents package (full repo-level `pnpm test` has unrelated failures in `examples/` that require real `OPENAI_API_KEY` etc.)
- [x] `pnpm format:check` and `pnpm --filter '@livekit/agents' lint` both clean on the changed files
- [x] `pnpm --filter '@livekit/agents' exec tsc --noEmit` clean

## Additional Notes

Includes a changeset entry (`@livekit/agents`: patch).

The `_activeStt` field is documented as not safe to share an adapter across concurrent sessions/streams — the same caveat already implicitly applies to the existing `_status` array, and the typical pattern is one adapter per session.

This PR is independent of the upstream STT `FallbackAdapter` shutdown-race fix that I understand is also planned for upstream (the `forwardOrShutdown` / `isShuttingDown` work). Both can land in either order.

### DIFF
--- a/.changeset/fallback-adapter-active-child-attribution.md
+++ b/.changeset/fallback-adapter-active-child-attribution.md
@@ -1,0 +1,13 @@
+---
+'@livekit/agents': patch
+---
+
+fix(stt): reflect active child in `FallbackAdapter` `model`/`provider`
+
+`audio_recognition.refreshUserTurnSttAttributes` reads these on every
+STT event to stamp `gen_ai.request.model` / `gen_ai.provider.name`
+on the `user_turn` span. With static wrapper labels, every span
+reported `FallbackAdapter` / `livekit` regardless of which provider
+actually transcribed — so a mid-turn fallover was invisible in
+traces. Track the elected child from both the streaming and
+recognize paths and surface its identifiers.
--- a/agents/src/stt/fallback_adapter.test.ts
+++ b/agents/src/stt/fallback_adapter.test.ts
@@ -333,3 +333,94 @@ describe('FallbackSpeechStream (streaming path)', () => {
     expect(adapter.status[1]?.available).toBe(true);
   });
 });
+
+describe('FallbackAdapter dynamic model/provider getters', () => {
+  // The OTel `gen_ai.request.model` / `gen_ai.provider.name` attributes on
+  // the `user_turn` span are refreshed on every STT event by
+  // `audio_recognition.refreshUserTurnSttAttributes`. Without dynamic
+  // getters that reflect the active child, those attributes are frozen at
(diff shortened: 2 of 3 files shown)


toubatbrian:
@julien-lottie could you confirm if this change is working with real e2e agent runs?

julien-lottie:
> @julien-lottie could you confirm if this change is working with real e2e agent runs?

Yes. This is in prod atm in our system.

toubatbrian (approved):

PR #1517, opened 2026-05-15
Refactor ToolContext to parity class taking a list of Tool | Toolset

toubatbrian:
## Summary

Refactor `Agent({ tools })` to accept a list of `FunctionTool | ProviderDefinedTool | Toolset` entries instead of the legacy `Record<string, FunctionTool>` map, matching the shape of the Python `ToolContext`. `llm.tool({ ... })` now requires an explicit `name`. **This is a breaking change with no backward compatibility shim**.

## Key changes

- **`llm.tool({ name, ... })`** — `name` is now required and stored on the returned `FunctionTool`.
- **`Toolset` base class** — stateful container for a group of tools, with `id`, `tools: readonly Tool[]`, and overridable `setup()` / `aclose()` lifecycle hooks.
- **`ToolContext` is a class** with Python parity:
  - constructor takes `readonly (FunctionTool | ProviderDefinedTool | Toolset)[]`
  - `functionTools` (map view, includes tools flattened from toolsets)
  - `providerTools`, `toolsets`, `tools` (original list)
  - `flatten()`, `getFunctionTool(name)`, `updateTools()`, `copy()`, `equals()`
  - `ToolContext.empty()`
  - Internal storage mirrors Python: original `_tools` list + denormalized `_functionToolsMap` + `_providerTools` + `_toolsets`. Tools from a `Toolset` are recursively flattened into the function/provider collections; duplicate function names referencing different instances throw.
- **`Agent({ tools })` and `agent.updateTools(tools)`** now accept the same list shape.
- All callers across `agents/`, `plugins/`, and `examples/` updated. Tests refreshed and `ToolContext`-class coverage added in `tool_context.test.ts`.

## Usage

```ts
const getWeather = llm.tool({
  name: 'getWeather',
  description: 'Get the weather for a location.',
  parameters: z.object({ location: z.string() }),
  execute: async ({ location }) => `Sunny in ${location}`,
});

const locationTools = new llm.Toolset({
  id: 'location_tools',
  tools: [getWeather],
});

new voice.Agent({
  instructions: 'You are a helpful assistant.',
  tools: [getWeather, locationTools],
});
```

## Test Plan

- `pnpm build` — all 32 packages build
- `pnpm test agents/src/llm/tool_context.test.ts agents/src/llm/tool_context.type.test.ts agents/src/voice/agent.test.ts agents/src/voice/agent_activity.test.ts agents/src/voice/generation_tools.test.ts` — green
- `pnpm lint` — clean (only pre-existing warnings remain)
- `pnpm format:check` — clean
- Remaining failures in the full `pnpm test` run all require external API keys / network and pre-date this change.

## Notes

- The standalone `Toolset` PR (#1485) will be rebased on top of this once this lands.
- A migration helper `toolListFromRecord(map)` is exported for any external code still holding the legacy map shape — it returns a list and stamps the map key onto unnamed tools.


agents/src/llm/tool_context.ts
```diff
@@ -190,38 +196,177 @@ export interface FunctionTool<
...
+export class Toolset {
+  readonly #id: string;
+  readonly #tools: Tool[];
 
-export function isSameToolContext(ctx1: ToolContext, ctx2: ToolContext): boolean {
-  const toolNames = new Set(Object.keys(ctx1));
-  const toolNames2 = new Set(Object.keys(ctx2));
+  constructor({ id, tools }: { id: string; tools: readonly Tool[] }) {
+    this.#id = id;
+    this.#tools = [...tools];
```
toubatbrian: can you dedup here? For tools with same name, later tool overrides earlier

toubatbrian: Done in c3e0a0c. `updateTools()` now uses later-wins semantics — duplicate function names just overwrite the earlier entry instead of throwing. Test updated: `later tool with the same name overrides the earlier one`.

---

agents/src/llm/tool_context.ts
```diff
@@ -190,38 +196,126 @@ export interface FunctionTool<
...
+
+    for (const tool of tools) {
+      if (isProviderDefinedTool(tool)) {
+        this._providerTools.push(tool);
+        continue;
+      }
+      if (isFunctionTool(tool)) {
+        if (!tool.name) {
+          throw new Error('FunctionTool is missing a name');
+        }
```
toubatbrian: Tool name is already required, so this check is unnecessary right?

agents/src/inference/llm.ts
```diff
@@ -379,7 +383,7 @@ export class LLMStream extends llm.LLMStream {
       )) as OpenAI.ChatCompletionMessageParam[];
 
       const tools = this.toolCtx
-        ? Object.entries(this.toolCtx).map(([name, func]) => {
+        ? Object.entries(this.toolCtx.functionTools).map(([name, func]) => {
```
toubatbrian: This is wrong. We need to process flattened tools (include toolsets and provider tool)

Just checking functionTools is not enouigh

agents/src/llm/chat_context.ts
```diff
@@ -835,7 +835,11 @@ export class ChatContext {
         continue;
       }
 
-      if (toolCtx !== undefined && isToolCallOrOutput(item) && toolCtx[item.name] === undefined) {
+      if (
+        toolCtx !== undefined &&
+        isToolCallOrOutput(item) &&
+        toolCtx.getFunctionTool(item.name) === undefined
```
toubatbrian: Again, the logic here needs to be changed once we add toolset. This logic need to be future-proof

agents/src/voice/testing/run_result.ts
```diff
@@ -853,7 +854,7 @@ export class MessageAssert extends EventAssert {
 
     const stream = llm.chat({
       chatCtx,
-      toolCtx: { check_intent: checkIntentTool },
+      toolCtx: new ToolContext([checkIntentTool]),
```
toubatbrian: toolCtx should accepts direct array assignment, make UX cleaner

agents/src/voice/agent_activity.ts
```diff
@@ -484,7 +485,7 @@ export class AgentActivity implements RecognitionHooks {
       }
     }
 
-    const initialTools = Object.keys(this.tools);
+    const initialTools = Object.keys(this.tools.functionTools);
```
toubatbrian: Again, should count for other types of tools like toolsets (future PR) and provider tools

agents/src/llm/tool_context.ts
toubatbrian: Done in 39fdd63 — `tool()` enforces non-empty name at construction, so the runtime check inside `updateTools()` is redundant. Removed.

---

agents/src/inference/llm.ts
toubatbrian: Fixed in 39fdd63. The builder now iterates `this.toolCtx.flatten()` (all function tools + provider tools, with toolset-flattened tools included once the Toolset PR lands) and discriminates on tool type — `FunctionTool` becomes an OpenAI function definition, `ProviderDefinedTool` is skipped here until AJS-112 wires up the per-provider mapping.

---

agents/src/llm/chat_context.ts
toubatbrian: Done in 39fdd63. Added `ToolContext.hasTool(name)` which matches function tools by `name` and provider tools by `id`, and switched the filter predicate here to `!toolCtx.hasTool(item.name)`. Toolset-flattened tools land in `functionTools` once the Toolset PR merges, so they're covered automatically.

---

agents/src/voice/testing/run_result.ts
toubatbrian: Done in 39fdd63. Added a `ToolCtxInput = ToolContext | readonly ToolContextEntry[]` type and a `toToolContext()` normalizer; `LLM.chat({ toolCtx })` and `LLMStream` now accept either form. Updated this call site to `toolCtx: [checkIntentTool]`.

---

agents/src/voice/agent_activity.ts
toubatbrian: Done in 39fdd63. The initial-tools list now includes both `Object.keys(this.tools.functionTools)` and `this.tools.providerTools.map(t => t.id)`. Toolset-flattened function tools land in `functionTools` automatically once the Toolset PR is on top, so they're already counted.

---

agents/src/llm/tool_context.ts
```diff
@@ -190,38 +196,157 @@ export interface FunctionTool<
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const tools: FunctionTool<any, UserData, any>[] = [];
+  for (const [name, fn] of Object.entries(record)) {
+    if (!fn) continue;
+    if (!fn.name) {
+      fn.name = name;
+    }
+    tools.push(fn);
+  }
+  return tools;
```
toubatbrian: We will not have the previous shape since name is a required field in tool, so this method is not needed

agents/src/voice/agent.ts
```diff
@@ -269,7 +270,7 @@ export class Agent<UserData = any> {
   }
 
   get toolCtx(): ToolContext<UserData> {
-    return { ...this._tools };
+    return this._toolCtx;
```
toubatbrian: get a copy of it, not original

agents/src/voice/amd.ts
```diff
@@ -996,10 +997,11 @@ export class AMD extends (EventEmitter as new () => TypedEmitter<AMDCallbacks>)
       },
     });
 
-    const toolCtx: ToolContext = { save_prediction: savePrediction };
+    const toolList: import('../llm/index.js').ToolContextEntry[] = [savePrediction];
```
toubatbrian: Can we avoid this synttax?

agents/src/llm/tool_context.ts
toubatbrian: Done in 6a9e686 — removed entirely. Since `tool()` enforces a non-empty `name` at construction, the legacy `Record<string, FunctionTool>` map never gets produced by current code paths and the migration helper has no callers.

---

agents/src/voice/agent.ts
toubatbrian: Done in 6a9e686. `Agent.toolCtx` now returns `this._toolCtx.copy()` so external callers can't mutate the agent's internal state. Test updated to verify each call returns a fresh instance.

---

agents/src/voice/amd.ts
toubatbrian: Done in 6a9e686. Replaced the inline `import('../llm/index.js').ToolContextEntry` with a proper top-level type import alongside the existing `ToolContext` / `tool` / `isFunctionTool`. The local declaration is now just `const toolList: ToolContextEntry[] = [savePrediction]`.

---

PR #1534, opened 2026-05-18
fix(cartesia): surface TTS websocket server errors

bowdens:
## Description

Encountered this issue when Cartesia had a brief outage: our agents were going silent and were not throwing any errors.

The Cartesia TTS plugin's WebSocket receive loop silently swallowed server-returned error frames; it logged them and `continue`d. The base `SynthesizeStream` never got a thrown error, so `tts_error` was never emitted, retries never ran, and `ttsErrorCounts` / `maxUnrecoverableErrors` escalation never kicked in.

Ports the Python SDK fix ([livekit/agents#3028](https://github.com/livekit/agents/pull/3028) + [#3080](https://github.com/livekit/agents/pull/3080)).

## Changes Made

- Throw a retryable `APIConnectionError` on Cartesia error frames so the base `SynthesizeStream` retry path actually runs.
- Restructure `recvTask` branch order to match Python: any frame with `done: true` — including the error Cartesia returns for empty input on function-call turns — is treated as completion once the sentence stream has closed. Only *pure* error frames raise. Mirroring #3080 up front avoids shipping the same regression Python hit.
- Stop `recvTask`'s `catch` from swallowing `APIError`, and stop the outer `catch` from re-wrapping it via `toRetryableConnectionError`. Without these the new `throw` never reaches the base class.
- Patch changeset for `@livekit/agents-plugin-cartesia` - dunno if that's the right designation for this

## Pre-Review Checklist
- [x] **Build passes**: All builds (lint, typecheck, tests) pass locally.
- [x] **AI-generated code reviewed**.
- [x] **Changes explained**.
- [x] **Scope appropriate**.
- [ ] **Video demo**. n/a

## Testing

- [ ] Automated tests added/updated (if applicable): n/a
- [x] All tests pass

## Additional Notes
n/a


plugins/cartesia/src/tts.ts
```diff
@@ -484,7 +464,7 @@ export class SynthesizeStream extends tts.SynthesizeStream {
               );
               ws.close();
             }, this.#opts.chunkTimeout);
-          } else if (isDoneMessage(serverMsg)) {
+          } else if (serverMsg.done === true) {
```
charlotte-zhuang: FYI most error messages will also have `done === true`. It seems like this branch should explicitly check for `isDoneMessage` so that the error branch can continue throwing to retry?

bowdens: i've change it back to use `isDoneMessage`. i've also added a change to handle Cartesia empty input errors, but a better fix might be to not send the close packet when the tokenizer was empty. Thoughts?

plugins/cartesia/src/tts.ts
```diff
@@ -433,34 +441,8 @@ export class SynthesizeStream extends tts.SynthesizeStream {
             continue;
           }
 
-          // Handle error messages
-          if (isErrorMessage(serverMsg)) {
-            this.#logger.error({ error: serverMsg.error }, 'Cartesia returned error');
-            continue;
-          }
```
charlotte-zhuang, line 440: Rather than tracking non-fatal errors client-side, can you check the `status_code` from the error message?

I also think that it would be great to move error handling back to the top to avoid duplicating the "done" logic, but that's no big deal.

```
          // Handle error messages
          if (isErrorMessage(serverMsg)) {
            // Do not retry the connection on 4xx errors
            // since they can be safely ignored, e.g. empty transcripts
            if (400 <= serverMsg.status_code && serverMsg.status_code < 500) {
              this.#logger.debug({ error: serverMsg.error }, 'Cartesia sent a non-fatal error');
            } else {
              this.#logger.error({ error: serverMsg.error }, 'Cartesia returned error');
              throw new APIConnectionError({
                message: `Cartesia returned error: ${serverMsg.error}`,
                options: { retryable: true },
              });
            }
          }
```

plugins/cartesia/src/tts.ts
```diff
@@ -504,9 +486,57 @@ export class SynthesizeStream extends tts.SynthesizeStream {
...
+                pendingTimedTranscripts.push(
+                  createTimedString({
+                    text: word + ' ', // Add space after word for consistency
+                    startTime,
+                    endTime,
+                  }),
+                );
+              }
+            }
+          } else if (isErrorMessage(serverMsg)) {
```
charlotte-zhuang: In combination with my previous comment, it could be nice to revert this change so the code looks like this:

1. parse message
2. if fatal error, throw to retry. otherwise, log and go to step 3
3. if there are timestamps, emit them
4. if there is audio, emit it
5. if the message is `"type": "done"` or `"type": "error" AND "done": true`, send the last frame

charlotte-zhuang (commented):
I'm a bit concerned with the new change where the client tries to decipher server errors and think that the server errors could serve as a good source of truth.

I also think the new "reducer-like" pattern of branching on server message type is a bit confusing now since the "done" logic would have to run in 2 places.

LMK what you think!

I'm also happy to stack a PR on top of yours with what I had in mind if that's easier.

plugins/cartesia/src/tts.ts
charlotte-zhuang: Nice! I left another review. I suppose I actually think that with the new fatal + non-fatal error logic it makes sense to move away from this `else if (isErrorMessage(serverMsg) {} else if(isDoneMessage(serverMsg)) {}` logic. Sorry about that!

bowdens:
> I'm a bit concerned with the new change where the client tries to decipher server errors and think that the server errors could serve as a good source of truth.
> 
> I also think the new "reducer-like" pattern of branching on server message type is a bit confusing now since the "done" logic would have to run in 2 places.
> 
> LMK what you think!
> 
> I'm also happy to stack a PR on top of yours with what I had in mind if that's easier.

@charlotte-zhuang Thank you for your patience in reviewing this!

I've made some changes to address your comments, let me know if it lines up with your thinking.

charlotte-zhuang (approved):
thanks for the fix!

charlotte-zhuang:
FYI I couldn't actually test the changes since I'm not sure how to run `agents-js` now that [Agents Playground](https://agents-playground.livekit.io/) is gone.

charlotte-zhuang:
I tested it and it seems to work well!

plugins/cartesia/src/tts.ts
```diff
@@ -433,34 +435,24 @@ export class SynthesizeStream extends tts.SynthesizeStream {
...
+          // Handle error frames first. 4xx (e.g. empty-transcript on
+          // function-call turns) is non-fatal — log and fall through so an
+          // accompanying done:true still triggers the unified close path
+          // below. 5xx bubbles up so the base SynthesizeStream can retry.
+          if (isErrorMessage(serverMsg)) {
+            if (serverMsg.status_code >= 400 && serverMsg.status_code < 500) {
+              this.#logger.debug({ error: serverMsg.error }, 'Cartesia sent a non-fatal error');
+            } else {
+              this.#logger.error({ error: serverMsg.error }, 'Cartesia returned error');
+              throw new APIConnectionError({
```
davidzhao: since we have a status code here, we should throw a [APIStatusError](https://github.com/livekit/agents-js/blob/710d112f70ce1134f3098729cda6a950d1deb66c/agents/src/_exceptions.ts#L70C14-L70C28) instead, which carries that detail.

bowdens: :pray: f6d32e2101411c54ad332d964ee09f26d364526d

davidzhao (approved):

PR #1543, opened 2026-05-19
Surface AssemblyAI language confidence metadata

chasef07:
## Summary

- Parse `language_confidence` from AssemblyAI Universal Streaming `Turn` messages.
- Attach the value to emitted transcript alternatives as `metadata.assemblyai.languageConfidence`.
- Cover interim, preflight, and final transcript events with a local WebSocket unit test.

## Why

The AssemblyAI stream already returns both `language_code` and `language_confidence` on turn events when language detection is active. The JS plugin currently maps `language_code` into the standard `SpeechData.language` field but drops the accompanying confidence score.

Keeping the confidence in provider-specific metadata gives applications a practical signal for routing, language-switch detection, analytics, and debugging without changing the framework's core `SpeechData` shape. It also follows the existing `SpeechData.metadata` contract for provider diagnostics that do not map to standard fields.

The metadata is nested under `assemblyai` so it remains clearly provider-specific and does not collide with metadata from other STT providers.

## Validation

- `pnpm exec vitest run plugins/assemblyai/src/stt-metadata.test.ts`
- `pnpm exec prettier --check plugins/assemblyai/src/stt.ts plugins/assemblyai/src/stt-metadata.test.ts .changeset/assemblyai-language-confidence.md`

I also ran `pnpm --filter @livekit/agents-plugin-assemblyai api:check`; it reached API Extractor but exited nonzero on the existing missing API report/release-tag/unresolved-link warnings for this plugin rather than a new type error.

### DIFF
--- a/.changeset/assemblyai-language-confidence.md
+++ b/.changeset/assemblyai-language-confidence.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-assemblyai': patch
+---
+
+Surface AssemblyAI language confidence on speech event metadata.
--- a/plugins/assemblyai/src/stt-metadata.test.ts
+++ b/plugins/assemblyai/src/stt-metadata.test.ts
@@ -1,0 +1,132 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { stt as sttLib } from '@livekit/agents';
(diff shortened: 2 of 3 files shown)


chasef07:
bump

chasef07:
[dlange-aai](https://github.com/dlange-aai)

tinalenguyen (approved):

PR #1544, opened 2026-05-19
fix(google): surface empty Gemini realtime turnComplete as recoverable error

### LINKED ISSUE #1450: Google Realtime plugin silently swallows empty Gemini responses (port of agents#4249)
## Bug Description

The `@livekit/agents-plugin-google` Realtime plugin silently ignores empty server content from Gemini, causing the agent to enter `speaking` state with no audio output. This is the JS equivalent of the bug fixed in the Python SDK by https://github.com/livekit/agents/pull/4249.

## Current Behavior

In [`realtime_api.ts` (line ~1132)](https://github.com/livekit/agents-js/blob/main/plugins/google/src/beta/realtime/realtime_api.ts), when Gemini returns a response with no meaningful server content, the plugin silently drops it:

```ts
if (!hasServerContent) {
  response.serverContent = undefined;
  if (LK_GOOGLE_DEBUG) {
    this.#logger.debug('ignoring empty server content');
  }
}
```

When Gemini sends `turnComplete` with empty content (a [known server-side bug](https://github.com/googleapis/python-genai/issues/2117)), the generation is marked done but no audio was produced — the agent "speaks" silence.

## Expected Behavior

Match the Python SDK fix (PR #4249): raise an `APIStatusError` on empty responses so the SDK's retry/fallback mechanisms can recover, rather than silently accepting a failed turn.

## Impact

In production phone ordering calls using `gemini-3.1-flash-live-preview`, we observe **~50% of deployed calls** experiencing at least one silent stall. The agent enters `speaking` state but produces no audio. The WebSocket stays open, no error events fire. Our application-level stall recovery (realtime text nudge after 8s) catches most of these, but the plugin should handle this at the source.

- Zero stalls locally (shorter conversations, fewer tool calls)
- High stall rate on deployed agents handling real calls (longer contexts, more tool calls — conditions that worsen the Gemini bug per googleapis/python-genai#2117)

## Environment

- `@livekit/agents-plugin-google`: 1.3.4 (also verified in 1.4.0 — still present)
- `@livekit/agents`: 1.3.4
- Model: `gemini-3.1-flash-live-preview`
- Runtime: Node.js on LiveKit Cloud (us-east)

## Related Issues

- https://github.com/livekit/agents/pull/4249 — Python SDK fix for this exact issue
- https://github.com/livekit/agents/issues/4066 — Original bug report (Python)
- https://github.com/googleapis/python-genai/issues/2117 — Gemini server-side root cause (~40 confirmations)
- https://github.com/googleapis/js-genai/issues/707 — Same premature `turnComplete` in JS SDK
- https://github.com/livekit/agents/issues/4706 — Gemini empty streaming responses (~50% intermittent rate)
- https://github.com/google-gemini/live-api-web-console/issues/117 — Google acknowledged as "known issue"

nikhilgupta58:
## Summary

Closes #1450.

Gemini's Live API occasionally emits a `turnComplete` for a generation that produced no audio, no text, and no tool calls (upstream bug [googleapis/python-genai#2117](https://github.com/googleapis/python-genai/issues/2117)). When that happens today, `@livekit/agents-plugin-google` silently marks the generation done; the agent enters `speaking` state with nothing to play, the turn ends, and downstream `AgentActivity` retry/fallback never kicks in. The issue reporter sees this on ~50% of production phone-call turns against `gemini-3.1-flash-live-preview`.

This PR ports the *intent* of the already-merged Python sibling fix [livekit/agents#4249](https://github.com/livekit/agents/pull/4249) — surface the empty response as a retryable error — to the JS realtime streaming model, where the analog of `raise APIStatusError(retryable=True)` is `emit('error', { recoverable: true })`.

## The fix

```ts
// realtime_api.ts: handleServerContent turnComplete branch
const completingGen = this.generationPendingTurnComplete ?? this.currentGeneration;
if (
  completingGen &&
  !this.generationHasOutput(completingGen) &&
  !completingGen._hadToolCall
) {
  this.emitError(
    new APIStatusError({
      message:
        'Gemini realtime: turn completed with no audio, text, or tool call output (see livekit/agents-js#1450)',
      options: { retryable: true, requestId: completingGen.responseId },
    }),
    true,
  );
}
```

Plus the two false-positive guards described next.

## False-positive guards

1. **Tool-call-only turns.** A valid generation can emit zero audio and zero text when the model only produces a function call. We track this with a new `_hadToolCall?: boolean` field on `ResponseGeneration`, initialized to `false` in `startNewGeneration` and set to `true` inside `handleToolCall`.

2. **Same-message ordering race.** A single `LiveServerMessage` can carry both `serverContent.turnComplete` and `toolCall`. Inside `onReceiveMessage`, `handleServerContent` runs **before** `handleToolCall`, so the empty check would see `_hadToolCall === false` and emit a false-positive error. We fix this by setting `_hadToolCall = true` synchronously in `onReceiveMessage` the moment we see `response.toolCall`, before delegating to `handleServerContent`:

   ```ts
   if (response.toolCall && this.currentGeneration) {
     this.currentGeneration._hadToolCall = true;
   }
   ```

3. **Stashed pending generation.** When `startNewGeneration` runs while the previous gen still has an open function channel, it stashes the previous gen as `generationPendingTurnComplete`. The next `turnComplete` belongs to that **stashed** gen, not `currentGeneration`. The check resolves the correct generation as `this.generationPendingTurnComplete ?? this.currentGeneration` and emits the error with the stashed gen's `responseId`.

## Test plan

Five new tests in `plugins/google/src/beta/realtime/realtime_api.test.ts` under `describe('Google Realtime empty turnComplete handling (#1450)')`:

- `emits recoverable error when turnComplete arrives with no audio, text, or tool calls` — asserts `APIStatusError`, `recoverable: true`, `retryable: true`, `requestId` matches the empty gen.
- `does not emit error when turnComplete arrives after audio output` — `_firstTokenTimestamp` set.
- `does not emit error when text output was emitted` — `outputText: 'hi'`.
- `does not emit error when toolCall arrived in the same LiveServerMessage` — verifies the C1 guard: with `_hadToolCall` set synchronously before `handleServerContent` runs, no error fires.
- `checks the stashed pending generation when generationPendingTurnComplete is set` — verifies the C3 fix: when the active `currentGeneration` has tool calls but the stashed gen is empty, the error fires against the stashed gen's `responseId`, not the active one's.

```
$ npx vitest run plugins/google/src/beta/realtime/realtime_api.test.ts

 RUN  v4.0.17

 ✓ plugins/google/src/beta/realtime/realtime_api.test.ts (12 tests) 5ms

 Test Files  1 passed (1)
      Tests  12 passed (12)
```

Verification commands run locally (Node 22, pnpm 9.7):

- `npx vitest run plugins/google/src/beta/realtime/realtime_api.test.ts` — green.
- `npx vitest run plugins/google/src/llm.test.ts` — green (1 skipped, no API key).
- `pnpm format:check` — green.
- `pnpm lint` (google plugin) — clean for changed files; preexisting `no-explicit-any` warnings on `loggableClientEvent`/`loggableServerMessage` (lines 1307/1336/1344) are untouched by this PR.
- `pnpm api:check` (google plugin) — same preexisting api-extractor incompatibility with TS 5.9 that exists on `main` (unrelated `export * as` syntax error in generated `dist/index.d.ts`). No new public surface added by this PR.
- `pnpm changeset` added as patch on `@livekit/agents-plugin-google`.

## Backwards compatibility

- No public API change. No new exports, no new options.
- Consumers of `'error'` events on the realtime session will now see one recoverable error on previously-silent empty `turnComplete` events — the desired behavior per the issue.
- Audio, text, tool-call-only, and mixed turns are unaffected.

## Out of scope

- Suppression config option (no demand; Python fix doesn't have one either).
- Porting the LLM-side empty-response check from livekit/agents#4249 to `plugins/google/src/llm.ts`. Happy to file as a follow-up.
- Symmetric handling of `generationComplete` (Python fix didn't either).

cc @toubatbrian

### DIFF
--- a/.changeset/google-realtime-empty-turn-complete.md
+++ b/.changeset/google-realtime-empty-turn-complete.md
@@ -1,0 +1,15 @@
+---
+'@livekit/agents-plugin-google': patch
+---
+
+fix(google): surface empty Gemini realtime turnComplete as recoverable error
+
+Gemini's Live API occasionally emits a `turnComplete` with no audio, no
+text, and no tool calls (upstream bug
+[googleapis/python-genai#2117](https://github.com/googleapis/python-genai/issues/2117)),
+leaving the agent in `speaking` state with nothing to play. The realtime
+plugin now emits a recoverable `APIStatusError` on those turns so the
+upstream `AgentActivity` can retry, matching the intent of the Python
+sibling fix in [livekit/agents#4249](https://github.com/livekit/agents/pull/4249).
+Tool-call-only turns and turns with any audio or text output are
+unaffected. Closes [#1450](https://github.com/livekit/agents-js/issues/1450).
--- a/plugins/google/src/beta/realtime/realtime_api.test.ts
+++ b/plugins/google/src/beta/realtime/realtime_api.test.ts
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import { Behavior, FunctionResponseScheduling } from '@google/genai';
-import { llm } from '@livekit/agents';
+import { APIStatusError, llm } from '@livekit/agents';
 import { describe, expect, it, vi } from 'vitest';
 import { RealtimeSession } from './realtime_api.js';
 
@@ -13,23 +13,32 @@ type ToolCallStatus = {
   createdAt: number;
 };
 
+type GenerationLike = {
+  functionChannel: {
+    closed: boolean;
+    write: ReturnType<typeof vi.fn>;
+  };
+  outputText?: string;
+  _firstTokenTimestamp?: number;
+  _hadToolCall?: boolean;
+  responseId?: string;
+};
+
 type RealtimeSessionInternals = {
   options: {
     toolBehavior?: Behavior;
     toolResponseScheduling?: FunctionResponseScheduling;
     vertexai?: boolean;
   };
-  currentGeneration?: {
-    functionChannel: {
-      closed: boolean;
-      write: ReturnType<typeof vi.fn>;
-    };
-  };
+  currentGeneration?: GenerationLike;
+  generationPendingTurnComplete?: GenerationLike;
+  earlyCompletionPending?: boolean;
   pendingToolCallIds: Set<string>;
   toolCallStatuses: Map<string, ToolCallStatus>;
   toolResponseCallIds: WeakMap<Record<string, unknown>, string>;
   sendClientEvent: ReturnType<typeof vi.fn>;
   markCurrentGenerationDone: ReturnType<typeof vi.fn>;
+  emitError: ReturnType<typeof vi.fn>;
   getToolResultsForRealtime(
     ctx: llm.ChatContext,
     vertexai: boolean,
@@ -40,6 +49,11 @@ type RealtimeSessionInternals = {
       name?: string;
       args?: Record<string, unknown>;
     }>;
+  }): void;
+  handleServerContent(serverContent: {
+    turnComplete?: boolean;
+    generationComplete?: boolean;
+    interrupted?: boolean;
   }): void;
   clearPendingToolCallIdsForResponses(functionResponses: Array<Record<string, unknown>>): void;
 };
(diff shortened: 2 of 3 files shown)


nikhilgupta58:
Re: Devin's note about the test gap on `onReceiveMessage` branch ordering — accurate. The unit tests cover the post-condition (the empty-check honors `_hadToolCall`) but not the synchronous pre-set itself, because the `#logger` private field blocks `Object.create(prototype)` from driving `onReceiveMessage` end-to-end.

The existing inline comment at the pre-set already explains the ordering reason. A future refactor lifting `#logger` to a regular private would unlock direct testing — happy to do that as a follow-up if maintainers want it in scope.

PR #1562, opened 2026-05-20
fix(telemetry): preserve OTEL logging for stale loggers

### LINKED ISSUE #1561: Agent observability drops logs from loggers captured before OTEL setup
## Summary

Agent observability logging currently swaps the process-global Pino logger when `enableOtelLogging()` runs. Any code that captured `log()` or `log().child(...)` before cloud tracing/recording setup keeps writing to the original Pino instance, so those records never reach the OTEL/Pino cloud exporter and are missing from the observability logs page.

## Minimal reproduction

```ts
initializeLogger({ pretty: false, level: 'info' });
const staleLogger = log().child({ capturedBeforeOtel: true });

initPinoCloudExporter({
  cloudHostname: 'example.livekit.cloud',
  roomId: 'RM_test',
  jobId: 'AJ_test',
});
enableOtelLogging();

staleLogger.info('log from stale logger'); // not exported today
log().info('log from fresh logger');       // exported today
```

This mirrors agent code paths where objects such as job/session helpers can capture a child logger before recording/observability setup enables OTEL logging.

## Expected behavior

Loggers captured before OTEL setup should continue exporting records once observability is enabled.

## Actual behavior

Only logger instances fetched after `enableOtelLogging()` are wired to the OTEL destination.

## Proposed fix

Create loggers with a stable multistream destination from initialization and gate the OTEL destination with an enabled flag. Then `enableOtelLogging()` only flips the flag instead of replacing the global logger, so stale root/child logger instances keep their existing stream and begin exporting once enabled.

Cay-Zhang:
## Summary

- keep a stable Pino multistream logger from initialization so captured root/child loggers retain the OTEL destination
- make `enableOtelLogging()` flip the OTEL export gate instead of replacing the global logger
- add a minimal regression test proving a logger captured before OTEL setup and a fresh logger both export after enablement
- add a patch changeset

Fixes #1561

## Validation

- `pnpm test agents/src/log.test.ts`
- `pnpm format:check`
- `pnpm exec eslint -f unix "agents/src/log.ts" "agents/src/log.test.ts"`
- `pnpm build:agents`

Note: `pnpm --filter @livekit/agents lint` currently fails on pre-existing warnings across unrelated files after the changed-file Prettier errors were fixed. `pnpm --filter @livekit/agents api:check` currently fails on API Extractor not supporting existing `export * as ___` syntax in `agents/dist/index.d.ts`.

### DIFF
--- a/.changeset/stale-loggers-observe.md
+++ b/.changeset/stale-loggers-observe.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+fix(telemetry): export observability logs from logger instances captured before OTEL setup.
--- a/agents/src/log.test.ts
+++ b/agents/src/log.test.ts
@@ -1,0 +1,42 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { enableOtelLogging, initializeLogger, log } from './log.js';
+import { PinoCloudExporter, initPinoCloudExporter } from './telemetry/pino_otel_transport.js';
+
+const OTEL_ENABLED_KEY = Symbol.for('@livekit/agents:otelEnabled');
+
+function resetOtelLoggingState() {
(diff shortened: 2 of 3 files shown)


toubatbrian (approved):
LGTM!

PR #1575, opened 2026-05-21
feat(realtime): add reasoning configuration for gpt-realtime-2 models

yaniv-peretz:
setting [reasoning](https://developers.openai.com/api/reference/resources/realtime/subresources/calls/methods/accept#(resource)%20realtime.calls%20%3E%20(method)%20accept%20%3E%20(params)%200%20%3E%20(param)%20reasoning%20%3E%20(schema)) for gpt-realtime-2 model

### DIFF
--- a/plugins/openai/src/realtime/realtime_model.ts
+++ b/plugins/openai/src/realtime/realtime_model.ts
@@ -159,10 +159,9 @@ export class RealtimeModel extends llm.RealtimeModel {
   }
 
   constructor(
-    options: (
-      | { model: api_proto.ReasoningCapableModel; reasoning?: api_proto.Reasoning }
-      | { model?: string; reasoning?: never }
-    ) & {
+    options: {
+      model?: string; 
+      reasoning?: api_proto.Reasoning;
       voice?: string;
       /** @deprecated Unused in GA API (v1). Temperature is no longer supported. */
       temperature?: number;
--- a/plugins/openai/src/realtime/realtime_model.ts
+++ b/plugins/openai/src/realtime/realtime_model.ts
@@ -48,6 +48,7 @@ interface RealtimeOptions {
   maxResponseOutputTokens?: number | 'inf';
   speed?: number;
   tracing?: api_proto.TracingConfig | null;
+  reasoning?: api_proto.Reasoning;
   apiKey?: string;
   baseURL: string;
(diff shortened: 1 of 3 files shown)


plugins/openai/src/realtime/realtime_model.ts
```diff
@@ -158,8 +159,10 @@ export class RealtimeModel extends llm.RealtimeModel {
   }
 
   constructor(
-    options: {
-      model?: string;
+    options: (
+      | { model: api_proto.ReasoningCapableModel; reasoning?: api_proto.Reasoning }
```
tinalenguyen: i don't think we'll need `ReasoningCapableModel `here, we can just check and warn at runtime if the model doesn't accept `Reasoning`

tinalenguyen (commented):
thanks for the PR! i left a comment and added a changeset

plugins/openai/src/realtime/realtime_model.ts
admin-bot-pulse-dev: I'll leave it as you call.
But let me argue against it.

Actually, `compile-time-error` is much better development experience then `runtime error` for the reason that the error would be catched much faster.

Much better to fail to compile
Gpt-1.5 + reasoning high -> "you are doing something wrong here"

Then a runtime error that could be potentially missed until trying to figure out why it didn't work.

tinalenguyen: Thank you for the reply, I see where you are coming from in terms of user experience. I still think we should   maintain a looser constructor to ensure that it is future proof against newer models, and this would match the pattern set by other plugins in the repo and also in the Python agents repo.

yaniv-peretz: OK, I'll remove so it is not a forever argument.

But in broad terms, I understand the need to be consistent, but I'd say long term trajectory should favor the actual type (even if it is way more complex) cause this just means earlier failures (compile time) and not "this failed in production and now we need to figure out... turns out gpt-1.5 + reasoning is a  no go".

anyway removing it

yaniv-peretz: done
@tinalenguyen please resolve the conversation.

tinalenguyen (approved):
thank you for the PR!!

PR #1620, opened 2026-05-27
feat(agents): remove ttsPronunciationMap Agent option

u9g:
## Summary

Removes the `ttsPronunciationMap` Agent option and the `TTSPronunciationMap` type. This feature duplicated the general text-transform mechanism: pre-TTS pronunciation replacements are already covered by `tts_text_transforms` / the `replace` transform (`voice/transcription/text_transforms.ts`), which also matches how the Python `agents` library handles this (it never had a dedicated pronunciation map — it uses `tts_text_transforms` + `text_transforms.replace`).

## Changes (`agents/src/voice/agent.ts`)
- Drop the `TTSPronunciationMap` type and its export from `voice/index.ts`
- Drop the `ttsPronunciationMap` option, private field, constructor wiring, and public getter on `Agent`
- Remove the `createTTSPronunciationMapStream` / `drainTTSPronunciationBuffer` helpers
- Feed `text` straight into the TTS stream in `ttsNode`, preserving the cancel-on-cleanup behavior
- Remove the now-unused `ReadableStreamDefaultReader` import

Net: +4 / −126.

## Migration
Use a text transform instead:

```ts
new Agent({
  // ...
  // ttsPronunciationMap: { LiveKit: 'live kit' },   // removed
})
```

Apply `replace({ LiveKit: 'live kit' })` via the TTS text-transform pipeline instead.

## Test plan
- `pnpm --filter @livekit/agents build` succeeds (incl. `tsc` declaration emit)
- `pnpm --filter @livekit/agents lint` reports no errors on changed files (only pre-existing `any` warnings)
- Added a `minor` changeset

### DIFF
--- a/.changeset/remove-tts-pronunciation-map.md
+++ b/.changeset/remove-tts-pronunciation-map.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': minor
+---
+
+Remove the `ttsPronunciationMap` Agent option (and the `TTSPronunciationMap` type). Use the general `tts_text_transforms` / `replace` text transform for pre-TTS pronunciation replacements instead.
--- a/agents/src/voice/agent.ts
+++ b/agents/src/voice/agent.ts
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import type { AudioFrame } from '@livekit/rtc-node';
 import { AsyncLocalStorage } from 'node:async_hooks';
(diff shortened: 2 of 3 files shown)


.changeset/remove-tts-pronunciation-map.md
```diff
@@ -0,0 +1,5 @@
+---
+'@livekit/agents': minor
```
toubatbrian, line 2: This should be patch?

toubatbrian (approved):
Otherwise, LGTM

PR #1628, opened 2026-05-27
(openai realtime 2): fix message playout

tinalenguyen:
previously if there was a second message within one generation, it would not be played out

it seems there is no transcript pacing but that may be a pre existing issue ?

## Summary
1. Reorder audio/text forwarding setup in processOneMessage
  Moved the if (audioOutput) { ... } block above the transcriptionNode / performTextForwarding block to match the Python source order. Also changed the first-text-fut fallback condition
   from else if (output.textOut) (gated on audioOutput being null) to a standalone if (!output.audioOut && output.textOut) — so it triggers whenever audio forwarding wasn't actually set
   up, matching Python's if audio_out is None and text_out is not None.

  2. Tighten the playout-await guard
  Replaced let playoutPromise: Promise<...> | null = null plus the trailing await playoutPromise! non-null assertion with a captured playoutEv: PlaybackFinishedEvent | undefined, only
  assigned when the await completes without interruption. Avoids re-awaiting an already-resolved promise and removes the !.
  
  3. Stamp createdAt explicitly when inserting assistant messages
  In addRealtimeMessageOutputs, added createdAt: startedSpeakingAt to the ChatMessage.create call. Previously createdAt defaulted to Date.now() at end-of-generation, so a user STT
  commit landing during agent playout would sort before msg1/msg2 in ChatContext. Now both assistant messages get stamped with the first-frame timestamp, matching Python's
  msg.created_at = started_speaking_at.

### DIFF
--- a/.changeset/realtime-multi-message-ordering.md
+++ b/.changeset/realtime-multi-message-ordering.md
@@ -1,0 +1,16 @@
+---
+'@livekit/agents': patch
+---
+
+fix(realtime): process all messages in multi-message realtime generations
+
+Reorders audio/text forwarding setup inside `processOneMessage` to match the
+Python source order (audio first, then text), and tightens the playout-await
+guard so `playoutPromise` is only awaited when not interrupted. This fixes a
+case where the second message in a multi-message realtime response (e.g.
+`gpt-realtime-2` preambles) could be dropped.
+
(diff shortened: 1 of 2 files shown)


toubatbrian (approved):
Verified by agents-cli, LG!

PR #1653, opened 2026-05-30
fix(tts): treat zero-token tool-only turns as clean end-of-stream in FallbackAdapter

ubinatus:
## Summary
- Track whether the streaming fallback adapter actually forwarded any text tokens to the active TTS provider.
- Treat a completed stream with zero input text tokens as an empty, successful end-of-stream instead of a provider failure.
- Add a regression test for a zero-token streaming TTS turn to ensure providers remain available and fallback does not cascade.

## Root Cause
Tool-only LLM turns can finish without emitting any spoken text. In that case, `FallbackSynthesizeStream` reached the no-audio guard and threw `TTS stream completed but no audio was received`. The fallback adapter then marked the provider unavailable and moved through every configured TTS instance, eventually surfacing `all TTS instances failed` even though empty audio was the correct response.

## Fix
The streaming forwarding path now counts text tokens pushed to the provider. If no audio is received and the adapter input has ended with zero text tokens, it emits the end-of-stream sentinel, waits for the input reader to settle, and returns successfully. If text was present but no audio came back, the existing fallback error path is preserved.

## Validation
- `pnpm exec prettier --write agents/src/tts/fallback_adapter.ts agents/src/tts/fallback_adapter.test.ts`
- `pnpm exec vitest run agents/src/tts/fallback_adapter.test.ts`
- `pnpm --dir agents exec eslint -f unix src/tts/fallback_adapter.ts src/tts/fallback_adapter.test.ts`
- `pnpm --dir agents exec throws-check src/tts/fallback_adapter.ts`

### DIFF
--- a/agents/src/tts/fallback_adapter.test.ts
+++ b/agents/src/tts/fallback_adapter.test.ts
@@ -190,6 +190,49 @@ describe('TTS FallbackAdapter', () => {
     await adapter.close();
   });
 
+  it('should end cleanly when the stream input has zero text tokens', async () => {
+    const primary = new MockTTS('primary');
+    const secondary = new MockTTS('secondary');
+    const adapter = new FallbackAdapter({
+      ttsInstances: [primary, secondary],
+      maxRetryPerTTS: 0,
+      recoveryDelayMs: 60_000,
+    });
+
+    const stream = adapter.stream();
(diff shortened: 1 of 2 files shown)


ubinatus:
Duplicate of #1455

PR #1662, opened 2026-06-01
fix(voice): don't strand playback-segment counter on interrupted-while-paused frame

enriqueespaillat-gyde:
## Summary

`ParticipantAudioOutput.captureFrame` can permanently desync its playback-segment counter when a frame arrives while the output is paused (interruption detection) and is then interrupted. Afterwards every `waitForPlayout()` on that output blocks forever, which can hang the agent.

## Root cause

`captureFrame` calls `super.captureFrame(frame)` — which bumps the base `AudioOutput`'s `playbackSegmentsCount` — *before* the pause/interrupt gate:

```ts
super.captureFrame(frame);            // playbackSegmentsCount++

if (!this.playbackEnabledFuture.done) {
  this.audioSource.clearQueue();
  await Promise.race([this.playbackEnabledFuture.await, this.interruptedFuture.await]);
  if (this.interruptedFuture.done) {
    return;                           // pushedDuration is never incremented
  }
}
```

When a frame bails at that early `return` (interrupted while paused), the segment was already counted, but:

- `pushedDuration` is never incremented, so
- `flush()` short-circuits on `if (!this.pushedDuration) return` and never starts `waitForPlayoutTask`, so
- `onPlaybackFinished` is never called and `playbackFinishedCount` never catches up.

`playbackSegmentsCount` is now permanently ahead of `playbackFinishedCount`, so the next segment's `waitForPlayout()` (`while (playbackFinishedCount < playbackSegmentsCount) await ...`) never resolves.

### Why it can hang the agent

The stalled turn's reply pipeline awaits `audioOutput.waitForPlayout()` and so never reaches `_markGenerationDone()`. Because the speech handle is also never interrupted, `AgentActivity.mainTask`'s `await speechHandle.waitIfNotInterrupted([speechHandle._waitForGeneration()])` has neither of its two exit conditions and parks forever — the agent goes silent until the participant disconnects. This is easiest to hit when a tool-call-only turn (no spoken text) immediately follows a barge-in. (Same hang surface as the `mainTask` issues #836 / #1089, but a distinct root cause in the audio output.)

## Fix

Register the playback segment (`super.captureFrame`) only after the pause/interrupt gate is cleared, so a frame that bails never counts a segment.

## Test

Adds a `ParticipantAudioOutput captureFrame segment accounting` suite to `_output.test.ts`:

- a frame interrupted while paused must not register a segment, and a following `waitForPlayout()` resolves (this fails before the fix — `playbackSegmentsCount` is left at 1 and `waitForPlayout()` hangs);
- the normal (non-paused) path still registers and pushes the frame.

### DIFF
--- a/.changeset/audiooutput-segment-counter-hang.md
+++ b/.changeset/audiooutput-segment-counter-hang.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Fix ParticipantAudioOutput stranding its playback-segment counter when a frame is interrupted while paused. `captureFrame` registered the segment (via `super.captureFrame`) before the pause/interrupt gate, so a frame that bailed at the gate left `playbackSegmentsCount` ahead of `playbackFinishedCount` forever — every subsequent `waitForPlayout()` then blocked, which could hang the agent's main loop (the stalled turn is never interrupted, so `mainTask` parks on `_waitForGeneration()`). The segment is now counted only after the gate is cleared.
--- a/agents/src/voice/room_io/_output.test.ts
+++ b/agents/src/voice/room_io/_output.test.ts
@@ -5,6 +5,8 @@ import { describe, expect, it, vi } from 'vitest';
 import { Future } from '../../utils.js';
 import { ParticipantAudioOutput } from './_output.js';
 
+type CaptureFrameArg = Parameters<ParticipantAudioOutput['captureFrame']>[0];
+
 const nextTick = () => new Promise<void>((resolve) => setImmediate(resolve));
 
 describe('ParticipantAudioOutput waitForPlayoutTask', () => {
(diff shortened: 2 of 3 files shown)


enriqueespaillat-gyde:
@toubatbrian this is an issue we run into often, im not entirely confident on the fix. Essentially causes the agent to hang if interrupted mid tool call.

toubatbrian (approved):

toubatbrian:
> Essentially causes the agent to hang if interrupted mid tool call.

@enriqueespaillat-gyde Just to clarify, is this before or after this fix? From my local testing on your PR, I think this is the right fix.

enriqueespaillat-gyde:
> > Essentially causes the agent to hang if interrupted mid tool call.
> 
> @enriqueespaillat-gyde Just to clarify, is this before or after this fix? From my local testing on your PR, I think this is the right fix.

I think the only problem for us - is that it doesn't fully fix the issue. I think I got closer to the root cause here - https://github.com/livekit/agents-js/pull/1732

I'm less confident on the right fix though.

PR #1681, opened 2026-06-02
docs(cartesia): add cartesia stt to readme

charlotte-zhuang:
This change adds Cartesia STT to the list of plugins ported from `agents`, done in #1587.

I suppose the npm package doesn't actually support Cartesia STT yet since a new release is required. Let me know if you prefer to keep the README in sync with the latest released package rather than what's implemented in `main`.


tinalenguyen (approved):
Hi @charlotte-zhuang, thanks for the PR! I think it's fine to update based on what is in `main`

PR #1736, opened 2026-06-09
Async Toolsets

toubatbrian:
## Summary

Adds first-class support for **long-running, non-blocking tools** to the JS voice pipeline. A tool can now acknowledge the user immediately, keep working in the background, and deliver its result later as a fresh assistant turn — without holding up the conversation, and without the LLM hallucinating a result while the work is still in flight.

The feature has two layers:

1. **Non-blocking tool execution** — any tool can call `ctx.update(...)` to return control to the conversation immediately while it continues running. When it finishes, its output is enqueued and the agent generates a natural follow-up.
2. **`AsyncToolset`** — a `Toolset` with its own `ToolExecutor` so background work has a well-defined **scope** (agent-scoped or session-scoped) and a deterministic **lifecycle** at agent handoff / session close (cancel cancellable work, drain the rest).

It also ships the supporting machinery: duplicate-call policies, cooperative cancellation via `AbortSignal`, a `ctx.foreground()` escape hatch for tools that need to grab the floor mid-flight (e.g. to ask a follow-up question), and two built-in management tools the model can call (`lk_agents_get_running_tasks`, `lk_agents_cancel_task`).

## High-level functionality

- **Go non-blocking on demand.** First `ctx.update(message)` resolves the tool call to the model with an interim message and flags the call `__livekit_agents_tool_non_blocking`. The agent keeps listening; the tool keeps running.
- **Deferred replies.** When a non-blocking tool finally returns (or calls `ctx.update` again), the output is enqueued and delivered as a new assistant turn, gated on the session being idle so it never talks over the user.
- **Scoping.**
  - *Agent-scoped* `AsyncToolset` (placed in an `Agent`'s `tools`): its in-flight work is drained/cancelled when that agent hands off.
  - *Session-scoped* `AsyncToolset` (placed on the `AgentSession`): its work **survives agent handoffs** and is only torn down at session close.
- **Cancellation.** Tools flagged `ToolFlag.CANCELLABLE` can be cancelled — by the model (`lk_agents_cancel_task`), by a duplicate `replace`, or by scope teardown. Cancellation is cooperative: the tool's `execute()` receives an `AbortSignal`.
- **Duplicates.** Per-tool `onDuplicate` policy decides what happens when the model calls a tool that's already running: `allow` | `reject` | `confirm` | `replace`.
- **Foreground hand-back.** `ctx.foreground(fn)` lets a running background tool grab the conversation floor (e.g. to run a nested `AgentTask` that asks the user a question) and then return to the background.

### 1. A non-blocking tool

```ts
import { llm } from '@livekit/agents';
import { z } from 'zod';

const bookFlight = llm.tool({
  name: 'book_flight',
  description: 'Book a flight. Runs in the background.',
  flags: llm.ToolFlag.CANCELLABLE, // opt in to cooperative cancellation
  onDuplicate: 'confirm',
  parameters: z.object({
    origin: z.string(),
    destination: z.string(),
    date: z.string(),
  }),
  execute: async ({ origin, destination, date }, { ctx, abortSignal }) => {
    // First update => the tool call resolves NOW with this interim message,
    // the agent stops blocking, and execution continues here in the background.
    await ctx.update(`Searching flights ${origin} → ${destination} on ${date}…`);

    const result = await searchAndBook({ origin, destination, date }, abortSignal);

    // Returned value is delivered later as a fresh assistant turn.
    return `Booked ${result.airline}, confirmation ${result.code}.`;
  },
});
```

### 2. Grouping into an `AsyncToolset` (scoped lifecycle)

```ts
import { AsyncToolset } from '@livekit/agents';

// Agent-scoped: drained/cancelled when this agent hands off.
const bookingTools = AsyncToolset.create({
  id: 'booking',
  tools: [bookFlight, cancelBooking],
});

const agent = new Agent({ instructions: '…', tools: [bookingTools] });
```

```ts
// Session-scoped: survives handoffs, torn down at session close.
const session = new AgentSession({
  /* stt, llm, tts, … */
  tools: [longRunningJobs],
});
```

### 3. Foreground hand-back (ask a follow-up mid-flight)

```ts
execute: async (args, { ctx, abortSignal }) => {
  await ctx.update('Working on it…');

  // Grab the floor to collect a missing detail via a nested AgentTask.
  const { emailAddress } = await ctx.foreground(async () => {
    ctx.session.say('I need your email to confirm.');
    return collectEmailTask().run();
  });
  if (abortSignal.aborted) throw new Error('aborted');

  return finishBooking(args, emailAddress);
};
```

### 4. Configuring async behavior (templates + per-scope override)

```ts
// Session/agent level:
new AgentSession({ toolHandling: { asyncOptions: { updateTemplate, duplicateConfirmTemplate } } });
new Agent({ toolHandling: { asyncOptions: { … } } });

// Per-toolset override:
AsyncToolset.create({ id: 'booking', tools: [...], toolHandling: { asyncOptions: { … } } });
```

Customizable templates (all have sensible defaults): `updateTemplate`, `duplicateRejectTemplate`, `duplicateConfirmTemplate`, `replyAtTailTemplate`, `replyMaybeCoveredTemplate`.

### 5. Built-in management tools (auto-injected)

When any cancellable tool is in scope, the framework advertises two tools to the model so it can reason about background work:

- `lk_agents_get_running_tasks` — list cancellable in-flight calls.
- `lk_agents_cancel_task(call_id)` — cancel one.

## Important notices

- **Cancellation is cooperative — you must honor `abortSignal`.** If your `execute()` ignores the signal, a "cancelled" tool keeps running. On scope teardown the executor waits for such tools up to `LK_DRAIN_TOOL_TIMEOUT_MS` and then logs an **error** telling you the tool wasn't abortable. Cancelling is only permitted when the tool is `CANCELLABLE` **and** the speech allows interruptions.
- **Non-cancellable background tools block handoff/teardown until they finish.** A `drain()` awaits non-cancellable in-flight tools to completion (matching Python's `cancel_all(cancellable_only=True)`) — by design, so their deferred result is delivered rather than dropped. Keep genuinely long, non-cancellable work off the agent-scoped path if you can't tolerate that wait at handoff.
- **Agent-scoped vs session-scoped is a real semantic choice.** Agent-scoped work is torn down at handoff; session-scoped work survives. Pick deliberately.
- **`ctx.update` is the non-blocking trigger.** A tool that never calls `ctx.update` behaves like a normal blocking tool. The *first* `update` is what flips it non-blocking; subsequent updates/return are delivered as deferred replies.
- **Deferred replies are idle-gated, not immediate.** Results land as a new assistant turn once the session is idle, so timing is conversational — don't assert on them firing the instant `execute()` returns.
- **`IGNORE_ON_ENTER` is not honored for tools nested inside a toolset** (only top-level function tools are filtered during `onEnter`). This matches the Python framework contract (`# TODO(long): add IGNORE_ON_ENTER to ToolSet?`). Toolsets pass through `onEnter` intact so their executor scoping survives.
- **Preemptive generation reuse with cancellable tools.** The preemptive-generation snapshot captures the full merged tool set (session + agent + injected management tools) so the reuse check isn't invalidated whenever a cancellable/session-scoped tool is present.
- **Tool/system isolation, env knobs.** Drain deadlines are env-overridable: `LK_DRAIN_PLAYOUT_TIMEOUT_MS`, `LK_DRAIN_TOOL_TIMEOUT_MS`.

## Testing

- Unit suites: `agent_activity.test.ts` (pause/drain deadlock guards, preemptive snapshot), `tool_executor.test.ts` (duplicate policies, drain, reply re-arm), `async_toolset.test.ts`, `synchronizer.test.ts`, `audio_recognition_handoff.test.ts`.
- End-to-end via `agents-cli` (voice + text): book-flight async flow with `ctx.foreground` email capture runs to completion; the agent recovers from every `scheduling paused` and delivers the final deferred confirmation.


agents/src/voice/agent.ts
```diff
@@ -140,6 +141,7 @@ export interface AgentOptions<UserData> {
   llm?: LLM | RealtimeModel | LLMModels;
   tts?: TTS | TTSModelString;
   turnHandling?: TurnHandlingOptions;
+  toolHandling?: ToolHandlingOptions;
```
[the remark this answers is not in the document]
toubatbrian: false alarm

agents/src/voice/tool_executor.ts
```diff
@@ -0,0 +1,663 @@
...
+    void rawPromise.finally(() => clearTimeout(timer));
+  }
+
+  async drain(): Promise<void> {
+    const tasks = [...this.runningTasks.values()];
+
+    // Cancellable tools: signal abort + abandon. Non-cancellable: let them run.
+    for (const task of tasks) {
+      if (task.allowCancellation) {
+        await this.cancel(task.ctx.functionCall.callId);
```
longcw: this may raise if `allowInterruptions` is False?

agents/src/voice/agent_activity.ts
```diff
@@ -4245,14 +4282,41 @@ export class AgentActivity implements RecognitionHooks {
...
+
+    await this.setupToolsetList(toSetup);
     // Re-flatten now that any factory toolsets have resolved their tools, so they're advertised.
     this.agent._toolCtx.updateTools(this.agent._toolCtx.tools);
   }
 
   private async closeToolsets(): Promise<void> {
     if (!this._toolsetsSetup) return;
     this._toolsetsSetup = false;
+    // down, so a non-cancellable async tool finishes instead of being cancelled.
```
longcw:
```suggestion
```

longcw (commented):
tested locally and it works well. something nit:

agents/src/voice/agent_activity.ts
```diff
@@ -1710,6 +1723,36 @@ export class AgentActivity implements RecognitionHooks {
     return this.agentSession.chatCtx;
   }
 
+  async waitForIdle(
```
longcw, line 1833: what will happen if the agent activity closed when waiting on this?

toubatbrian: fixed, now we can abort the waitForIdle

agents/src/voice/tool_executor.ts
toubatbrian: Fixed: `drain()` now calls a guard-free `forceCancelTask()` (teardown should force-cancel regardless), while `cancel()` keeps the guards for the LLM path. Added a red→green regression test.

longcw (approved):

PR #1740, opened 2026-06-09
feat(console): inference runner support + recording parity for lk console

u9g:
Stacked on #1706. Rounds out the console CLI runner so real-world agents work under `lk agent console`.

## Changes

**Inference runner support + `--record` parity in console mode** (7981e4a)
- Local `InferenceRunner`s (e.g. the LiveKit turn detector) now initialize and serve in console mode, matching worker mode
- `console --record` wires up the same local recording output as the Python console

**Fail `SupervisedProc.initialize()` fast when the child dies during startup** (cd333ef)
- A child process that exits mid-initialize previously left `initialize()` hanging until its timeout; it now rejects as soon as the process dies

## Testing
- E2E via `lk agent console examples/src/basic_agent.ts`: full voice session with cloud inference (Deepgram STT, GPT-4.1-mini, Cartesia TTS) plus the local turn-detector inference runner


agents/src/console.ts
```diff
@@ -114,41 +119,83 @@ export async function runConsole({
...
-    await runWithJobContextAsync(ctx, async () => agent.entry(ctx));
+    // Plugins that registered an InferenceRunner (e.g. the livekit turn
+    // detector) run inference in a supervised child process, same as the
+    // worker path. Without any runners the fallback executor just raises if
+    // reached.
+    let inferenceExecutor: InferenceExecutor = new ConsoleInferenceExecutor();
+    if (Object.keys(InferenceRunner.registeredRunners).length > 0) {
+      inferenceProc = new InferenceProcExecutor({
+        runners: InferenceRunner.registeredRunners,
+        initializeTimeout: 30000,
```
toubatbrian: Python uses `initialize_timeout=5 * 60` for the inference executor, should console use the same 5 minute timeout?

agents/src/ipc/supervised_proc.ts
```diff
@@ -195,13 +193,42 @@ export abstract class SupervisedProc {
...
+      },
+    );
+    const onExit = once(this.proc!, 'exit', { signal: abort.signal }).then(([code, signal]) => {
+      throw new Error(`process exited before initializing (code ${code}, signal ${signal})`);
+    });
+
+    try {
+      await Promise.race([onMessage, onExit, timeout]);
+      this.init.resolve();
+    } catch (err) {
```
toubatbrian, line 222: Should `initialize()` clean up the child on failure here?

agents/src/job.ts
```diff
@@ -366,7 +373,19 @@ export class JobContext<ProcessUserData = Record<string, unknown>> {
 
     const report = this.makeSessionReport(session);
 
-    // TODO(brian): Implement CLI/console
+    // Console recording: dump the session report to a local file (mirrors python).
+    const agentsConsole = AgentsConsole.getInstance();
```
toubatbrian, line 377: File: agents/src/voice/agent_session.ts

nit: python's `--record` forces audio recording even if the agent passes `record: false` (`agent_session.py` ORs `(c.enabled and c.record)` into the recorder check). Here it only checks `_enableRecording`, so `lk console --record` + `session.start({ record: false })` yields a `session_report.json` but no `audio.ogg`.

```suggestion
      const consoleForcesRecord = consoleInst.enabled && consoleInst.record;
      if (this.input.audio && this.output.audio && (this._enableRecording || consoleForcesRecord)) {
```

agents/src/console.ts
u9g: done.

agents/src/ipc/supervised_proc.ts
u9g: done.

agents/src/job.ts
u9g: done.

toubatbrian (approved):

PR #1765, opened 2026-06-11
fix(voice/room_io): non-delta transcription final-stream races

### LINKED ISSUE #1759: Non-delta transcription output publishes wrong/empty text on the final stream (latestText race in ParticipantTranscriptionOutput)
### Describe the bug

Affected: @livekit/agents 1.4.4, still present on main
(agents/src/voice/room_io/_output.ts).

Setup: voice.AgentSession with Deepgram STT (interimResults: true), user
transcription forwarded to the room (non-delta path, default
transcriptionEnabled). Client renders via useTranscriptions
(@livekit/components-react 2.9.21).

Bug 1: final stream races with the next segment's capture

In ParticipantTranscriptionOutput (non-delta), handleFlush() schedules
flushTaskImpl, which later writes this.latestText to a new stream with
lk.transcription_final: "true". But captureText() assigns
this.latestText = payload before handleCaptureText() awaits the pending
flush task. When the STT emits several is_final chunks in a burst (normal for
Deepgram mid-utterance finals), the next chunk's capture overwrites
latestText before the previous segment's flush has written it, so segment A's
final stream is published carrying segment B's text.

Observed live: the learner said "So, you made a big purchase with, a service.
You tell me what exactly it was?" The first segment's interims displayed
correctly, then its final stream arrived carrying service. (the next chunk's
text), and the client — which keys one entry per lk.segment_id and takes the
last write — replaced the full sentence with the fragment.

Suggested fix: snapshot the text when the flush is scheduled and pass it as an
argument instead of reading the shared field inside the task:

tsprotected handleFlush() {
  const currWriter = this.writer;
  this.writer = null;
  const textToFlush = this.latestText;
  this.flushTask = Task.from((controller) =>
    this.flushTaskImpl(currWriter, textToFlush, controller.signal),
  );
}

Bug 2: resetState() wipes the first capture's text

captureText() sets this.latestText = payload, then handleCaptureText()
runs resetState() (which sets latestText = '') when a new segment starts.
So for any segment whose first event is already final (no prior interims —
again common with multi-final bursts), the final stream publishes an empty
string. An empty write produces no chunk, so subscribers keyed on the segment
never receive the final text at all.

Suggested fix: re-assign the captured text after the reset in
handleCaptureText, or stop clearing latestText in resetState().

Compounding client-side behavior (components-react)

setupTextStream (components-core 0.12.13) keeps one entry per
lk.segment_id, overwrites text with each new stream's payload, and never
updates streamInfo — so the corrupted final-stream text silently replaces
correct interim text, and lk.transcription_final permanently reflects the
first stream received for the segment (i.e. "false"). Happy to file that
separately against components-js if useful.

### Relevant log output

_No response_

### Describe your environment

  System:
    OS: macOS 15.7.7
    CPU: (10) arm64 Apple M1 Max
    Memory: 434.92 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.22.3 - /Users/jonas/.nvm/versions/node/v22.22.3/bin/node
    Yarn: 1.22.22 - /Users/jonas/.nvm/versions/node/v22.14.0/bin/yarn
    npm: 10.9.8 - /Users/jonas/.nvm/versions/node/v22.22.3/bin/npm
    pnpm: 11.5.2 - /Users/jonas/.nvm/versions/node/v22.14.0/bin/pnpm

_No response_

### Additional information

_No response_

tsushanth:
## Summary

`ParticipantTranscriptionOutput` (non-delta path, default user-transcription forwarding) had two race conditions that surfaced with Deepgram-style mid-utterance final bursts where multiple `is_final` chunks arrive back-to-back.

### Bug 1 — next-segment capture overwrites the in-flight flush text

`handleFlush()` schedules `flushTaskImpl`, which reads `this.latestText` when it later writes the `lk.transcription_final: \"true\"` stream. When the next segment's first `captureText()` lands before that task runs, it overwrites `latestText`, so segment A's final stream publishes segment B's text.

Observed in the issue: the learner said *\"So, you made a big purchase with, a service. You tell me what exactly it was?\"* — segment A's final stream arrived carrying just *\"service.\"* (the next chunk's fragment), and the client — keyed one entry per `lk.segment_id`, last write wins — replaced the full sentence.

**Fix:** snapshot `latestText` when the flush task is scheduled and pass it as a parameter to `flushTaskImpl`.

```diff
 protected handleFlush() {
   const currWriter = this.writer;
   this.writer = null;
-  this.flushTask = Task.from((controller) => this.flushTaskImpl(currWriter, controller.signal));
+  const textToFlush = this.latestText;
+  this.flushTask = Task.from((controller) =>
+    this.flushTaskImpl(currWriter, textToFlush, controller.signal),
+  );
 }
```

### Bug 2 — `resetState()` wipes the captured text on a final-only first event

`captureText()` sets `this.latestText = payload`, then `handleCaptureText()` runs `resetState()` (which sets `latestText = ''`) when a new segment starts. For any segment whose first event is already `is_final` — common with multi-final bursts — the final stream publishes an empty string. An empty write produces no chunk, so subscribers keyed on the segment never receive the final text.

**Fix:** restore `latestText` from the captured payload immediately after the `resetState()` call in `handleCaptureText`. The legacy class (`ParticipantLegacyTranscriptionOutput`) is unaffected — it tracks `pushedText` separately and resets that explicitly.

```diff
 if (!this.capturing) {
   this.resetState();
   this.capturing = true;
+  this.latestText = text;
 }
```

## Test plan

- [x] Two new regression tests in [agents/src/voice/room_io/_output.test.ts](agents/src/voice/room_io/_output.test.ts) — both fail on `main`, both pass after the fix
- [x] `pnpm vitest run agents/src/voice/room_io/_output.test.ts` — 5/5 pass (3 existing + 2 new)
- [x] `pnpm build:agents` — clean
- [x] `pnpm lint` — no new warnings on touched files
- [x] `pnpm format:check` — clean
- [x] Changeset added (patch)
- [x] Legacy `ParticipantLegacyTranscriptionOutput` semantics untouched

Closes #1759

### DIFF
--- a/.changeset/fix-non-delta-transcription-final-text.md
+++ b/.changeset/fix-non-delta-transcription-final-text.md
@@ -1,0 +1,20 @@
+---
+'@livekit/agents': patch
+---
+
+Fix `ParticipantTranscriptionOutput` non-delta final-stream publishing two
+races that surfaced with Deepgram-style mid-utterance final bursts.
+
+1. `handleFlush()` previously read `this.latestText` from inside
+   `flushTaskImpl`, so a `captureText()` for the next segment landing
+   before the flush task ran would overwrite the field and cause segment
+   A's `lk.transcription_final: "true"` stream to publish segment B's
+   text. The text to flush is now snapshotted when the task is scheduled
+   and passed in as a parameter.
+
+2. When the first event for a fresh segment was already `is_final`,
+   `handleCaptureText` called `resetState()` after `captureText` had set
+   `latestText`, clearing it back to `""`. The subsequent final stream
+   then published an empty string. `latestText` is now restored from the
+   captured payload immediately after `resetState()` so the same-tick
+   final preserves the captured text.
--- a/agents/src/voice/room_io/_output.test.ts
+++ b/agents/src/voice/room_io/_output.test.ts
@@ -2,8 +2,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import { describe, expect, it, vi } from 'vitest';
-import { Future } from '../../utils.js';
-import { ParticipantAudioOutput } from './_output.js';
+import { Future, type Task } from '../../utils.js';
+import { ParticipantAudioOutput, ParticipantTranscriptionOutput } from './_output.js';
 
 type CaptureFrameArg = Parameters<ParticipantAudioOutput['captureFrame']>[0];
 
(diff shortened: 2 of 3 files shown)


tsushanth:
Thanks for the read. The PR scope was the in-flight \`latestText\` race, but you're right that the asymmetry — \`currentId\` still read lazily inside \`createTextWriter\` — leaves \`setParticipant()\` exposed to the same shape of bug whenever a flush is in-flight when it fires. Pushed [a98b30d](https://github.com/livekit/agents-js/pull/1765/commits/a98b30d) closing it symmetrically:

- \`handleFlush()\` now also snapshots \`this.currentId\` into a local \`segmentIdToFlush\` and threads it through to \`flushTaskImpl\`
- \`flushTaskImpl\` seeds the \`final: 'true'\` attribute bundle with the snapshotted \`ATTRIBUTE_TRANSCRIPTION_SEGMENT_ID\` before calling \`createTextWriter(attributes)\`
- \`createTextWriter\` is now non-clobbering on \`SEGMENT_ID\`: it falls back to \`this.currentId\` only when the caller hasn't already set one, so the existing live-capture path is unchanged

Now \`setParticipant()\`'s synchronous \`flush() → resetState()\` can't shift the segment id under the queued flush — both the text and the id refer to the turn that was actually being captured. The room_io unit tests still pass (\`pnpm test agents/src/voice/room_io\`: 13/13).

tsushanth:
Good catch — \`trackId\` and \`participantIdentity\` are the remaining lazily-read fields and have the exact same shape: \`setParticipant()\` writes \`participantIdentity\` and \`trackId\` (via \`findMicrophoneTrackId\`) before calling \`flush() + resetState()\`, so the queued \`flushTaskImpl\` ends up reading the next turn's values when it builds attributes / opens the stream.

Pushed [05d6d0c](https://github.com/livekit/agents-js/pull/1765/commits/05d6d0c) closing it symmetrically:

- \`handleFlush()\` snapshots \`trackIdToFlush\` and \`participantIdentityToFlush\` alongside the existing \`textToFlush\` / \`segmentIdToFlush\`
- \`flushTaskImpl\` now takes them as parameters, builds \`ATTRIBUTE_TRANSCRIPTION_TRACK_ID\` off the snapshot, and threads the identity into the \`createTextWriter\` call
- \`createTextWriter\` gains an optional \`participantIdentityOverride\` — used as the \`senderIdentity\` on \`streamText({...})\` when the snapshotted flow passes it in; live captures keep reading from \`this.participantIdentity\`

Build clean, room_io tests 13/13 pass.

tsushanth:

PR #1768, opened 2026-06-11
fix(ipc): make SupervisedProc.initialize() always settle (closes #1748 wedge)

### LINKED ISSUE #1748: ProcPool wedges permanently when a warming child dies before its first IPC message — initialize() never settles, worker keeps accepting jobs it can never launch
## Summary

In `@livekit/agents` ≥1.4.x (verified on 1.4.4 and 1.4.5 — the relevant files are byte-identical), a warming child process that **dies or hangs during import/prewarm without sending its first IPC message** permanently wedges the worker's process pool: the worker stays registered and available, keeps logging `received job request` and accepting, but never launches another job until the process is externally restarted.

We hit this repeatedly in a production contact-center deployment (5 independent wedge events across 2 days, each black-holing all calls routed to the worker for hours). Related: the review on #1669 already confirmed "the bug is real in JS" before that PR was closed unmerged; see also #909 and #927.

## Mechanism (file:line refs from 1.4.4 dist)

1. `SupervisedProc.initialize()`'s only completion path is the child's first IPC message: `await once(this.proc, "message")` (`dist/ipc/supervised_proc.js:139`). A child killed mid-prewarm (e.g. kernel OOM, V8 heap abort, import crash in the agent module) emits neither `"message"` nor `"error"` — normal exit fires no `"error"` — so `events.once` **pends forever**.
2. The `initializeProcessTimeout` timer only rejects the side future `this.init` (`supervised_proc.js:123-125`). It does **not** kill the child and does **not** unblock `initialize()`.
3. The `init` rejection makes `run()` throw at `await this.init.await`, which `start()`'s catch swallows into a single WARN — `supervised process run failed` (`supervised_proc.js:46-49`). Because `run()` died at its first await, the exit listener / ping loop / memory monitor were never attached.
4. Meanwhile `ProcPool.procWatchTask` is still parked at `await proc.initialize()` (`dist/ipc/proc_pool.js:85`), holding **both** `initMutex` and its `procMutex` slot — the releases sit in a `finally` that is unreachable while `initialize()` pends. With `numIdleProcesses: 1` that is the **only** slot, so `run()` blocks at `procMutex.lock()` (`proc_pool.js:120`) forever, `warmedProcQueue` never refills, and every accepted job blocks at `warmedProcQueue.get()` (`proc_pool.js:41`) — unbounded, no timeout.
5. The worker keeps reporting available (load loop is CPU-based), so the server keeps dispatching jobs into the black hole.

Production log signature: exactly one `supervised process run failed` WARN per wedge (≈60 s after the last successful prewarm), then `received job request` lines forever with zero launches. We matched this 5/5 against our wedge onsets.

## Regression vs 1.0.46

In 1.0.46, `start()` called `this.run()` bare (no `.catch`), so the same init-timeout rejection escaped as an **unhandledRejection** — in our deployment that crashed the worker process and the orchestrator restarted it. Accidentally self-healing. 1.4.x's added `.catch` turns the same failure into a silent permanent wedge. (Not arguing for the crash — arguing the timeout should actually recover the pool.)

## Minimal repro

```js
const path = require("path");
const { fork } = require("child_process");
const fs = require("fs");
const base = path.dirname(require.resolve("@livekit/agents")); // dist/
const { initializeLogger } = require(path.join(base, "log.cjs"));
initializeLogger({ pretty: false, level: "silent" });
const { SupervisedProc } = require(path.join(base, "ipc/supervised_proc.cjs"));

const hang = "/tmp/lk-hang.cjs";
fs.writeFileSync(hang, "setInterval(() => {}, 1000);\n"); // never sends a message

const p = new SupervisedProc(300 /* initializeTimeout */, 500, 0, 0, 60000, 60000, 60000);
p.createProcess = () => fork(hang, [], { stdio: ["ignore", "ignore", "ignore", "ipc"] });

(async () => {
  await p.start();
  const outcome = await Promise.race([
    p.initialize().then(() => "resolved", (e) => "rejected: " + e.message),
    new Promise((r) => setTimeout(() => r("STILL PENDING AFTER 2s (WEDGED)"), 2000)),
  ]);
  console.log("initialize() outcome:", outcome); // 1.4.4/1.4.5: STILL PENDING AFTER 2s (WEDGED)
  process.exit(0);
})();
```

(Same scenario with a child that exits before its first message also wedges; with a child whose first message isn't `initializeResponse`, the throw happens after `clearTimeout` and `procWatchTask`'s empty catch proceeds to `await proc.join()` where `init` never settles — wedge with zero log lines.)

## Suggested fix

Make `initialize()` always settle and reclaim the child:
- race the first message against the child's `exit` event and the init timeout,
- on any failure: kill the child, reject `init`, and **throw**, so `procWatchTask`'s catch/finally release the mutex slots and the pool replenishes.

We are running exactly that as a local patch (diff below) in production; happy to open it as a PR if useful. Care is needed that a late-losing `once(proc, "exit")` rejection is pre-handled, otherwise every *normal* child exit after a successful init becomes an unhandledRejection.

```diff
   async initialize() {
     var _a;
     const timer = setTimeout(() => {
       this.init.reject(new Error("runner initialization timed out"));
+      try { this.proc?.kill("SIGKILL"); } catch {}
     }, this.#opts.initializeTimeout);
     if (!((_a = this.proc) == null ? void 0 : _a.connected)) {
-      this.init.reject(new Error("process not connected"));
-      return;
+      const err = new Error("process not connected");
+      this.init.reject(err);
+      clearTimeout(timer);
+      throw err;
     }
     this.proc.send({ case: "initializeRequest", value: { ... } });
-    await once(this.proc, "message").then(([msg]) => {
-      clearTimeout(timer);
-      if (msg.case !== "initializeResponse") {
-        throw new Error("first message must be InitializeResponse");
-      }
-    });
+    const firstMessage = once(this.proc, "message").then(([msg]) => {
+      if (msg.case !== "initializeResponse") throw new Error("first message must be InitializeResponse");
+    });
+    const exited = once(this.proc, "exit").then(() => {
+      throw new Error("process exited before initialization completed");
+    });
+    firstMessage.catch(() => {});
+    exited.catch(() => {}); // late race losers must not become unhandledRejection
+    try {
+      await Promise.race([firstMessage, exited, this.init.await]);
+    } catch (err) {
+      this.init.reject(err);
+      try { this.proc?.kill("SIGKILL"); } catch {}
+      throw err;
+    } finally {
+      clearTimeout(timer);
+    }
     this.init.resolve();
   }
```

## Environment

- `@livekit/agents` 1.4.4 (also inspected 1.4.5 — `supervised_proc.js` / `proc_pool.js` identical)
- Node.js 24.x, Linux (ECS/EC2), self-hosted workers (`apiKey`/`apiSecret`, no `workerToken`)
- `numIdleProcesses: 1`, `initializeProcessTimeout: 60s`; trigger in our case was memory pressure killing/stalling the warming child during a heavy import+VAD prewarm

tsushanth:
## Summary

In `@livekit/agents` ≥1.4.x, a warming child process that **dies or hangs before sending its first IPC message** permanently wedges the worker's process pool: the worker keeps reporting available and accepting `received job request`, but never launches another job until externally restarted.

I hit this against the production wedge described in livekit/agents-js#1748 — the reporter saw 5 wedge events across 2 days in a contact-center deployment, each black-holing every job routed to the worker for hours.

## Mechanism

1. `SupervisedProc.initialize()` only completes via `await once(proc, 'message')` ([supervised_proc.ts:217](<https://github.com/livekit/agents-js/blob/main/agents/src/ipc/supervised_proc.ts#L217>)). A child killed mid-prewarm (OOM, V8 heap abort, import crash) emits neither `'message'` nor `'error'`, so `once()` pends forever.
2. The `initializeProcessTimeout` timer only rejects the side `init` future — it does not kill the child and does not unblock `initialize()`.
3. `ProcPool.procWatchTask` is parked at `await proc.initialize()` ([proc_pool.ts:109](<https://github.com/livekit/agents-js/blob/main/agents/src/ipc/proc_pool.ts#L109>)) holding **both** `initMutex` and its `procMutex` slot. With `numIdleProcesses: 1` that is the only slot, so `run()` blocks at `procMutex.lock()` forever, `warmedProcQueue` never refills, and every accepted job blocks at `warmedProcQueue.get()` — unbounded, no timeout.
4. The worker's load loop is CPU-based, so the server keeps dispatching jobs into the black hole.

## Fix

`initialize()` now races three signals so it always settles:

1. **firstMessage** — the happy path (the same `once(proc, 'message')` as today)
2. **exit** — the child crashed/exited before `initializeResponse`
3. `this.init` — the timeout above (or any other init rejection)

On timeout, the child is SIGKILLed so the exit-race can settle. Late race losers (`firstMessage` after a successful resolve via `exit`, or `exit` after a successful first message) attach a no-op `.catch` so a normal post-init child exit does not surface as an unhandled rejection.

`ProcPool.procWatchTask` already has a `try { ... } catch {}` around `proc.initialize()` ([proc_pool.ts:108-118](<https://github.com/livekit/agents-js/blob/main/agents/src/ipc/proc_pool.ts#L108-L118>)) that releases the acquired slot on failure, so once `initialize()` throws, the pool replenishes as intended. The early-return when `!proc.connected` is also converted to a throw for the same reason.

A stale comment in `start()` ([supervised_proc.ts:99-102](<https://github.com/livekit/agents-js/blob/main/agents/src/ipc/supervised_proc.ts#L99-L102>)) said the run-catch intentionally avoids killing the child because killing would race the `once('message')` and deadlock `initialize()`. That race no longer exists; updated the comment.

## Test plan

- [X] Three init-timeout tests in [agents/src/ipc/supervised_proc.test.ts](<agents/src/ipc/supervised_proc.test.ts>):
  - **Existing** "slow init child" tests updated to assert `initialize()` now rejects (instead of silently resolving) and that `join()` still resolves.
  - **New** regression test for the livekit/agents-js#1748 wedge: a child that never sends a message and never exits. `initialize()` must reject within the timeout — confirmed it pends past 2 s on `main` and rejects in <200 ms after the fix.
- [X] `pnpm vitest run agents/src/ipc/supervised_proc.test.ts` — 16/16 pass (13 existing + 3 in this area). Verified the new test fails on `main` by reverting only the production change.
- [X] `pnpm build:agents` — clean
- [X] `pnpm lint` — no new warnings on touched files
- [X] `pnpm format:check` — clean
- [X] Changeset added (patch)

Closes livekit/agents-js#1748

### DIFF
--- a/.changeset/fix-supervised-proc-init-wedge.md
+++ b/.changeset/fix-supervised-proc-init-wedge.md
@@ -1,0 +1,22 @@
+---
+'@livekit/agents': patch
+---
+
+Fix `ProcPool` wedging permanently when a warming child process dies or
+hangs before its first IPC message.
+
+`SupervisedProc.initialize()` only completed via
+`await once(proc, 'message')`, which never settled when the child exited
+or crashed mid-prewarm (kernel OOM, V8 heap abort, import crash). The
+init timeout rejected the side `init` future but left `initialize()`
+itself pending, so `ProcPool.procWatchTask` was parked at
+`await proc.initialize()` holding both `initMutex` and its `procMutex`
+slot — the worker kept reporting available and accepting jobs that could
+never launch.
+
+`initialize()` now races three signals — first message, child `exit`,
+and the init timeout — and kills the child on timeout. Late race losers
+swallow their own rejection so a normal child exit after a successful
+init never surfaces as an unhandled rejection. The pool's
+`procWatchTask` already catches initialization failures, so its mutex
+slots release and the pool replenishes as intended.
--- a/agents/src/ipc/supervised_proc.test.ts
+++ b/agents/src/ipc/supervised_proc.test.ts
@@ -126,17 +126,19 @@ describe('IPC send on dead process', () => {
 });
 
 describe('init timeout rejection handling', () => {
-  it('does not produce unhandled rejection when init times out', async () => {
-    // Regression test: before the fix, run() was called without await in start().
-    // When init timed out, the rejection in run()'s `await this.init.await` escaped
-    // as an unhandled rejection — crashing the Node.js process.
+  it('rejects initialize() and produces no unhandled rejections when init times out', async () => {
+    // Regression test: before the fix, initialize() would either silently
+    // resolve (after a late first-message arrived) or — for a child that
+    // never sent a message — hang forever, wedging the pool. It must now
+    // reject so the caller can release its mutex slots. The Future's
+    // self-attached `.catch` and the race losers' attached `.catch` ensure
+    // no rejection escapes as unhandled.
     const unhandled: unknown[] = [];
     const handler = (reason: unknown) => unhandled.push(reason);
     process.on('unhandledRejection', handler);
 
-    // Child that responds AFTER the timeout — simulates slow init under CPU pressure.
-    // Timeout fires at 50ms (init.reject), child responds at 200ms (once() resolves).
-    // Before the fix, init.reject caused an unhandled rejection in run().
+    // Child that responds AFTER the timeout — simulates slow init under
+    // CPU pressure. Timeout fires at 50 ms, child would respond at 200 ms.
     const slowScript = join(tmpdir(), 'test_slow_init_child.mjs');
     writeFileSync(
       slowScript,
@@ -168,9 +170,7 @@ describe('init timeout rejection handling', () => {
     );
 
     await proc.start();
-    // initialize() returns normally: child responds at 200ms, once() resolves,
-    // but init was already rejected at 50ms — run() gets the rejection.
-    await proc.initialize();
+    await expect(proc.initialize()).rejects.toThrow(/timed out|exited before/);
 
     // Give the event loop a tick for any unhandled rejection to surface
     await new Promise((r) => setTimeout(r, 100));
@@ -185,8 +185,9 @@ describe('init timeout rejection handling', () => {
   });
 
   it('join() resolves after init timeout instead of hanging forever', async () => {
-    // When run() fails early (before registering proc event handlers),
-    // #join must still resolve so that join() and close() don't hang.
+    // When init fails (rejected), run() throws at `await this.init.await`,
+    // start()'s catch resolves #join, and join() must return promptly so
+    // that the pool can release its mutex slots and replenish.
     const slowScript = join(tmpdir(), 'test_slow_init_child_join.mjs');
     writeFileSync(
       slowScript,
@@ -210,7 +211,7 @@ describe('init timeout rejection handling', () => {
     const proc = new TestProc(50, 1000, 0, 0, 5000, 60000, 2500);
 
     await proc.start();
-    await proc.initialize();
+    await expect(proc.initialize()).rejects.toThrow();
 
     // join() must resolve within a reasonable time, not hang forever
     const result = await Promise.race([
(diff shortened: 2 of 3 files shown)


tsushanth:
Rebased onto current upstream.

tsushanth:
Closing — upstream already landed the equivalent fix on main (AbortController approach). Nothing left to contribute here.

RizqiSyahrendra:
> Closing — upstream already landed the equivalent fix on main (AbortController approach). Nothing left to contribute here.

@tsushanth hi, since you're closing the PR, may I know which version already fixed this issue? I still got "initialization timed out" at version 1.5.5

PR #1782, opened 2026-06-12
Fix voice handoff drain and playback stalls

toubatbrian:
Based on https://github.com/livekit/agents-js/pull/1732

## Description

Fixes several voice-pipeline edge cases that can leave an agent silent or stuck during task handoff, interruption, or reused STT pipelines. The main issue is that activity drain can wait forever on work that can no longer make progress, keeping scheduling/new-turn handling blocked.

## Changes Made

- Prevent `_pauseSchedulingTask()` from self-deadlocking when a drain is triggered from one of the activity's own speech tasks, or when drain is blocked only by done/interrupted speech tasks.
- Preserve the original STT input timestamp base when reusing an `STTPipeline` across handoff, so STT `endTime`-based timing does not drift after activity transfer.
- Fix audio forwarding resampling so sample-rate mismatches are handled after the first frame too, using the actual frame channel count.
- Fix playback segment accounting for interrupted-while-paused audio frames so dropped frames do not strand `waitForPlayout()`.
- Reconcile transcription synchronizer playback-counter drift when a downstream audio sink drops a frame before counting it.
- Add regression coverage for the drain, STT handoff timestamp, audio output segment accounting, and transcription synchronizer drift cases.

## Testing

- `npx vitest run agents/src/voice/agent_activity.test.ts agents/src/voice/agent_activity_handoff.test.ts agents/src/voice/room_io/_output.test.ts agents/src/voice/transcription/synchronizer.test.ts`
  - 4 test files passed
  - 61 tests passed
- `pnpm build:agents`
  - Build passed
  - Type declarations emitted
- `agents-cli` voice interruption smoke against `basic-agent-js`
  - Interruption detected
  - `getWeather` tool executed
  - Fresh assistant reply committed
  - Worker log scan clean: no `InvalidState`, unhandled rejection, scheduling-paused/user-input-drop messages

### DIFF
--- a/.changeset/clamp-stt-last-speaking-time.md
+++ b/.changeset/clamp-stt-last-speaking-time.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Clamp the STT-derived `lastSpeakingTime` to the current wall-clock time. When the STT stream's clock diverged from the activity's input epoch (e.g. a reused STT pipeline after an agent handoff), the transcript `endTime` could map to a timestamp minutes in the future, causing the end-of-turn bounce task to sleep that long before committing the user turn — the agent appeared to go silent mid-call even though LLM preemptive generation kept running.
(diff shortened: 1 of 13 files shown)


theomonnom (approved):

toubatbrian:
Need to include one more small fix

PR #1795, opened 2026-06-16
fix(avatar): normalise snake_case lk.playback_finished payload to avoid audio_end_ms=null

smorimoto:
## Summary

`DataStreamAudioOutput` decoded the `lk.playback_finished` RPC payload with a compile-time-only `as PlaybackFinishedEvent` cast. The LiveKit avatar DataStream protocol serialises that payload with **snake_case** keys (`playback_position`, `synchronized_transcript`), so the camelCase `playbackPosition` field read back `undefined`. On a user interruption that became `Math.floor(undefined * 1000) === NaN`, which `JSON.stringify` serialises as `null` inside `conversation.item.truncate`. The OpenAI Realtime API then rejected the truncate with an `invalid_type` error and the interrupted turn stalled.

This affects **every** avatar that uses `DataStreamAudioOutput` (Anam, Bey, D-ID, LemonSlice, Runway, Tavus, Trugen). `liveavatar` is unaffected because it tracks the playback position locally and never crosses the snake_case wire boundary.

## Root cause

1. The avatar worker invokes `lk.playback_finished`; the receiver casts the payload without remapping keys — `agents/src/voice/avatar/datastream_io.ts` (`JSON.parse(data.payload) as PlaybackFinishedEvent`).
2. The TS `PlaybackFinishedEvent` interface is camelCase — `agents/src/voice/io.ts` — so `playbackPosition` and `synchronizedTranscript` read `undefined` for a snake_case payload.
3. The interrupted realtime path reads the undefined field and overwrites the valid `0` default — `agents/src/voice/agent_activity.ts`.
4. `Math.floor(undefined * 1000)` is `NaN`, and the OpenAI plugin forwards it unguarded — `plugins/openai/src/realtime/realtime_model.ts` — where `JSON.stringify` turns `NaN` into `null`.

## Fix

- **Primary** — `DataStreamAudioOutput.handlePlaybackFinished` now normalises the wire payload through a new `parsePlaybackFinishedPayload` helper (snake_case primary, camelCase fallback, missing/`null`/non-finite position coerced to `0`). This also restores the previously-dropped `synchronizedTranscript` on interrupted turns.
- **Defence-in-depth** — a non-finite `audioEndMs` is now clamped to a valid non-negative integer in both `AgentActivity` and the OpenAI plugin's `truncate()`, so a malformed or absent playback position can never again serialise as `null`.

## Verification

Confirmed end-to-end against Anam's live avatar engine:

- The raw `lk.playback_finished` payload Anam emits on interruption is `{"playback_position": 2.0, "interrupted": true, "synchronized_transcript": null}` (snake_case), reproducing the original bug.
- Driving the **fixed** `DataStreamAudioOutput` through a real Anam session: `waitForPlayout().playbackPosition` is now `1.52` (finite), the computed `audio_end_ms` is `1520`, and the serialised event is `{"type":"conversation.item.truncate","audio_end_ms":1520}` — no longer `null`.

## Test plan

- [x] New unit test `playback_payload.test.ts` covers snake_case (incl. the exact captured Anam payload), camelCase fallback, missing/`null`/non-numeric position, and a legitimate zero.
- [x] `pnpm build` (full, Turborepo) passes.
- [x] Avatar and OpenAI realtime test suites pass.
- [x] ESLint and Prettier clean on changed files.

### DIFF
--- a/.changeset/fix-anam-avatar-truncate-null.md
+++ b/.changeset/fix-anam-avatar-truncate-null.md
@@ -1,0 +1,22 @@
+---
+'@livekit/agents': patch
+'@livekit/agents-plugin-openai': patch
+---
+
+Fix DataStream avatars (Anam, Bey, D-ID, LemonSlice, Runway, Tavus, Trugen) stalling the
+conversation on user interruption when paired with the OpenAI Realtime API.
+
+`DataStreamAudioOutput` parsed the `lk.playback_finished` RPC payload with a compile-time-only
+`as PlaybackFinishedEvent` cast. The LiveKit avatar protocol serializes that payload with
+snake_case keys (`playback_position`, `synchronized_transcript`) — confirmed against Anam's
+live engine, which emits `{"playback_position": 2.0, "interrupted": true, "synchronized_transcript": null}`
+— so the camelCase `playbackPosition` read back `undefined`. That became
+`Math.floor(undefined * 1000) === NaN`, which `JSON.stringify` serializes as `null` in
+`conversation.item.truncate`; the OpenAI Realtime API then rejected the truncate with an
+`invalid_type` error and the interrupted turn could not recover.
+
+`DataStreamAudioOutput` now normalizes the wire payload (snake_case primary, camelCase
+fallback), which also restores the previously-dropped `synchronizedTranscript` on interrupted
+turns. As defense-in-depth, the realtime truncate path now clamps a non-finite `audioEndMs` to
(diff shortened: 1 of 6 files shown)


smorimoto:
@davidzhao @chenghao-mou Could you review this?

smorimoto:
@toubatbrian Gentle ping

toubatbrian (approved):
Thanks for catching this!

PR #1818, opened 2026-06-17
Add scoped filler support to RunContext

toubatbrian:
## Summary
- Adds scoped `ctx.filler(...)` support for long-running tools using the Python filler behavior as reference.
- Waits for session idle before playing filler speech and resets the dwell timer on updates or speech activity.
- Adds focused RunContext tests covering scope lifecycle, idle gating, cancellation, and interruption behavior.

## Test plan
- `pnpm test agents/src/voice/run_context.test.ts`
- Pinned ESLint check on edited files
- `pnpm build:agents`

## Notes
- Full package lint is blocked in the nested worktree by ESLint plugin ambiguity.
- `api:check` is blocked by the existing API Extractor `export * as` limitation.
- `agents-cli` smoke was not run because there is no deterministic dispatchable filler probe in the repo yet.

### DIFF
--- a/agents/src/voice/run_context.ts
+++ b/agents/src/voice/run_context.ts
@@ -5,7 +5,12 @@ import { FunctionCall, FunctionCallOutput } from '../llm/chat_context.js';
 import type { Future } from '../utils.js';
 import type { AgentActivity } from './agent_activity.js';
 import type { AgentSession } from './agent_session.js';
-import type { SpeechHandle } from './speech_handle.js';
+import {
+  AgentSessionEventTypes,
+  type AgentStateChangedEvent,
+  type UserStateChangedEvent,
+} from './events.js';
+import { type SpeechHandle, isSpeechHandle } from './speech_handle.js';
 
 export type UnknownUserData = unknown;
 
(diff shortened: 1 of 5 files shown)


agents/src/voice/run_context.ts
```diff
@@ -197,3 +273,164 @@ function stringifyToolOutput(value: unknown): string {
...
+  }
+}
+
+async function waitUnlessAborted<T>(promise: Promise<T>, signal: AbortSignal): Promise<boolean> {
+  if (signal.aborted) return false;
+  const abortPromise = new Promise<false>((resolve) => {
+    signal.addEventListener('abort', () => resolve(false), { once: true });
+  });
+  const result = await Promise.race([promise.then(() => true), abortPromise]);
+  return result;
```
longcw: the abort listener is not removed when `waitForIdle()` wins the race?

longcw (approved):
left a comment, otherwise looks good to me. maybe also add an use case in the example?

agents/src/voice/run_context.ts
toubatbrian: fixed

PR #1819, opened 2026-06-17
Add object tool syntax compatibility

toubatbrian:
## Summary
- Allows legacy object-shaped tool registration (`{ toolName: llm.tool(...) }`) while preserving the new flat array syntax.
- Introduces anonymous function tool input normalization so object keys supply the required function tool name internally.
- Keeps Toolsets and provider tools array-only, since object notation can only name function tools.

## Test plan
- `pnpm exec vitest run agents/src/llm/tool_context.test.ts`
- `pnpm exec tsc -p test/tsconfig.tools-object-compat.json`
- `pnpm --filter @livekit/agents build:types`
- Prettier check on changed files
- Pinned ESLint check on changed files

## Notes
- Full package lint is blocked in the nested worktree by ESLint plugin ambiguity.
- `api:check` is blocked by the existing API Extractor `export * as` limitation.


agents/src/llm/tool_context.ts
```diff
@@ -175,19 +175,13 @@ export abstract class ProviderTool implements Tool {
   }
 }
 
-export interface FunctionTool<
-  Parameters extends JSONObject,
+export interface AnonFunctionTool<
+  Parameters extends JSONObject = JSONObject,
   UserData = UnknownUserData,
   Result = unknown,
-> extends Tool {
```
toubatbrian, line 182: why drop extends Tool here?

agents/src/llm/tool_context.ts
```diff
@@ -375,7 +388,22 @@ class ToolsetFactory extends Toolset {
  */
 export type ToolCtxInput<UserData = UnknownUserData> =
   | ToolContext<UserData>
-  | readonly ToolContextEntry<UserData>[];
+  | ToolContextInput<UserData>;
+
+export type ToolContextInput<UserData = UnknownUserData> =
```
toubatbrian: The naming is very confusing

agents/src/llm/tool_context.ts
toubatbrian: Restored the direct relationship here: `FunctionTool` now extends `Tool`. The anonymous object-map shape is now derived as `AnonFunctionTool = Omit<FunctionTool, "id" | "name">`, so the object shorthand can stay anonymous while named tools still satisfy the shared `Tool` contract.

agents/src/llm/tool_context.ts
toubatbrian: Renamed the input shapes by role: `ToolContextLike` for APIs that accept an existing context or init data, `ToolContextInit` for constructor/update data, and `ToolDefinitionMap` for the object shorthand. Also renamed `normalizeToolContextInput` to `normalizeToolContextInit` and updated plugin consumers.

toubatbrian:
Updated the PR in `4475c239`.

- `FunctionTool` now extends `Tool`; anonymous object-map tools are derived by omitting `id`/`name`.
- Renamed the confusing input types to `ToolContextLike`, `ToolContextInit`, and `ToolDefinitionMap`.
- Added a patch changeset and updated remaining provider/plugin consumers.
- Updated `examples/src/basic_agent.ts` to exercise object tool shorthand and emit a `debug_message` payload from `getWeather`.

Verification:
- `pnpm build`
- Voice E2E via local `agents-cli`: `basic-agent-js` resolved `{function_tools_executed(.function_calls[0].name="getWeather") and function_tools_executed(.function_call_outputs[0].output=~"sunny") and debug_message(.payload.tool="getWeather") and debug_message(.payload.surface="object-tool-shorthand") and conversation_item_added(.item.message.role="ASSISTANT")}[60s]` in 9384ms.

agents/src/llm/tool_context.ts
```diff
@@ -176,7 +176,7 @@ export abstract class ProviderTool implements Tool {
 }
 
 export interface FunctionTool<
-  Parameters extends JSONObject,
+  Parameters extends JSONObject = JSONObject,
   UserData = UnknownUserData,
   Result = unknown,
```
u9g, line 191: will `Result` always be `JSONObject`? It seems like it would be since it's part of a tool signature that will be serialized and sent to an LLM.

u9g: *not a blocker

u9g (changes requested):
I made a PR with my comments on the PR https://github.com/livekit/agents-js/pull/1823

(removes the type exports from the comment, cleans up the types, rollback changes to basic_agent, and remove some of the overloads on `tool()`)

agents/src/llm/tool_context.ts
toubatbrian: I checked the execution path. `Result` is the runtime return type from `execute()`, not the parameter schema/signature advertised to the LLM. It can be a primitive (`string`, `number`, `boolean`), arrays/maps/objects that pass runtime validation, or special values like handoff returns; the framework then serializes the accepted output into `FunctionCallOutput.output`. So `JSONObject` would be too narrow here. If this ever gets tightened, it should be a broader serializable-output type, not just object JSON.

u9g: Makes sense, we can leave it for now then.

u9g (approved):

PR #1822, opened 2026-06-18
refactor(testing): use `using` for withMockTools

u9g:
## Description

Switches `withMockTools` from a curried callback runner to a `Disposable` used with a `using` declaration, scoping mocks to a block instead of a callback.

## Changes Made

- Replace `AsyncLocalStorage` backing with a module-level `activeMockTools` binding; `withMockTools` now returns `{ [Symbol.dispose] }` that restores the prior value on block exit (preserves nesting/override semantics).
- Update unit tests and the drive-thru failure test to use `using` blocks.
- Import `ReadableStream` from `node:stream/web` in the test to fix a type mismatch with `performToolExecutions`.

## Pre-Review Checklist

## Testing

- [x] All tests pass

## Additional Notes

Targets `claude/port-tool-mocking-js-i2Fpr`.

### DIFF
--- a/agents/src/voice/testing/run_result.test.ts
+++ b/agents/src/voice/testing/run_result.test.ts
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: 2026 LiveKit, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
+import { ReadableStream } from 'node:stream/web';
 import { describe, expect, it } from 'vitest';
 import { z } from 'zod';
 import { FunctionCall } from '../../llm/chat_context.js';
--- a/agents/src/voice/testing/run_result.test.ts
+++ b/agents/src/voice/testing/run_result.test.ts
@@ -8,7 +8,7 @@ import { ToolContext, tool } from '../../llm/tool_context.js';
 import { Agent } from '../agent.js';
 import { performToolExecutions } from '../generation.js';
 import { SpeechHandle } from '../speech_handle.js';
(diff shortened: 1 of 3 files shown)


agents/src/voice/testing/run_result.test.ts
```diff
@@ -23,53 +24,53 @@ class AgentB extends Agent {
...
 
-    await withMockTools(AgentA, { tool1: mock })(async () => {
-      const store = mockToolsStorage.getStore();
-      expect(store).toBeDefined();
-      expect(store?.get(AgentA)?.tool1).toBe(mock);
-    });
+    {
+      using _mock = withMockTools(AgentA, { tool1: mock });
+      expect(activeMockTools).toBeDefined();
+      expect(activeMockTools?.get(AgentA)?.tool1).toBe(mock);
```
toubatbrian, line 33: This is great! One thing I’m a bit worried about is that `using` declarations were introduced in TypeScript 5.2. Although 5.2 was released pretty early, users on TypeScript versions below 5.2 still won’t be able to use this syntax.

Should we make both approaches work? Maybe we can support the non-using pattern as the broadly compatible path, and encourage users to upgrade TypeScript in the doc?

toubatbrian: I don’t feel super strongly about this since TS 5.2 has been out for a while, so most users are probably fine. Mainly wanted to flag it in case we care about supporting older TS versions in the examples. I’ll leave the decision to you.

toubatbrian (approved):

PR #1829, opened 2026-06-18
fix: make api:check pass across the workspace

toubatbrian:
## Summary

`pnpm api:check` was failing for every package and was never wired into CI, so the breakage went unnoticed. This makes it a working **structural API gate** across the whole workspace (62/62 packages pass).

Root causes were:

1. **`@microsoft/api-extractor` was years out of date** (`7.43.7`, bundling TypeScript 5.4.2 while the repo uses 5.9.3). That engine hard-errored on the `export * as ns from '...'` syntax in `agents/src/index.ts`.
2. **Cross-package namespace resolution** — a plugin doing `extends voice.AvatarSession` makes the extractor follow `@livekit/agents`'s `export * as voice` namespace, which it couldn't resolve (`ts.SyntaxKind.SourceFile declaration ... not supported`).
3. **`bey` and `liveavatar`** had an `api:check` script but no `api-extractor.json`.
4. **Warnings-as-errors noise** from API Extractor conventions the repo doesn't adopt (release tags, etc.), plus a report-diff workflow that was never real (`.gitignore` has `*.md`; only 2 reports were force-added; not in CI).

## Changes

- **Upgrade `@microsoft/api-extractor` to `^7.58.9`** across all packages (+ lockfile). New engine bundles TS 5.9.3 and supports `export * as ns` natively.
- **`api-extractor-shared.json`:**
  - `bundledPackages: ["@livekit/agents", "@livekit/agents-plugin-*"]` so namespace re-exports (`voice`, `llm`) resolve across package boundaries.
  - `apiReport.enabled: false` — this repo doesn't track `etc/*.api.md` reports, so `api:check` validates that the public API type-checks and rolls up cleanly rather than diffing a committed report.
  - Silence lint-style advice the repo doesn't use (`ae-missing-release-tag`, `ae-forgotten-export`, `ae-internal-missing-underscore`, doc/link advice, …). Hard errors (unresolvable types, the namespace error, missing entry points) still fail.
- **Add `api-extractor.json`** for the `bey` and `liveavatar` plugins.
- **Remove the two stale committed reports** (`perplexity`, `soniox`) now that report comparison is off.

No source files are touched — this is purely a dev-tooling/config fix.

## Test plan

- [x] `pnpm api:check` → **62/62 successful**
- [x] No source changes; full build runs as part of `api:check`'s `^build` and succeeds
- [ ] (optional follow-up) wire `api:check` into CI now that it passes

## Notes

- No changeset added: this only changes a `devDependency` (`@microsoft/api-extractor`) and dev config, so there's nothing user-facing to release.
- Base is `1.5.0` (the active integration branch), not `main`.


api-extractor-shared.json
```diff
@@ -60,7 +60,11 @@
    * This would direct API Extractor to embed those types directly in the .d.ts rollup, as if they had been
    * local files for library1.
    */
-  "bundledPackages": [],
+  // Workspace packages are bundled so API Extractor can resolve the namespace
+  // re-exports (e.g. `voice`, `llm`) that plugins consume from `@livekit/agents`.
+  // Without this, API Extractor errors with "ts.SyntaxKind.SourceFile declaration
+  // which is not (yet?) supported" when a plugin references those namespaces.
+  "bundledPackages": ["@livekit/agents", "@livekit/agents-plugin-*"],
```
lukasIO, line 67: nice workaround!

api-extractor-shared.json
```diff
@@ -134,8 +138,12 @@
   "apiReport": {
     /**
      * (REQUIRED) Whether to generate an API report.
+     *
+     * Disabled: this repo gitignores `*.md` and does not track per-package
+     * `etc/*.api.md` reports, so `api:check` is used purely to validate that the
+     * public API type-checks and rolls up cleanly (not to diff a committed report).
      */
-    "enabled": true
+    "enabled": false
```
lukasIO, line 146: why wouldn't we want to enable this?

api-extractor-shared.json
```diff
@@ -390,17 +398,28 @@
...
-        // "addToApiReportFile": false
-      }
-
-      // "ae-extra-release-tag": {
-      //   "logLevel": "warning",
-      //   "addToApiReportFile": true
-      // },
-      //
-      // . . .
+        "logLevel": "none"
```
lukasIO, line 409: this seems a bit excessive with all the logs underneath also be setting to none. 

What's your thinking behind it?

api-extractor-shared.json
toubatbrian: Thanks! 😄 Took a while to find — the older api-extractor just hard-errored on `export * as ns`, the newer one resolves it once the workspace packages are bundled.

api-extractor-shared.json
toubatbrian: Good question. I disabled it mainly because of how it interacts with `bundledPackages`: with the namespace bundling, each *plugin* report would inline all of `@livekit/agents`'s API into its `.api.md`, so every plugin report becomes huge and churns on every core change. Combined with the repo gitignoring `*.md` (only 2 reports were ever force-added) and api:check not being in CI, it felt like the report wasn't really being used.

That said, I think it's genuinely valuable for the **core `@livekit/agents` package** — it doesn't bundle anything, so its report stays clean and gives a real public-API diff. Happy to enable it just for core (commit `agents/etc/agents.api.md` + a `!etc/*.api.md` gitignore exception) and keep plugins structural-only. Want me to do that?

api-extractor-shared.json
toubatbrian: Fair — it is heavy-handed. Two parts:

1. The explicit per-message list isn't redundant with `default`: api-extractor ships built-in per-message defaults (`ae-forgotten-export`, `ae-missing-release-tag`, etc. default to `warning`) that take precedence over the user's `default` entry, so `default: none` alone doesn't silence them — I had to set each one explicitly (verified empirically).
2. You're right that it silences more than ideal. The release-tag family + unresolved-link/inheritdoc are conventions this repo deliberately doesn't use, so those are safe to drop. The one I'd rather keep is `ae-forgotten-export` (catches types used in the public API but not exported) — but the core currently has ~4 (`OutputSchema`, `AgentConstructor`, `PromptTemplate`, `UpdatePromptArgs`), so keeping it as a warning means exporting those, which is a small public-API change. 

Happy to trim to just the release-tag + link families and fix those 4 forgotten exports if you'd prefer the gate to stay meaningful — let me know.

api-extractor-shared.json
lukasIO: ok, I think it makes sense to stay disabled for now. I had opened a PR to fix the underlying issue of api-extractor with the namespace export here https://github.com/microsoft/rushstack/pull/5810

Once that gets merged we can revert the `bundledPackages` workaround and re-enable here as well.

turbo.json
```diff
@@ -121,7 +121,7 @@
     },
     "api:check": {
       "cache": false,
-      "dependsOn": ["^build"]
+      "dependsOn": ["build"]
```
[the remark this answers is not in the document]
lukasIO: @toubatbrian this seems valid

lukasIO (approved):

api-extractor-shared.json
lukasIO: yeah, forgotten export sounds useful to keep

PR #1854, opened 2026-06-22
Fix 1.5.0 voice-pipeline regressions vs main

toubatbrian:
## Summary

Autonomous regression hunt across the 1.5.0 voice / realtime / tool-call / interruption pipeline vs `main` (`git diff main...1.5.0`, with `merge-base(main, 1.5.0) == main` tip, so the diff cleanly isolates 1.5.0's changes). One behavioral fix is applied; the rest of the prioritized diffs were classified as intentional 1.5.0 changes, enhancements, or bug fixes — not regressions.

## The fix: `function_tools_executed` dropped on interruption

**Symptom (seed):** a realtime tool call's `function_tools_executed` event is dropped when the generation is interrupted (Phonic self-interrupting multi-step cascade).

**What I found:** the emit gating is *identical* across `main`, `1.5.0`, **and** the Python oracle. In all three, both the realtime (`_realtimeGenerationTaskImpl`) and pipeline tool paths `return` on `speechHandle.interrupted` **before** the `emit(FunctionToolsExecuted, …)`:
- JS `main` realtime: interrupt-return at `agent_activity.ts:3534`, emit at `:3607`.
- JS `1.5.0` realtime: interrupt-return at `agent_activity.ts:3620`, emit at `:3693`.
- Python realtime: interrupt-return at `agent_activity.py:3654`, emit at `:3708`.

So this is **not** a `main`→`1.5.0` code divergence (the seed's "on main the emit is unconditional" does not match `git show main:agents/src/voice/agent_activity.ts`). It's a long-standing behavior in all implementations: a tool that finishes *just before* a self-interrupt has its `function_tools_executed` event dropped.

**The change (implements the requested behavior as a safe, additive hardening):** emit `function_tools_executed` for already-completed tools from the interrupted branches too — emitted **before** the in-flight tool execution is cancelled, so only genuinely-completed outputs are reported. It is observability-only: interrupted turns still skip the tool reply, chat-context update, and agent handoff (those stay gated on the non-interrupted path). Applied symmetrically to the realtime and pipeline paths via a small `emitInterruptedToolExecutions` helper.

> Note: this is a deliberate divergence from Python parity (Python currently drops the event on interrupt). It is additive and flow-neutral. If parity is preferred over this hardening, revert is a one-block change.

## Test plan

- [x] New deterministic unit tests in `agent_activity.test.ts` (`emitInterruptedToolExecutions`): emits for a tool completed before interruption; no emit when no tool finished.
- [x] `pnpm build:agents` passes.
- [x] Full `agent_activity.test.ts` suite passes (18/18).
- [x] Prettier formatting passes on edited files.
- Pre-existing/environmental (NOT introduced here): `amd.test.ts` fails 11/11 on the clean 1.5.0 base too; fresh-worktree `eslint` hits the duplicate `@typescript-eslint` plugin env error; `api:check` fails pre-existingly (`export * as`).

## Regressions NOT found

The other prioritized diffs were reviewed and classified as intentional, not regressions: ToolContext→class migration (phonic, chat_context, fallback_adapter), the new `ToolExecutor`/async-toolset feature (`tool_executor.ts`, `speech_handle.ts` circular-wait guard, `agent_session.ts` session tools), `maxToolSteps`+1 final-response enhancement, `audio_recognition.ts` metrics refactor + out-of-order guard, `agent_session.emit` record-only-when-recording memory fix, and `transcription/synchronizer.ts` (a transcript-drop *fix* in 1.5.0). The tool-output capture on abort (`generation.ts` `waitUntilAborted` → `toolCompleted`) is unchanged from main.

### DIFF
--- a/.changeset/emit-interrupted-tool-executions.md
+++ b/.changeset/emit-interrupted-tool-executions.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Emit `function_tools_executed` for tool calls that finished before a speech interruption. Both the realtime and pipeline tool-execution paths `return` on `speechHandle.interrupted` ahead of their normal emit site, so a tool whose output landed just before a self-interrupt (e.g. Phonic's self-interrupting multi-step cascade) had its result dropped from observability even though it executed. The event is now emitted from the interrupted branches as well — before the in-flight tool execution is cancelled, so only genuinely-completed outputs are reported. This is observability-only and does not change conversation flow: interrupted turns still skip the tool reply, chat-context update, and agent handoff.
--- a/agents/src/voice/agent_activity.test.ts
+++ b/agents/src/voice/agent_activity.test.ts
@@ -16,13 +16,20 @@
  */
 import { Heap } from 'heap-js';
 import { describe, expect, it, vi } from 'vitest';
-import { AgentConfigUpdate, ChatContext } from '../llm/chat_context.js';
+import {
+  AgentConfigUpdate,
+  ChatContext,
+  FunctionCall,
+  FunctionCallOutput,
+} from '../llm/chat_context.js';
 import { LLM, type LLMStream } from '../llm/llm.js';
 import { type Tool, ToolContext, ToolFlag, Toolset, tool } from '../llm/tool_context.js';
 import { Future, Task } from '../utils.js';
 import { _getActivityTaskInfo } from './agent.js';
 import { AgentActivity } from './agent_activity.js';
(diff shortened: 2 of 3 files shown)


toubatbrian:
Closing — investigation showed this is not a main→1.5.0 regression (the interrupt-before-emit gating is identical across JS main, JS 1.5.0, and Python). Emitting the event on interrupt would diverge from Python parity, which we want to avoid, so 1.5.0 keeps the existing parity behavior.

PR #1867, opened 2026-06-24
fix: use err.name check for AbortError instead of instanceof Error

### LINKED ISSUE #1712: EOU detection AbortError logged at error level - abort guard misses DOMException (instanceof Error + "This" vs "The")
### Summary

In `voice/audio_recognition.ts`, benign `AbortError`s from the EOU (end-of-utterance) detection task on normal turn/session teardown are logged at **error** level as `Error in EOU detection task:`, and the error serializes as a `DOMException` numeric constant table (`{ ABORT_ERR: 20, DATA_CLONE_ERR: 25, ... }`) rather than a useful message. This produces error-level log noise on essentially every turn.

### Root cause

The `bounceEOUTask` result handler tries to swallow aborts:

```js
this.bounceEOUTask.result.then(() => {
  this.logger.debug("EOU detection task completed");
}).catch((err) => {
  if (err instanceof Error && err.message.includes("This operation was aborted")) {
    return;
  }
  this.logger.error(err, "Error in EOU detection task:");
});
```

The abort raised by `AbortController`/`AbortSignal` is a **`DOMException` with `name === 'AbortError'`**, and under Node/Bun:

1. A `DOMException` is **not** `instanceof Error`, so the first condition is already false.
2. Its message is **`"The operation was aborted."`** (note "The", not "This"), so the `includes("This operation was aborted")` substring check would miss it even if `instanceof Error` passed.

Both conditions fail, so the expected abort falls through to `this.logger.error(...)`. Because the value is a `DOMException` (not a normal `Error`), the structured logger emits its enumerable constant table instead of `name`/`message`.

### Environment

- `@livekit/agents` 1.3.4
- Node/Bun runtime (observed on Bun 1.2.23, Linux x64)

### Suggested fix

Detect aborts by name, which is robust across runtimes and message wording:

```js
}).catch((err) => {
  if (err?.name === 'AbortError') {
    return;
  }
  this.logger.error(err, "Error in EOU detection task:");
});
```

(Optionally also normalize non-`Error` throwables before logging so any genuinely unexpected `DOMException` still logs a readable `name`/`message`.)

tsushanth:
Fixes #1712

`DOMException`, which is what `AbortSignal`/`AbortController` throws, is not `instanceof Error` in Node.js or Bun. As a result, guards written as:

```ts
if (err instanceof Error && err.name === 'AbortError')
```

never match a real abort. The `instanceof Error` condition is already false for `DOMException`, so the abort falls through to the error-level logger — producing the spurious `DOMException` constant table in logs on every normal turn teardown.

The `.name` property is not exclusive to `Error` subclasses; any thrown value can carry it. Dropping `instanceof Error` and using an optional-chaining cast:

```ts
if ((err as { name?: string })?.name === 'AbortError')
```

matches both `Error` subclasses and `DOMException` uniformly across runtimes, which is the same approach the Web Platform (and Node's own `AbortSignal` docs) recommends.

Six occurrences fixed across the agents package and the Google, ElevenLabs, and OpenAI plugins.

### DIFF
--- a/agents/src/llm/fallback_adapter.ts
+++ b/agents/src/llm/fallback_adapter.ts
@@ -241,7 +241,7 @@ class FallbackLLMStream extends LLMStream {
       }
 
       // Handle timeout errors
-      if (error instanceof Error && error.name === 'AbortError') {
+      if ((error as { name?: string })?.name === 'AbortError') {
         if (checkRecovery) {
           this._log.warn({ llm: llm.label() }, 'recovery timed out');
         } else {
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -1608,7 +1608,7 @@ export class AgentActivity implements RecognitionHooks {
     const onAbort = () => waitInactiveTask.cancel();
     signal.addEventListener('abort', onAbort, { once: true });
     const waitInactiveResult = waitInactiveTask.result.catch((error) => {
-      if (error instanceof Error && error.name === 'AbortError') {
+      if ((error as { name?: string })?.name === 'AbortError') {
         return;
       }
       throw error;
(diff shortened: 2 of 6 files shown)


tsushanth:

PR #1872, opened 2026-06-24
fix(voice): capture activity to local before awaits in closeImplInner

### LINKED ISSUE #1871: AgentSession.closeImplInner throws "Cannot read properties of undefined (reading 'currentSpeech')" when close races an activity transition
## Summary

`AgentSession.closeImplInner` reads `this.activity` across multiple `await` boundaries without capturing a local reference. If `session.close()` races with an activity transition that sets `this.activity = undefined` (e.g. an agent/sub-agent handoff or activity swap), `this.activity` becomes `undefined` mid-method and the `currentSpeech` access throws:

```
TypeError: Cannot read properties of undefined (reading 'currentSpeech')
    at AgentSession.closeImplInner (.../voice/agent_session.js:968)
        await this.activity.currentSpeech?.waitForPlayout();
    at AgentSession.closeImpl (.../voice/agent_session.js:945)
    at AgentSession.close (.../voice/agent_session.js:756)   // CloseReason.USER_INITIATED
    at .../ipc/job_proc_lazy_main.js:135                      // waitUntilTimeout(sessionClosePromise, SESSION_CLOSE_TIMEOUT)
```

## Version

`@livekit/agents` **1.4.8**. Confirmed the same pattern is present in **1.4.9** and **`main`** (the close path still reads `this.activity` across awaits without a local capture).

## Root cause

In `closeImplInner` (compiled excerpt, 1.4.8):

```js
if (this.activity) {
  if (!drain) {
    try { await this.activity.interrupt({ force: true }).await; } catch (e) { ... }
  }
  await this.activity.drain();                       // <-- await
  await this.activity.currentSpeech?.waitForPlayout();// <-- this.activity may now be undefined -> throws
  if (reason !== CloseReason.ERROR) {
    this.activity.commitUserTurn({ ... });
  }
  try { this.activity.detachAudioInput(); } catch (e) {}
}
...
this.emit(AgentSessionEventTypes.Close, createCloseEvent(reason, error));  // never reached on throw
```

`this.activity` is set to `undefined` in the activity-update path (`this.activity = void 0; this.activity = this.nextActivity;`). When a `close()` is in flight and an activity transition runs concurrently during one of the `await`s above, the guard `if (this.activity)` has already passed but `this.activity` is now `undefined`, so `this.activity.currentSpeech` throws.

Because the throw happens **before** `this.emit(Close, ...)`, the `Close` event is never emitted — so any application logic that depends on the `Close` event (we use it to classify why a call ended) silently never runs for that session.

## Reproduction conditions

- Agent/sub-agent **handoff** (or any activity swap) in progress, and
- `session.close(CloseReason.USER_INITIATED)` triggered around the same time (in our case `ctx.shutdown()` on call end).

It's timing-dependent, so not 100% reproducible, but recurs in production under normal call-end + handoff overlap.

## Suggested fix

Capture the activity in a local before the awaits and operate on that (or make the accesses null-safe), so a concurrent transition can't turn `this.activity` `undefined` mid-method:

```js
const activity = this.activity;
if (activity) {
  if (!drain) { try { await activity.interrupt({ force: true }).await; } catch (e) { ... } }
  await activity.drain();
  await activity.currentSpeech?.waitForPlayout();
  if (reason !== CloseReason.ERROR) activity.commitUserTurn({ ... });
  try { activity.detachAudioInput(); } catch (e) {}
}
```

Source: `agents/src/voice/agent_session.ts` — `closeImplInner`.

Happy to send a PR if useful.

tsushanth:
Fixes #1871

`closeImplInner` reads `this.activity` across multiple `await` boundaries without first capturing a local reference. When `close()` races an activity transition that sets `this.activity = undefined` between the initial `if (this.activity)` guard and the subsequent awaited calls, accesses like `this.activity.currentSpeech?.waitForPlayout()` throw a `TypeError`:

```
TypeError: Cannot read properties of undefined (reading 'currentSpeech')
    at AgentSession.closeImplInner (voice/agent_session.js:968)
```

Because the throw happens before `this.emit(Close, ...)`, the `Close` event is never emitted — so any application code that depends on the session close event silently never runs for those sessions.

The fix captures `this.activity` into a local `const activity` at the top of the block. All reads within the `if (activity)` branch then operate on the captured reference, which remains stable regardless of concurrent mutations to `this.activity`.

### DIFF
--- a/agents/src/voice/agent_session.ts
+++ b/agents/src/voice/agent_session.ts
@@ -1526,25 +1526,26 @@ export class AgentSession<
     this._onAecWarmupExpired();
     this.off(AgentSessionEventTypes.UserInputTranscribed, this._onUserInputTranscribed);
 
-    if (this.activity) {
+    const activity = this.activity;
+    if (activity) {
       if (!drain) {
         try {
-          await this.activity.interrupt({ force: true }).await;
+          await activity.interrupt({ force: true }).await;
         } catch (error) {
           this.logger.warn({ error }, 'Error interrupting activity');
         }
       }
 
-      await this.activity.drain();
+      await activity.drain();
       // wait any uninterruptible speech to finish
-      await this.activity.currentSpeech?.waitForPlayout();
+      await activity.currentSpeech?.waitForPlayout();
 
       if (reason !== CloseReason.ERROR) {
-        this.activity.commitUserTurn({ audioDetached: true, throwIfNotReady: false });
+        activity.commitUserTurn({ audioDetached: true, throwIfNotReady: false });
       }
 
       try {
-        this.activity.detachAudioInput();
+        activity.detachAudioInput();
       } catch (error) {
         // Ignore detach errors during cleanup - source may not have been set
       }


tsushanth:
recheck

PR #1878, opened 2026-06-25
fix: drain unused audio tee branches to prevent unbounded memory growth

### LINKED ISSUE #1462: AudioRecognition leaks all inbound audio frames when using realtime_llm turn detection without local VAD/STT
edit: dang usually claude labels itself -- this issue was fired up by Claude Code

`AudioRecognition` unconditionally `tee()`s the primary audio input stream into `vadInputStream` and `sttInputStream` branches, even when no VAD or STT consumer is configured. When using `realtime_llm` turn detection with a `RealtimeModel` (no local VAD or STT), both branches go unread and their internal buffers grow for the lifetime of the call.

**Impact:** ~300 KB/s RSS growth per active call. A 5-minute call retains ~90 MB; an 8-minute call ~150 MB. Memory drops cleanly on call end (not a cross-call leak), but under concurrent load this exhausts worker memory.

### Heap snapshot evidence

Comparison view, 30s baseline → 300s (single call, `@livekit/agents` 1.4.0):

| Constructor | # New | # Freed | # Delta | Size Delta |
|---|---|---|---|---|
| ArrayBuffer | 27,164 | 0 | +27,164 | +24.3 MB |
| AudioFrame | 27,164 | 0 | +27,164 | +1.7 MB |
| Int16Array | 27,164 | 0 | +27,164 | +1.3 MB |

27,164 frames / 270s = ~100 fps, matching rtc-node delivering 10ms chunks at 24kHz. Zero freed — every frame from the window is retained.

Retainer chain for a retained `AudioFrame`:

```
AudioFrame → {value, done: false} → Deque (auto-grown to 32,768 slots)
  → ReadableStreamDefaultController → ReadableStream
    → vadInputStream in AudioRecognition
      → audioRecognition in AgentActivity → AgentSession (GC root)
```

### Config that triggers it

```ts
  llm: new openai.realtime.RealtimeModel({ /* ... */ }),
  // No vad — server handles turn detection
  // No stt — realtime model handles transcription
});
```

With `turnDetection: 'realtime_llm'`. This is a valid, supported config — `AgentActivity.resolveInterruptionDetector()` accepts it without warning.

### Root cause

`audio_recognition.ts` lines 309–328 — both branches of the `if (interruptionDetection)` block create `vadInputStream` via `tee()` regardless of whether a VAD consumer exists. `createVadTask()` bails immediately with `if (!vad) return;`, so the branch is never drained. Same pattern for `sttInputStream` — `startSttTasks()` bails with `if (!this.stt) return;`, leaving the merged stream unread.

`mergeReadableStreams` eagerly reads from its inputs and enqueues into the output controller, so even though `primaryInputStream` is consumed by the merge, the output queue grows unbounded when nobody reads from `sttInputStream`.

### Related

- PR #570 introduced the current `tee()` architecture. It considered "realtime_llm requested but model lacks turn-detection capability" but not "realtime_llm valid + no VAD/STT consumers."
- PR #1201 (closed) attempted a broader `ReadableStream → Chan` migration, partly motivated by these backpressure issues.

### Separately: Promise.race reaction leak in `realtime_model.ts`

`RealtimeSession.forwardEvents()` calls `Promise.race([messageChannel.get(), abortFuture.await])` in a while loop. Each iteration attaches a new reaction record to the long-lived `abortFuture.await` promise, retaining the resolved event (including base64 audio strings). Adds ~17 MB per 5-min call. `Queue.get()` already accepts `{ signal }` — using that instead eliminates the leak.

### Fix

We have a working fix at [`jakswa/agents-js:fix/drain-unused-audio-streams`](https://github.com/jakswa/agents-js/compare/fix/drain-unused-audio-streams) (2 files, +25/−9 lines): skip the `tee()` when no consumer exists for a branch, drain `primaryInputStream` when neither VAD nor STT is configured, and replace `Promise.race` with `Queue.get({ signal })`. Feel free to cherry-pick, rewrite, or ignore — just want the fix to land in some form.

tsushanth:
When no VAD or STT consumer is configured — the common case with realtime LLM turn detection — `audio_recognition.ts` unconditionally calls `.tee()` on the primary input stream and assigns both branches to `vadInputStream` and `sttInputStream`. Neither branch is ever read, so the broadcast `TransformStream` queues every incoming audio frame with no backpressure release. In a sustained session this produces roughly 300 KB/s of RSS growth until the process OOMs or the connection drops.

The fix makes `.tee()` conditional on which consumers are actually present. When both VAD and STT are configured the stream is tee'd into both branches as before. When only VAD is present the stream is tee'd once. When only STT is present the primary stream is passed directly without any tee. When neither consumer exists a lightweight background reader drains the primary stream so the upstream broadcast transform stays unblocked without accumulating frames.

The corresponding null guards in `forwardInputAudioToStt` and `createVadTask` let TypeScript enforce the invariant and provide an explicit early-return path for callers that race before the stream is established.

A second, smaller leak lives in `RealtimeSession.forwardEvents` in the OpenAI realtime plugin. Each call allocated a `Future`, registered an `'abort'` listener on the signal, and then raced the channel get against that future. Because `Queue.get` already accepts an `AbortSignal` directly, the `Future` and its listener were redundant and kept the listener registered on the signal until the channel's pending promise resolved — preventing GC of the closure in long-lived sessions. The fix passes the signal into `Queue.get` directly and removes the auxiliary future entirely.

Fixes #1462

### DIFF
--- a/.changeset/fix-audio-tee-leak.md
+++ b/.changeset/fix-audio-tee-leak.md
@@ -1,0 +1,5 @@
+---
+"@livekit/agents": patch
+---
+
+fix: drain unused audio tee branches to prevent unbounded RSS growth when no VAD/STT consumer is configured
--- a/agents/src/llm/fallback_adapter.ts
+++ b/agents/src/llm/fallback_adapter.ts
@@ -241,7 +241,7 @@ class FallbackLLMStream extends LLMStream {
       }
 
       // Handle timeout errors
-      if (error instanceof Error && error.name === 'AbortError') {
+      if ((error as { name?: string })?.name === 'AbortError') {
         if (checkRecovery) {
           this._log.warn({ llm: llm.label() }, 'recovery timed out');
         } else {
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -1608,7 +1608,7 @@ export class AgentActivity implements RecognitionHooks {
     const onAbort = () => waitInactiveTask.cancel();
     signal.addEventListener('abort', onAbort, { once: true });
     const waitInactiveResult = waitInactiveTask.result.catch((error) => {
-      if (error instanceof Error && error.name === 'AbortError') {
+      if ((error as { name?: string })?.name === 'AbortError') {
         return;
       }
       throw error;
--- a/agents/src/voice/audio_recognition.ts
+++ b/agents/src/voice/audio_recognition.ts
@@ -336,8 +336,8 @@ export class AudioRecognition {
   // provider's logs for debugging.
   private sttRequestIds: string[] = [];
 
-  private vadInputStream: ReadableStream<AudioFrame>;
-  private sttInputStream: ReadableStream<AudioFrame>;
+  private vadInputStream: ReadableStream<AudioFrame> | null = null;
+  private sttInputStream: ReadableStream<AudioFrame> | null = null;
   /**
    * Active subscriber writers fed from {@link subscribersBroadcast}. Each
    * {@link subscribeAudioStream} call appends one entry; entries are dropped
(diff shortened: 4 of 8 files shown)


tsushanth:

PR #1887, opened 2026-06-26
fix(rime): send language parameter for TTS

chasef07:
## Summary

- Add `language` as the preferred Rime TTS language option while keeping `lang` as a deprecated compatibility alias.
- Send the documented Rime API parameter as `language` for both HTTP and WebSocket TTS requests.
- Add unit coverage for HTTP payloads, legacy `lang` fallback, and WebSocket query parameters.

## Why

The Rime API expects the language selection under the `language` parameter. The current JS plugin exposes `lang` and forwards `lang` into Rime request parameters, which makes code using the documented Rime parameter shape harder to write and can silently send the wrong query key.

This keeps the existing `lang` option working for users who already depend on it, but prefers `language` when both are provided so new code can use the provider's documented API spelling.

## Validation

- `pnpm exec vitest run plugins/rime/src/tts.test.ts`
- `pnpm --filter @livekit/agents-plugin-rime lint`
- `pnpm exec prettier --check plugins/rime/src/tts.ts plugins/rime/src/tts.test.ts .changeset/rime-language-param.md`
- `pnpm --filter @livekit/agents-plugin-rime build`

I also ran `pnpm --filter @livekit/agents-plugin-rime api:check`; it reached API Extractor but exited nonzero on the existing missing API report/release-tag warnings for this plugin rather than a new type error.

### DIFF
--- a/.changeset/rime-language-param.md
+++ b/.changeset/rime-language-param.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-rime': patch
+---
+
+Send Rime TTS language selection with the documented `language` API parameter while preserving `lang` as a legacy option.
--- a/plugins/rime/src/tts.test.ts
+++ b/plugins/rime/src/tts.test.ts
@@ -4,8 +4,62 @@
 import { initializeLogger } from '@livekit/agents';
 import { STT } from '@livekit/agents-plugin-openai';
 import { tts } from '@livekit/agents-plugins-test';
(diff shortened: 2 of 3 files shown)


tinalenguyen (commented):
hi @chasef07, thank you for the PR! it lgtm, but maybe we should throw a logger warning when `lang` is used as well

chasef07:
@tinalenguyen good idea! I am talking with the Rime team now. They may be patching this on their side. Will updat ethis week

PR #1888, opened 2026-06-26
feat(assemblyai): add streaming inactivity timeout option

chasef07:
## Summary

- Add `inactivityTimeout` to AssemblyAI `STTOptions`.
- Forward it to the AssemblyAI Universal Streaming WebSocket query as `inactivity_timeout`.
- Add a local WebSocket unit test that verifies the query parameter is emitted.

## Why

AssemblyAI's streaming API supports an inactivity timeout for closing idle sessions. The JS plugin already exposes other connection-time Universal Streaming parameters in `STTOptions`, but this timeout was not available, forcing applications to carry a patch or accept the provider default.

This option is useful for voice agents that need explicit idle-session behavior in production, especially when calls can pause or when deployments need predictable cleanup of provider-side streaming sessions.

The option is documented as seconds to match the AssemblyAI API field semantics and to avoid mixing it with the framework's millisecond turn-timing options (`minTurnSilence`, `maxTurnSilence`, etc.).

## Validation

- `pnpm exec vitest run plugins/assemblyai/src/stt.test.ts`
- `pnpm exec prettier --check plugins/assemblyai/src/stt.ts plugins/assemblyai/src/stt.test.ts .changeset/assemblyai-inactivity-timeout.md`

I also ran `pnpm --filter @livekit/agents-plugin-assemblyai api:check`; it reached API Extractor but exited nonzero on the existing missing API report/release-tag/unresolved-link warnings for this plugin rather than a new type error.

### DIFF
--- a/.changeset/assemblyai-inactivity-timeout.md
+++ b/.changeset/assemblyai-inactivity-timeout.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-assemblyai': patch
+---
+
+Add AssemblyAI streaming `inactivityTimeout` support.
--- a/plugins/assemblyai/src/stt.test.ts
+++ b/plugins/assemblyai/src/stt.test.ts
@@ -3,8 +3,32 @@
 // SPDX-License-Identifier: Apache-2.0
 import { VAD } from '@livekit/agents-plugin-silero';
 import { stt } from '@livekit/agents-plugins-test';
+import { once } from 'node:events';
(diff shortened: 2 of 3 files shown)


tinalenguyen (approved):
thank you for the PR! :D

chasef07:
Thank you! Super excited to use these features without a patch and
hopefully help others.

Now just waiting on agents 1.5.0! Hopefully tonight?!

PR #1896, opened 2026-06-26
fix(warm-transfer): capture job context for post-merge caller-room cleanup

### LINKED ISSUE #1895: Node WarmTransferTask post-merge cleanup calls getJobContext after job context is gone
### Description

In the Node agents SDK, `beta.WarmTransferTask` registers a post-merge caller-room cleanup listener after `connect_to_caller` succeeds. When either SIP participant disconnects after the bridge, the listener fires, but it throws before it can delete the caller room because the late event callback calls `getJobContext()` after the job/agent context is no longer available.

This means the SDK appears to detect the exact event needed for cleanup, but the built-in cleanup path fails with an unhandled rejection.

### Environment

- `@livekit/agents`: observed on `1.4.7`
- `@livekit/rtc-node`: `0.13.29`
- `livekit-server-sdk`: `2.15.3`
- Workflow: Node agents `beta.WarmTransferTask`, inbound PSTN/SIP call, warm transfer to another SIP participant
- I also checked the current upstream default branch and the same `getJobContext()` usage appears to still be present in `agents/src/beta/workflows/warm_transfer.ts`: https://github.com/livekit/agents-js/blob/106ce74dc5d993260a2006e1c348ee9488d9f6ec/agents/src/beta/workflows/warm_transfer.ts

### Scenario

1. Original caller is in the caller room with the agent.
2. `WarmTransferTask` dials the transfer destination into the human-agent room.
3. The transfer destination accepts, and `connect_to_caller` bridges/moves the transfer destination into the caller room.
4. The application closes the agent session with reason `transfer_completed` after the bridge, so the AI agent is no longer participating in the human-to-human call.
5. Later, either human SIP participant hangs up.

### Expected behavior

When the SDK-registered `ParticipantDisconnected` handler fires after the bridge, it should either:

- delete the caller room / disconnect the remaining SIP participant successfully, or
- no-op safely if cleanup is unavailable.

It should not throw an unhandled rejection from `getJobContext()`.

### Actual behavior

The SDK logs that a participant disconnected from the caller room, then throws because the handler calls `getJobContext()` after the job context is gone.

This reproduces when either side disconnects after the bridge:

```text
participant disconnected from caller room, closing
    participantIdentity: "human-agent-sip"

Unhandled promise rejection
Error: no job context found, are you running this code inside a job entrypoint?
    at getJobContext (.../@livekit/agents/src/job.ts:49:11)
    at Room.onCallerParticipantDisconnected (.../@livekit/agents/src/beta/workflows/warm_transfer.ts:341:41)
    at Room.emit (node:events:520:35)
    at Room.Room.processFfiEvent (.../@livekit/rtc-node/src/room.ts:554:14)
    at FfiClient.Room.onFfiEvent (.../@livekit/rtc-node/src/room.ts:509:18)
```

And similarly for the original caller:

```text
participant disconnected from caller room, closing
    participantIdentity: "sip_+1XXXXXXXXXX"

Unhandled promise rejection
Error: no job context found, are you running this code inside a job entrypoint?
    at getJobContext (.../@livekit/agents/src/job.ts:49:11)
    at Room.onCallerParticipantDisconnected (.../@livekit/agents/src/beta/workflows/warm_transfer.ts:341:41)
```

### Source-level diagnosis

From the current implementation:

- `onEnter()` captures the caller room while job context exists:
  - `this._callerRoom = getJobContext().room`
- After `connect_to_caller` succeeds, the task registers a late room event handler:
  - `this._callerRoom.on(RoomEvent.ParticipantDisconnected, this.onCallerParticipantDisconnected)`
- But the later callback constructs the room service client using `getJobContext()` again:
  - `const rooms = new RoomServiceClient(getJobContext().info.url)`

That late callback can run after the job entrypoint/agent session has completed, so `getJobContext()` throws before `deleteRoom()` is attempted.

### Workaround attempted

We tried adding an application-level `participant_left` webhook reaper that lists participants and deletes the room when a post-transfer SIP room has one remaining SIP participant. That did not reliably tear down the original caller room in live testing, and in any case the SDK's own in-process listener is still throwing when it sees the disconnect.

### Suggested fix

A few possible SDK-side fixes seem viable:

1. Capture the `JobContext`, `RoomServiceClient`, or required connection/config values while still inside `onEnter()` / `mergeCalls()`, then use the captured values in `onCallerParticipantDisconnected`.
2. Use `getJobContext(false)` in the late callback and no-op/log if the context is gone, avoiding the unhandled rejection.
3. Expose an option to disable the built-in post-merge caller-room cleanup so applications can fully own teardown without racing the SDK handler.

Related community thread with the same live-test behavior: https://community.livekit.io/t/warmtransfertask-how-to-tear-down-a-2-party-sip-room-when-one-party-hangs-up-after-the-agent-has-left/1499

toubatbrian:
## Summary

`WarmTransferTask` registers a `RoomEvent.ParticipantDisconnected` listener on the caller room after `connect_to_caller` merges the human rep in. That handler called `getJobContext()` to build a `RoomServiceClient` and delete the caller room when a party hangs up.

The handler runs from a **native rtc-node FFI callback**, whose `AsyncLocalStorage` context is pinned to `FfiClient`-singleton creation — not the job's context. So `getJobContext()` reads an empty (or stale) store and throws, surfacing as an **unhandled promise rejection** and leaving the 2-party SIP room undeleted after the agent has left.

### Runtime evidence

Probes against the real `@livekit/rtc-node` confirmed the mechanism:

- `FfiClient` created **inside** the job context → ALS store present in the native callback.
- `FfiClient` created **outside** the job context (the real worker case: the job `Room` is built before `runWithJobContextAsync`, and a pooled child process creates the singleton before this job) → store **absent** in the native callback, regardless of where the listener was registered or the event emitted.

Post-fix probe in the same failing scenario: `broken_getJobContext_wouldThrow: true` vs `fixed_usable: true` inside the identical native callback.

## Fix

- Capture the `JobContext` eagerly in `onEnter()` (`this._jobCtx`), where the live context is provably available.
- The late `ParticipantDisconnected` handler now uses `this._jobCtx.deleteRoom(callerRoomName)` instead of `getJobContext()` at emit time. `JobContext.deleteRoom()` also forwards the job's API key/secret (the old `new RoomServiceClient(url)` dropped credentials and relied on env) and no-ops on fake jobs.
- `mergeCalls()` now sources its `RoomServiceClient` URL + credentials from the captured context.

Closes #1895.

## Test plan

- [x] `pnpm build:agents` (incl. `tsc` declarations)
- [x] ESLint + Prettier on changed file
- [x] Mechanism + post-fix runtime probes (described above)
- [ ] Live SIP end-to-end: run a warm transfer, then hang up the caller (and separately the rep) after the bridge; confirm the caller room is deleted and there is no "no job context found" unhandled rejection in the worker logs.

### DIFF
--- a/.changeset/false-interruption-audio-replay.md
+++ b/.changeset/false-interruption-audio-replay.md
@@ -1,0 +1,7 @@
+---
+'@livekit/agents': patch
+---
+
+fix(room_io): stop dropping audio on false interruptions
+
+`ParticipantAudioOutput.pause()` cleared the entire native `AudioSource` queue, permanently discarding up to `queueSizeMs` (rtc-node default: 1000ms) of generated-but-unplayed audio. On a false interruption (pause then resume) those frames were never replayed, so up to ~1s of agent speech was lost mid-sentence from both the live call and the observability recording. The output now keeps a rolling window of recently pushed frames, captures the unplayed tail on pause, and replays it on resume, while still discarding it on a real interruption (`clearBuffer()`).
--- a/.changeset/warm-transfer-job-context.md
+++ b/.changeset/warm-transfer-job-context.md
@@ -1,0 +1,7 @@
+---
+'@livekit/agents': patch
+---
+
+fix(warm-transfer): capture job context for post-merge caller-room cleanup
+
+`WarmTransferTask`'s post-merge `RoomEvent.ParticipantDisconnected` listener called `getJobContext()` to build a `RoomServiceClient` and delete the caller room. That handler runs from a native rtc-node FFI callback whose `AsyncLocalStorage` context is pinned to `FfiClient`-singleton creation, not to the job's context — so `getJobContext()` read an empty (or stale) store and threw, surfacing as an unhandled promise rejection and leaving the 2-party SIP room undeleted when a participant hung up after the bridge. The task now captures the `JobContext` eagerly in `onEnter()` (while the live context is available) and uses `jobCtx.deleteRoom()` in the late handler, which also passes the job's API credentials instead of relying on environment variables.
--- a/agents/src/beta/workflows/warm_transfer.ts
+++ b/agents/src/beta/workflows/warm_transfer.ts
@@ -6,7 +6,7 @@ import { type DisconnectReason, type ParticipantKind, Room, RoomEvent } from '@l
 import { AccessToken, RoomServiceClient, SipClient, type VideoGrant } from 'livekit-server-sdk';
 import { z } from 'zod';
 import type { LLMModels, STTModelString, TTSModelString } from '../../inference/index.js';
-import { getJobContext } from '../../job.js';
+import { type JobContext, getJobContext } from '../../job.js';
 import type {
   ChatContext,
   Instructions,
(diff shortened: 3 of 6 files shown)


toubatbrian:
Closing in favor of a fresh branch (brian/warm-transfer-fix) rebased on the 1.5.0 js-ification work. Same fix will be re-applied there.

PR #1898, opened 2026-06-27
fix(warm-transfer): capture job context for post-merge caller-room cleanup

toubatbrian:
## Summary

Fixes [#1895](https://github.com/livekit/agents-js/issues/1895): `WarmTransferTask`'s post-merge cleanup throws an unhandled rejection and leaves the 2-party SIP room undeleted.

After `mergeCalls()`, the task registers a `RoomEvent.ParticipantDisconnected` listener on the caller room to delete that room once a party hangs up. The handler called `getJobContext()`, but it runs from a native `@livekit/rtc-node` FFI callback whose `AsyncLocalStorage` context is pinned to `FfiClient`-singleton creation — **not** the job's context. So by the time the late event fires (the agent has already left, the session is closed), `getJobContext()` reads an empty/stale store and throws. The caller room is never deleted.

### The fix

- Capture the `JobContext` eagerly in `onEnter()` (while the live context is still available) and use `jobCtx.deleteRoom()` in the late handler. This avoids the dead `getJobContext()` lookup and uses the job's own API credentials instead of relying on process env vars.

### Also in this PR

- **Refactor to `AgentTask.create`.** `WarmTransferTask` is rebuilt with the functional `createWarmTransferTask` factory (now exported) for a more JS-native shape, keeping parity with the Python `warm_transfer.py` behavior without re-introducing the context bug.
- **Python parity.** The result and secondary-room naming now match the Python workflow: `WarmTransferResult.humanAgentIdentity`, participant identity `human-agent-sip`, and room suffix `-human-agent`.
- **`waitUntilAborted` moved to `utils.ts`** (out of `generation.ts`) so the task can reuse it, plus a pre-aborted short-circuit that swallows the abandoned promise's late rejection.
- **Example hardening.** `examples/src/warm_transfer.ts` pins `agentName` to force **explicit dispatch**. Without an explicit name the worker auto-dispatches into *every* new room — including the human agent room the task creates — putting a second agent on the line with the human agent and producing overlapping voices.

## Test plan

- [x] Warm transfer end-to-end: caller → agent → `transfer_to_human` → human agent dialed via SIP → `connect_to_caller` → `mergeCalls()`.
- [x] After the bridge, have a party hang up; confirm the caller room is deleted and **no** unhandled `getJobContext()` rejection appears in worker logs.
- [x] Confirm the human agent hears a single agent (no overlapping voices) with the example running under explicit dispatch.
- [x] `pnpm build:agents`.

### DIFF
--- a/agents/src/utils.ts
+++ b/agents/src/utils.ts
@@ -942,6 +942,71 @@ export function waitUntilTimeout<T, E extends Error = IdleTimeoutError>(
   ]).finally(() => clearTimeout(timer)) as Promise<Throws<T, E>>;
 }
 
+/** Result of {@link waitUntilAborted}: either the settled value, or an abort marker. */
+export type Aborted<T> =
+  | {
+      result: T;
+      isAborted: false;
+    }
+  | {
+      result: undefined;
+      isAborted: true;
+    };
+
+/**
+ * Race a promise against an `AbortSignal`. Unlike a plain reject-on-abort race,
+ * this resolves with a tagged result so callers can branch on `isAborted`
+ * instead of catching. On abort it resolves `{ result: undefined, isAborted: true }`;
+ * otherwise it resolves `{ result, isAborted: false }`. A rejection of the
+ * underlying promise is propagated. The abort listener is always cleaned up.
+ *
+ * An already-aborted signal short-circuits immediately to the abort result.
+ *
+ * Note: the underlying promise is not cancelled — it keeps running; this only
+ * stops waiting for it. Because the promise is passed already-created, the
+ * caller's operation has already started, so a pre-aborted signal won't prevent
+ * that side effect (pass/await a factory yourself if you need to avoid starting it).
+ */
+export async function waitUntilAborted<T>(
+  promise: Promise<T>,
+  signal: AbortSignal,
+): Promise<Aborted<T>> {
+  if (signal.aborted) {
+    // We're abandoning the promise, but it's already running: swallow any late
+    // rejection so it doesn't surface as an unhandled rejection (mirrors how the
+    // .catch below consumes a rejection that loses the race to an abort).
+    void promise.catch(() => {});
+    return { result: undefined, isAborted: true };
+  }
+
+  const abortFut = new Future<Aborted<T>>();
+
+  const resolveAbort = () => {
+    if (!abortFut.done) {
+      abortFut.resolve({ result: undefined, isAborted: true });
+    }
+  };
+
+  signal.addEventListener('abort', resolveAbort);
+
+  promise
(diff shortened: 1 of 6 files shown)


agents/src/utils.ts
```diff
@@ -942,6 +942,71 @@ export function waitUntilTimeout<T, E extends Error = IdleTimeoutError>(
...
+ * underlying promise is propagated. The abort listener is always cleaned up.
+ *
+ * An already-aborted signal short-circuits immediately to the abort result.
+ *
+ * Note: the underlying promise is not cancelled — it keeps running; this only
+ * stops waiting for it. Because the promise is passed already-created, the
+ * caller's operation has already started, so a pre-aborted signal won't prevent
+ * that side effect (pass/await a factory yourself if you need to avoid starting it).
+ */
+export async function waitUntilAborted<T>(
```
chenghao-mou, line 970: I think we have some redundancy with `waitUnlessAborted`/`sleepUnlessAborted` from `run_context.ts`.

agents/src/utils.ts
```diff
@@ -942,6 +942,71 @@ export function waitUntilTimeout<T, E extends Error = IdleTimeoutError>(
...
+  promise: Promise<T>,
```
chenghao-mou, line 971: If a `toolExecTask` finishes before being aborted (in `generation.ts`), should we surface that?

chenghao-mou (approved):
One question, otherwise lgtm

agents/src/utils.ts
toubatbrian: Will cleanup in a different PR

agents/src/utils.ts
toubatbrian:
```ts
  promise
    .then((r) => {
      if (!abortFut.done) {
        abortFut.resolve({ result: r, isAborted: false });
      }
    })
```

The `.then` matches the case where promise finish before abort, or unless I misunderstand certain things?

chenghao-mou: Oh, I was referring to the early return on line 974.

toubatbrian: I think this shouldn’t happen. If the promise finishes before line 974, that would only mean `toolExecTask` was implemented & evaluated synchronously, and the event loop wouldn’t have had a chance to propagate the abort status to the abort signal yet.

All the code before `toolExecTask` in `generation.ts` is also evaluated synchronously and is gated by `while (!signal.aborted) { ... }`. So either the promise evaluates all the way to the end before observing `signal.aborted`, or the signal changes before the promise finishes evaluation.

In either case, the function should still end up with the expected behavior.

PR #1910, opened 2026-06-29
fix(voice): resume speech interrupted before its first frame (#1909)

### LINKED ISSUE #1909: Agent speech silently dropped when interrupted before the first audio frame (resumeFalseInterruption) — port of Python #5039
### Summary

With `resumeFalseInterruption: true`, a brief user sound that arrives **after `say()` is issued but before the agent's first TTS audio frame is forwarded** pauses the speech, leaves `firstFrameFut` unresolved, and the turn is dropped — the user hears **no audio at all** and the turn is dropped from history. The call then sits in silence.

This is the JS counterpart of livekit/agents#5038 (Python), which was fixed by livekit/agents#5039. That fix does not appear to be ported to `agents-js` (the relevant code path is unchanged on `main`).

### Environment

| Field | Value |
|---|---|
| `@livekit/agents` | 1.4.6 (relevant path identical on 1.4.7 and current `main`) |
| Node | 22.x |
| Turn detection | streaming STT, `turnDetection: "stt"` |
| Interruption config | `{ mode: "adaptive", minWords: 2, minDuration: 500, resumeFalseInterruption: true }` |
| Transport | outbound telephony (SIP); mechanism is transport-independent |

### Steps to reproduce

1. Outbound call. On answer, the agent speaks a fixed opener: `session.say(text, { allowInterruptions: true })`.
2. The TTS has a non-trivial time-to-first-byte (a few hundred ms), so the first audio frame is still in flight after `say()` returns.
3. The callee makes a brief sound ("hello?") in that pre-first-frame window.

### Expected

The brief false interruption pauses the speech and then **resumes and plays it** once the false interruption clears (per `resumeFalseInterruption`), or the speech is interrupted cleanly and re-attempted.

### Actual

`firstFrameFut` never resolves; the first audio frame arrives after the segment is torn down and is discarded; the turn is dropped from history. **No audio reaches the user**, and the interruption is not counted (`session` reports zero interruptions). `minWords` is irrelevant — the path that fires has no word-count gate.

### Root cause (JS)

Two behaviors combine.

**1. `onStartOfSpeech` pauses a not-yet-playing speech, ungated.** In `voice/agent_activity.ts`, `onStartOfSpeech` pauses the current speech as soon as user VAD-start fires, guarded only by:

```ts
agentState !== "speaking"            // agent's first audio frame hasn't played yet
&& pauseEnabled()
&& _currentSpeech.allowInterruptions  // the opener is interruptible
```

There is no `minWords` and no duration check on this branch, so any user sound in the pre-first-frame window pauses the speech.

**2. The pause leaves `firstFrameFut` unresolved, so audio + transcript are dropped.** In `voice/generation.ts`, `forwardAudio`'s `finally` rejects the future when no frame was forwarded:

```ts
if (!out.firstFrameFut.done) {
  out.firstFrameFut.reject(new Error("audio forwarding cancelled before playback started"));
}
audioOutput.flush();
if (signal?.aborted) audioOutput.clearBuffer();
```

Downstream, the reply task only preserves the synchronized transcript when `firstFrameFut.done && !firstFrameFut.rejected`, so on the rejected (no-first-frame) path the transcript is blanked and the audio is discarded — the JS analogue of the `else: forwarded_text = ""` overwrite called out in livekit/agents#5038.

Observable log signature on the dropped turn:

```
SegmentSynchronizerImpl.markPlaybackFinished called before text/audio input is done
SegmentSynchronizerImpl.onPlaybackStarted called after close
playback_finished called more times than playback segments were captured
```

### Minimal reproduction (mechanism)

The destruction half is deterministic against the real `performAudioForwarding`: when the forwarding signal is aborted before the first frame, zero frames reach the sink and `firstFrameFut` rejects.

| Case | Frames to output | `firstFrameFut` | `clearBuffer` |
|---|---|---|---|
| Frames flow, no cancel | 3 | resolved | 0 |
| Cancelled before first frame | 0 (silent) | rejected | 1 |
| TTS yields no frame | 0 (silent) | rejected | 0 |

I can attach a runnable `AgentSession`-level reproduction if helpful.

### Relationship to the Python fix

livekit/agents#5038 describes this exact failure and was fixed by livekit/agents#5039, which relocates `first_frame_fut` handling to the callers and avoids blanking the generated text on the pre-first-frame path. The JS port still rejects `firstFrameFut` in `forwardAudio` and gates transcript preservation on `!firstFrameFut.rejected`, so the same defect is present.

One porting subtlety: in the JS `Future`, `cancel()` also sets `rejected = true`, so a literal "cancel instead of reject" transcription of #5039 would not change the downstream `!firstFrameFut.rejected` check. The JS fix likely needs to preserve the generated audio/transcript on the no-first-frame path explicitly.

### Ask

Port the #5039 fix to `agents-js` (or confirm the intended approach), so a brief false interruption arriving before the first audio frame no longer drops the entire turn — and ideally resumes it, per `resumeFalseInterruption`. Happy to open a PR with a regression test if that's welcome.

enriqueespaillat-gyde:
## Summary

Port of livekit/agents#5039 (Python issue livekit/agents#5038) to `agents-js`.

When the agent is in the **"thinking"** state and the user makes a brief sound **before the first TTS audio frame is forwarded**, the speech is silently dropped: no audio reaches the user and the turn is dropped from conversation history. This happens with `resumeFalseInterruption: true` (the default).

Closes #1909.

## Root cause

Two behaviors combine in `agents/src/voice/`:

1. **`agent_activity.ts` → `onStartOfSpeech`** pauses the current speech when the agent is not yet `"speaking"` (thinking state). This pause is **intentional** — it stops the agent talking over a user who barges in during the thinking state — and is **preserved** by this PR.
2. **`generation.ts` → `forwardAudio`** registered its `PLAYBACK_STARTED` listener *inside the forwarding task* and, in its `finally`, **rejected** `firstFrameFut` whenever no frame had played. The thinking-state pause buffers the (short) TTS frames; the forwarding task then finishes before playback starts → `firstFrameFut` is rejected and the listener removed. When the false interruption clears and the output resumes, the buffered first frame plays but **nothing is listening**, so the future stays rejected. The reply tasks gate transcript preservation on `firstFrameFut.done && !firstFrameFut.rejected`, so the resumed turn is blanked from history even though audio reached the user.

## Fix

Mirrors the spirit of #5039 — make the pre-first-frame pause **recoverable** while keeping the thinking-state pause:

- Move the `PLAYBACK_STARTED` listener from `forwardAudio` to `performAudioForwarding` so it **outlives** the forwarding task; a late first frame (e.g. after a `resumeFalseInterruption` resume) can still resolve `firstFrameFut`.
- Stop rejecting `firstFrameFut` in `forwardAudio`'s `finally`.
- Settle the future in the reply tasks (say / pipeline / realtime) after playout finishes or is interrupted, which also removes the listener.

### JS-specific note

Unlike Python, the JS `Future` (`agents/src/utils.ts`) has no `cancel()` distinct from `reject()` — `reject()` sets `rejected = true`. So a literal "cancel instead of reject" transcription of #5039 would **not** change the downstream `!firstFrameFut.rejected` gate. This fix preserves the audio/transcript on the no-first-frame path by **relocating resolution** so the late first frame resolves the future, rather than relying on a cancel/reject distinction.

## Tests

New regression test `generation_interrupt_before_first_frame.test.ts` reproduces the thinking-state pause before the first frame with a pausable mock output:

- **False interruption** → speech resumes, frames are forwarded, `firstFrameFut` resolves, and the synchronized transcript would be preserved.
- **Genuine interruption after a resume** → the partial synchronized transcript is kept (turn not lost from history).

Both cases **fail on `main` and pass with this change**. The existing `generation_tts_timeout.test.ts` "ignores PLAYBACK_STARTED from another segment" assertion is updated to the new contract (`forwardAudio` no longer rejects the future).

- Full `agents` test suite: green (0 failed).
- `pnpm build`, ESLint, Prettier: green.

### DIFF
--- a/.changeset/interrupt-before-first-frame.md
+++ b/.changeset/interrupt-before-first-frame.md
@@ -1,0 +1,9 @@
+---
+'@livekit/agents': patch
+---
+
+Fix agent speech being silently dropped when interrupted before its first audio frame plays (#1909, port of livekit/agents#5039).
+
+When the agent is in the "thinking" state and the user makes a brief sound before the first TTS frame is forwarded, `onStartOfSpeech` pauses the not-yet-playing speech (this thinking-state pause is intentional and preserved). The frames were still captured into the paused output buffer, but `forwardAudio`'s `finally` block rejected `firstFrameFut` (and removed its `PLAYBACK_STARTED` listener) whenever no frame had played yet. So when a false interruption cleared and the output resumed, the buffered first frame played but nothing was listening — the future stayed rejected, and because the reply tasks gate transcript preservation on `firstFrameFut.done && !firstFrameFut.rejected`, the resumed turn was dropped from history even though audio reached the user.
+
+The `PLAYBACK_STARTED` listener now lives in `performAudioForwarding` so it outlives the forwarding task, and `forwardAudio` no longer rejects the future; a late first frame (e.g. after a `resumeFalseInterruption` resume) can still resolve it. The reply tasks settle the future after playout finishes or is interrupted to remove the listener. A genuine interruption after a resume now keeps its partial synchronized transcript instead of losing the turn.
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -2442,58 +2442,62 @@ export class AgentActivity implements RecognitionHooks {
         .catch(() => this.logger.debug('firstFrameFut cancelled before first frame'));
     }
 
-    await speechHandle.waitIfNotInterrupted(tasks.map((task) => task.result));
+    try {
+      await speechHandle.waitIfNotInterrupted(tasks.map((task) => task.result));
 
-    if (audioOutput) {
-      await speechHandle.waitIfNotInterrupted([audioOutput.waitForPlayout()]);
-    }
-
-    if (speechHandle.interrupted) {
-      replyAbortController.abort();
-      await cancelAndWait(tasks, AgentActivity.REPLY_TASK_CANCEL_TIMEOUT);
       if (audioOutput) {
-        audioOutput.clearBuffer();
-        await audioOutput.waitForPlayout();
+        await speechHandle.waitIfNotInterrupted([audioOutput.waitForPlayout()]);
       }
-    }
 
-    if (addToChatCtx) {
-      const replyStoppedSpeakingAt = Date.now();
-      const replyAssistantMetrics: MetricsReport = {};
-      if (replyTtsGenData?.ttfb !== undefined) {
-        replyAssistantMetrics.ttsNodeTtfb = replyTtsGenData.ttfb;
-      }
-      if (replyStartedSpeakingAt !== undefined) {
-        replyAssistantMetrics.startedSpeakingAt = replyStartedSpeakingAt / 1000; // ms -> seconds
-        replyAssistantMetrics.stoppedSpeakingAt = replyStoppedSpeakingAt / 1000; // ms -> seconds
-
-        if (replyStartedForwardingAt !== undefined) {
-          replyAssistantMetrics.playbackLatency =
-            (replyStartedSpeakingAt - replyStartedForwardingAt) / 1000; // ms -> seconds
+      if (speechHandle.interrupted) {
+        replyAbortController.abort();
+        await cancelAndWait(tasks, AgentActivity.REPLY_TASK_CANCEL_TIMEOUT);
+        if (audioOutput) {
+          audioOutput.clearBuffer();
+          await audioOutput.waitForPlayout();
         }
       }
 
-      const message = ChatMessage.create({
-        role: 'assistant',
-        content: textOut?.text || '',
-        interrupted: speechHandle.interrupted,
-        metrics: replyAssistantMetrics,
-      });
-      this.agent._chatCtx.insert(message);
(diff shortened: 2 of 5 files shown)


enriqueespaillat-gyde (commented):
r? @toubatbrian - please let me know if I should stop tagging you and if theres another process to follow :) There's been an uptick in PRs past few months so just want to make sure you see this one. Its hit us a few times in production.

toubatbrian:
Hi @enriqueespaillat-gyde — thanks again for this fix and the excellent test suite. It turns out #1960 addresses the same root cause (the `firstFrameFut` rejection in `forwardAudio`'s `finally`) from the avatar side, and the two PRs edit the same lines in `generation.ts` / `agent_activity.ts`, so whichever landed first would force a painful rebase on the other.

To simplify coordination we've opened #1966, which combines both: your structure (listener in `performAudioForwarding`, caller-side settle helper, `say()` path coverage) is the backbone, your tests are ported nearly verbatim, and you're credited as a co-author on the commit. Would appreciate your eyes on it to make sure nothing misrepresents your original intent — we'll close this PR in favor of the combined one once you've had a chance to look.

enriqueespaillat-gyde:
Closing in favor of #1966

PR #1916, opened 2026-06-30
fix(voice): commit interrupted speech without a synchronized transcript

ShayneP:
On a midplayout interruption, `forwardedTextFor` and the realtime reply path returned `synchronizedTranscript ?? ''`, so an interrupted-but-heard reply with no playback-aligned transcript (common with avatar outputs) produced no `conversation_item_added` and was dropped from `chatCtx`. Fall back to the forwarded generation text, matching the Python SDK's `forwarded_text` behaviour.

### DIFF
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -2593,7 +2593,7 @@ export class AgentActivity implements RecognitionHooks {
     const forwardedTextFor = (output: SegmentOutput): string => {
       if (output.played === 'skipped') return '';
       if (output.played === 'partial' && output.audioOut) {
-        return output.synchronizedTranscript ?? '';
+        return output.synchronizedTranscript || output.textOut?.text || '';
       }
       return output.textOut?.text ?? '';
     };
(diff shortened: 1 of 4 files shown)


agents/src/voice/agent_activity.ts
```diff
@@ -2593,7 +2593,7 @@ export class AgentActivity implements RecognitionHooks {
     const forwardedTextFor = (output: SegmentOutput): string => {
```
u9g: https://github.com/livekit/agents/blob/6e55cad5815feaccf0f83df4dde5e386eb8dbc34/livekit-agents/livekit/agents/voice/generation.py#L494-L501

agents/src/voice/agent_activity.ts
```diff
@@ -3427,7 +3427,7 @@ export class AgentActivity implements RecognitionHooks {
         const interrupted = output.played === 'partial';
         let forwardedText = output.textOut?.text || '';
```
u9g: https://github.com/livekit/agents/blob/6e55cad5815feaccf0f83df4dde5e386eb8dbc34/livekit-agents/livekit/agents/voice/agent_activity.py#L3646-L3654

u9g (approved):

PR #1935, opened 2026-07-02
fix(openai): add receive-side timeout to Responses WebSocket LLM

enriqueespaillat-gyde:
## Description

The OpenAI **Responses** LLM can run over a persistent WebSocket (`plugins/openai/src/ws/llm.ts`, selected via `useWebSocket: true`, the default). That path had **no receive-side timeout**: after a `response.create` is sent, `WSLLMStream` awaits `reader.read()` on the request's `StreamChannel` in an unbounded loop. If the server — or an intermediary OpenAI-compatible gateway (e.g. LiteLLM) — returns **no events** while the socket stays **open and error-free**, the receive loop blocks forever. There is no watchdog, and the SDK's retry/reconnect never engages because nothing throws.

We hit this in production: a voice agent went silent for ~67s after the caller answered, then the caller hung up. On the gateway side, a single Responses WebSocket stayed open and error-free for the whole call (no close, no error frame), so nothing ever broke the wait.

The only timeout in the WS path today is the **connection-establishment** timeout in `connectWs` (a `setTimeout` around the handshake) — nothing guards event *reception*. Note the sibling HTTP branch in `responses/llm.ts` already passes `{ timeout: connOptions.timeoutMs }` to `client.responses.create`; the WS branch dropped that guarantee.

## Changes Made
- **First-event / inactivity deadline on the WS receive path** (`ResponsesWebSocket`). Each queued request may carry a `timeoutMs`. If the server sends no event within that window of the `response.create` (first-event deadline), or the gap between two consecutive events exceeds it (inactivity deadline), the request's channel is aborted with a **retryable `APIConnectionError`**, the timed-out entry is removed from the FIFO queue (so a reused connection can't be misrouted to a dead head entry), and the socket is torn down so the pool reconnects on the next turn. The retryable error engages the base `LLMStream` retry loop (`agents/src/llm/llm.ts`) — the same mechanism the existing `websocket_closed` path relies on.
- **New `responseTimeoutMs` option**, wired `LLM` → `WSLLM` → `WSLLMStream`. Default **15000 ms** (well above the ~1s healthy first-event latency observed in prod, so it won't false-fire); set to `0` to disable. Deliberately **not** reused from the connect `timeoutMs` — they're different concerns.
- **Parallel-generation guard** (ported from the Python plugin). Track in-flight generations (`_onStreamStarted` / `_onStreamFinished`); when turns overlap, reset the stored `previous_response_id` / chat-ctx continuation chain once all overlapping generations drain, and only take the `previous_response_id` continuation shortcut when idle (`#activeStreams === 0`). Previously the TS `WSLLM` used the stored continuation unconditionally.
- **Post-loop retryable** (ported from Python). If the receive loop drains without a `response.completed`, throw a retryable `APIConnectionError` instead of returning `false` silently.
- The existing TS-only `previous_response_not_found` recovery in `#handleError` is preserved.
- Changeset added (`@livekit/agents-plugin-openai`: patch).

## Pre-Review Checklist
- [x] **Build passes**: `pnpm --filter @livekit/agents-plugin-openai build`, ESLint, and Prettier pass locally. (`api:check` fails identically on a clean checkout of `main` due to a pre-existing api-extractor / bundled-TS `export * as` limitation — unrelated to this change.)
- [x] **AI-generated code reviewed**
- [x] **Changes explained**
- [x] **Scope appropriate**

## Testing
- [x] Automated tests added (`plugins/openai/src/ws/llm.test.ts`):
  - first-event deadline fires → retryable `APIConnectionError` (no infinite hang), socket torn down;
  - inactivity deadline fires after the first event;
  - healthy stream (events arrive promptly) is unaffected and the socket is left open;
  - `responseTimeoutMs: 0` disables the timer;
  - parallel-generation guard: serial turns chain via `previous_response_id`, overlapping turns suppress it and reset the continuation chain afterward (driven end-to-end through an in-process WebSocket server).
- [x] All tests pass (`vitest run plugins/openai`: 56 passed).

## Additional Notes
The retryable `APIConnectionError` reuses the same reconnect/retry semantics as the existing `websocket_closed` handling, so behavior on failure matches what reviewers already expect from this file.

### DIFF
--- a/.changeset/openai-ws-first-event-timeout.md
+++ b/.changeset/openai-ws-first-event-timeout.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-openai': patch
+---
+
+Add a receive-side (first-event / inactivity) timeout to the OpenAI Responses WebSocket LLM. Previously, if a `response.create` was sent but the server (or an intermediary gateway) left the socket open and silent, the receive loop blocked forever — no watchdog, and retries never engaged because nothing threw. The persistent WS now aborts the turn with a retryable `APIConnectionError` (reconnecting on the next turn) when no event arrives within `responseTimeoutMs` of the request, or when the gap between two consecutive events exceeds it. Configurable via the new `responseTimeoutMs` option (default 15000 ms; set to `0` to disable). Also ports two robustness features from the Python plugin: a parallel-generation guard that resets the stored `previous_response_id` continuation chain when turns overlap, and a retryable error when the receive loop drains without a completed response.
--- a/plugins/openai/src/responses/llm.ts
+++ b/plugins/openai/src/responses/llm.ts
@@ -36,6 +36,20 @@ export interface LLMOptions {
    * @default true
    */
   useWebSocket?: boolean;
+
+  /**
+   * Receive-side deadline in milliseconds for the persistent Responses
+   * WebSocket (only applies when {@link useWebSocket} is `true`). After a
+   * request is sent, the turn is aborted and retried if the server sends no
+   * event within this window, or if the gap between two consecutive events
+   * exceeds it — guarding against the socket staying open but silent, which
+   * would otherwise hang the turn indefinitely. Set to `0` to disable. This is
+   * distinct from the connection-establishment / per-request `connOptions`
+   * timeout used by the HTTP path.
+   *
+   * @defaultValue 15000
+   */
+  responseTimeoutMs?: number;
 }
 
 type HttpLLMOptions = Omit<LLMOptions, 'useWebSocket'>;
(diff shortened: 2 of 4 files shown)


enriqueespaillat-gyde:
Good catch on the race — but incrementing in `chat()` while releasing in `run()`'s finally underflows the counter: `run()` is called once per retry attempt (`maxRetry+1`), so one increment vs N decrements sends `#activeStreams` negative → `=== 0` never holds again → continuation is permanently disabled after the first retry. (Same issue with incrementing in the constructor alone.)

Fixed in 2fdfa673: reserve in the `WSLLMStream` constructor (synchronous during `chat()`, after the continuation decision) so same-tick chats can't both continue, and guard the release with `#streamReleased` so it fires exactly once per stream regardless of retries. Added a regression test that fails if the reservation is deferred back into `run()`.

PR #1957, opened 2026-07-05
fix(deepgram): per-iteration abort promise prevents AudioFrame retention on long-running streams

### LINKED ISSUE #1950: Memory leak in the Deepgram STT plugin: audio frames pile up for the entire session on long-running streams
## What I observe

Using the Deepgram STT plugin in a Node.js app to transcribe a **continuous** audio
stream over a **long** session, the process memory grows linearly for the whole
duration of the session and only drops back to baseline when the STT stream is closed.

Over a ~1h session it grows by ~0.5 GB per open stream, with no plateau. On a
multi-hour live the process is eventually OOM-killed mid-session. The abrupt release
at stream close (rather than gradual reclamation during the session) indicates
something is retained for the whole lifetime of the stream instead of being freed
frame by frame.

## How the stream is used in my app

Nothing exotic — the documented streaming STT usage, just sustained for a long time
(irrelevant details elided):

    const speech = stt.stream();               // one long-lived stream
    const audio = new AudioStream(track, 16000); // remote 16 kHz track

    // push every track frame into the stream (~100 frames/s, 10 ms frames)
    const reader = audio.getReader();
    for (;;) {
      if (done) break;
      speech.pushFrame(value);
    }

    // drain transcription events concurrently
    for await (const event of speech) {
      // ... handle FINAL_TRANSCRIPT ...
    }

    // the stream stays open for the whole session; closed only at teardown:
    // speech.close();

## Minimal reproduction (no app logic)

The same growth reproduces with a bare STT stream fed synthetic 10 ms silent frames at
100/s — no room, no AgentSession, no application code:

    import { initializeLogger } from "@livekit/agents";
    import { STT as DeepgramSTT } from "@livekit/agents-plugin-deepgram";
    import { AudioFrame } from "@livekit/rtc-node";

    initializeLogger({ pretty: true, level: "warn" });
    const stt = new DeepgramSTT({ apiKey: process.env.DEEPGRAM_API_KEY!, model: "nova-3-medical" as never });
    const stream = stt.stream();
    void (async () => { for await (const _ of stream) { /* drain */ } })();

    const silence = new Int16Array(160); // 10 ms @ 16 kHz mono
    setInterval(() => stream.pushFrame(new AudioFrame(silence.slice(), 16000, 1, 160)), 10); // 100 frames/s
    setInterval(() => console.log((process.memoryUsage().rss / 1048576 | 0) + " MiB"), 5000);

The growth happens even during pure silence (no transcripts produced), so it is not
tied to transcript volume — it tracks the number of input frames pushed.

## What the heap shows

V8 heap snapshots taken during the session show retained `AudioFrame` objects
accumulating 1:1 with the number of frames pushed (~100/s at 16 kHz). The retention
chain is entirely inside the plugin: each retained frame is held by a `PromiseReaction`
on a single long-lived `Future` (`waitForAbort(this.abortSignal)`) that `sendTask`
races on every frame via `Promise.race([this.input.next(), abortPromise])`. The
reaction chain on that promise grows by one node per frame and is released only when
the promise settles (stream close). No application object appears anywhere in the
retention chain. Full numbers in "Relevant output".

This matches the behaviour described in nodejs/node#17469 (reactions attached to the
losing operand of `Promise.race` are not detached while that operand stays pending).

## Environment

- `@livekit/agents-plugin-deepgram@1.4.9`, `@livekit/agents@1.4.9`
- Node.js v22.14.0

Happy to share the full heap snapshots or the reproduction output if that helps.

### Relevant log output

```text
❯ $env:REPRO_BATCH=10; $env:REPRO_SECONDS=75; npx tsx scripts/leak-repro.ts
t=  5s  frames=  3150  rss=186.8 MiB  heapUsed=64.7 MiB  heapTotal=97.3 MiB  external=8.5 MiB  arrayBuffers=1.6 MiB  | Δrss=+5.6MiB (67.1 MiB/min)
t= 10s  frames=  6360  rss=173.1 MiB  heapUsed=62.4 MiB  heapTotal=97.3 MiB  external=9.4 MiB  arrayBuffers=2.5 MiB  | Δrss=+-8.1MiB (-48.6 MiB/min)
t= 15s  frames=  9580  rss=174.5 MiB  heapUsed=76.0 MiB  heapTotal=97.3 MiB  external=10.6 MiB  arrayBuffers=3.7 MiB  | Δrss=+-6.7MiB (-26.9 MiB/min)
t= 20s  frames= 12800  rss=176.7 MiB  heapUsed=75.3 MiB  heapTotal=97.8 MiB  external=11.5 MiB  arrayBuffers=4.6 MiB  | Δrss=+-4.5MiB (-13.6 MiB/min)
t= 25s  frames= 16010  rss=180.9 MiB  heapUsed=73.0 MiB  heapTotal=100.6 MiB  external=12.4 MiB  arrayBuffers=5.5 MiB  | Δrss=+-0.3MiB (-0.7 MiB/min)
t= 30s  frames= 19230  rss=184.8 MiB  heapUsed=72.2 MiB  heapTotal=102.8 MiB  external=13.3 MiB  arrayBuffers=6.4 MiB  | Δrss=+3.6MiB (7.2 MiB/min)
t= 35s  frames= 22450  rss=188.7 MiB  heapUsed=85.9 MiB  heapTotal=105.3 MiB  external=14.5 MiB  arrayBuffers=7.6 MiB  | Δrss=+7.5MiB (12.9 MiB/min)
t= 40s  frames= 25670  rss=192.6 MiB  heapUsed=85.2 MiB  heapTotal=107.8 MiB  external=15.4 MiB  arrayBuffers=8.5 MiB  | Δrss=+11.4MiB (17.0 MiB/min)
t= 45s  frames= 28860  rss=196.8 MiB  heapUsed=79.6 MiB  heapTotal=110.8 MiB  external=16.2 MiB  arrayBuffers=9.4 MiB  | Δrss=+15.6MiB (20.8 MiB/min)
t= 50s  frames= 32020  rss=201.3 MiB  heapUsed=83.7 MiB  heapTotal=113.3 MiB  external=17.2 MiB  arrayBuffers=10.4 MiB  | Δrss=+20.1MiB (24.1 MiB/min)
t= 55s  frames= 35240  rss=204.8 MiB  heapUsed=83.0 MiB  heapTotal=115.8 MiB  external=18.1 MiB  arrayBuffers=11.3 MiB  | Δrss=+23.6MiB (25.7 MiB/min)
t= 60s  frames= 38440  rss=208.7 MiB  heapUsed=93.4 MiB  heapTotal=118.3 MiB  external=19.3 MiB  arrayBuffers=12.4 MiB  | Δrss=+27.4MiB (27.4 MiB/min)
t= 65s  frames= 41620  rss=213.0 MiB  heapUsed=100.6 MiB  heapTotal=120.6 MiB  external=20.3 MiB  arrayBuffers=13.5 MiB  | Δrss=+31.7MiB (29.3 MiB/min)
t= 70s  frames= 44830  rss=216.9 MiB  heapUsed=98.3 MiB  heapTotal=123.1 MiB  external=21.2 MiB  arrayBuffers=14.4 MiB  | Δrss=+35.7MiB (30.5 MiB/min)
```

### Describe your environment

System:
  OS: Windows 11 10.0.26200
  CPU: (16) x64 AMD Ryzen 7 8845HS w/ Radeon 780M Graphics
  Memory: 31.28 GB
Binaries:
  Node: 22.14.0
  npm: 10.9.2
npmPackages:
  @livekit/agents: ^1.4.9 => 1.4.9
  @livekit/agents-plugin-deepgram: ^1.4.9 => 1.4.9
  @livekit/agents-plugin-elevenlabs: ^1.4.9 => 1.4.9

1. Create `repro.ts` with the code from the "Minimal reproducible example" below
   (a bare `DeepgramSTT().stream()` fed synthetic 16 kHz silent frames, no room / AgentSession).
2. Install the packages: `npm i @livekit/agents @livekit/agents-plugin-deepgram @livekit/rtc-node tsx`
3. Run it with a valid Deepgram key:
   `DEEPGRAM_API_KEY=<key> npx tsx repro.ts`
   (defaults to an accelerated ~1000 frames/s so the leak shows within a minute;
    set `REPRO_BATCH=1` for a faithful real-time 100 frames/s rate.)
4. Watch the per-5s memory log: after ~15 s of JIT/GC warmup, `arrayBuffers` and
   `external` climb linearly at ~320 bytes per pushed frame (one 10 ms / 160-sample
   frame) and `rss` trends steadily upward — with no plateau.
5. The memory is reclaimed only when the stream is closed at the end of the run,
   never during it. The growth tracks the number of input frames, not transcript
   volume (it happens even on pure silence).

### Additional information

I generate a small standalone reproduction script (with the help of Claude code) that
isolates the plugin from any application code and makes the leak visible in under a
minute by pushing synthetic 16 kHz frames at an accelerated rate. It logs the full
`process.memoryUsage()` breakdown every 5 s; `arrayBuffers` and `external` grow linearly
at ~320 bytes per pushed frame (exactly one 10 ms / 160-sample frame) and are never
reclaimed until the stream is closed. Happy to share the script and its output.

```typescript
/**
 * Minimal, self-contained reproduction of the Deepgram STT memory leak.
 *
 * It exercises ONLY `@livekit/agents-plugin-deepgram`'s `SpeechStream` — no
 * ContextBuffer, no VisionSampler, no metrics, no AgentSession, no LiveKit room,
 * none of our business logic. A bare STT stream is fed synthetic 10 ms silent
 * audio frames at 100 frames/s (exactly what a 16 kHz track produces), while RSS
 * and heap are logged every 5 s.
 *
 * Purpose: prove the linear memory growth originates in the plugin's `sendTask`
 * loop (`Promise.race([this.input.next(), abortPromise])`), independent of how
 * our agent consumes audio. If RSS climbs linearly here, the leak is upstream.
 *
 * Run:  npx tsx scripts/leak-repro.ts
 * Needs: DEEPGRAM_API_KEY in the environment (loaded from .env like the app).
 */
import "dotenv/config";
import { initializeLogger } from "@livekit/agents";
import { STT as DeepgramSTT } from "@livekit/agents-plugin-deepgram";
import { AudioFrame } from "@livekit/rtc-node";

const SAMPLE_RATE = 16_000;
const CHANNELS = 1;
const SAMPLES_PER_FRAME = SAMPLE_RATE / 100; // 160 samples = 10 ms frame → 100 frames/s
const RUN_SECONDS = Number(process.env["REPRO_SECONDS"] ?? "180");
// Frames pushed per 10 ms tick. 1 = real-time (100/s); raise it to compress many
// minutes of live audio into a short run so the linear leak is visible fast.
const BATCH = Number(process.env["REPRO_BATCH"] ?? "1");

function fmtMiB(bytes: number): string {
  return `${(bytes / (1024 * 1024)).toFixed(1)} MiB`;
}

async function main(): Promise<void> {
  // The plugin logs through the agents logger; initialize it first (mandatory order).
  initializeLogger({ pretty: true, level: "warn" });

  const apiKey = process.env["DEEPGRAM_API_KEY"];
  if (!apiKey) throw new Error("DEEPGRAM_API_KEY missing");

  const stt = new DeepgramSTT({ apiKey, model: "nova-3-medical" as never });
  const stream = stt.stream();

  // Drain transcription events so the output side never applies backpressure —
  // we only care about the input/send side where frames accumulate.
  void (async () => {
    try {
      for await (const _ev of stream) {
        /* discard: silent audio yields little, and events are text, not audio */
      }
    } catch {
      /* stream closed on shutdown — expected */
    }
  })();

  // A single reusable silent frame's payload; a fresh AudioFrame wraps it each
  // tick so every pushed frame is a distinct object (mirroring real capture).
  const silence = new Int16Array(SAMPLES_PER_FRAME); // all zeros = silence

  const startRss = process.memoryUsage().rss;
  const startedAt = Date.now();
  let pushed = 0;

  // Push one 10 ms frame every 10 ms → 100 frames/s, matching a live 16 kHz track.
  const pusher = setInterval(() => {
    for (let i = 0; i < BATCH; i++) {
      const frame = new AudioFrame(silence.slice(), SAMPLE_RATE, CHANNELS, SAMPLES_PER_FRAME);
      stream.pushFrame(frame);
      pushed++;
    }
  }, 10);

  // Report the full process memory breakdown + pushed-frame count every 5 s.
  const reporter = setInterval(() => {
    const m = process.memoryUsage();
    const elapsed = (Date.now() - startedAt) / 1000;
    const grownMiB = (m.rss - startRss) / (1024 * 1024);
    const ratePerMin = elapsed > 0 ? (grownMiB / elapsed) * 60 : 0;
    console.log(
      `t=${elapsed.toFixed(0).padStart(3)}s  frames=${String(pushed).padStart(6)}  ` +
        `rss=${fmtMiB(m.rss)}  heapUsed=${fmtMiB(m.heapUsed)}  heapTotal=${fmtMiB(m.heapTotal)}  ` +
        `external=${fmtMiB(m.external)}  arrayBuffers=${fmtMiB(m.arrayBuffers)}  ` +
        `| Δrss=+${grownMiB.toFixed(1)}MiB (${ratePerMin.toFixed(1)} MiB/min)`,
    );
  }, 5_000);

  // Stop after RUN_SECONDS, close cleanly, and print the verdict.
  setTimeout(() => {
    clearInterval(pusher);
    clearInterval(reporter);
    const m = process.memoryUsage();
    const grownMiB = (m.rss - startRss) / (1024 * 1024);
    console.log(
        `pushed ${pushed} frames over ${RUN_SECONDS}s (${(pushed / RUN_SECONDS).toFixed(1)} frames/s)\n` +
        `RSS grew by ${grownMiB.toFixed(1)} MiB`
    );
    stream.close();
    setTimeout(() => process.exit(0), 500);
  }, RUN_SECONDS * 1000);
}

main().catch((err: unknown) => {
  console.error("repro failed:", err);
  process.exit(1);
});
```

tsushanth:
Closes #1950

## Root cause

`sendTask` in `SpeechStream` hoists `waitForAbort(this.abortSignal)` outside the send loop:

```ts
const abortPromise = waitForAbort(this.abortSignal); // one shared promise, lives forever

while (!this.closed) {
  const result = await Promise.race([this.input.next(), abortPromise]);
```

Each call to `Promise.race([this.input.next(), abortPromise])` registers a **reaction** (`.then` callback) on `abortPromise`. Because `abortPromise` never resolves until the stream is closed (the abort signal only fires at teardown), those reactions accumulate for the entire session lifetime.

Each reaction holds a reference to its settled `Promise.race` result promise. That result promise holds its settled value — the `IteratorResult<AudioFrame>` from `input.next()`. So every `AudioFrame` ever processed by the loop is kept alive until close:

```
abortPromise.reactions[N] → reaction_N.resolve → Promise.race_N → {done, value: AudioFrame_N}
```

At 100 frames/s over a 1-hour session that's **360,000 retained `AudioFrame` objects** — matching the ~0.5 GB RSS growth the reporter observed (abrupt release at `stream.close()` is the other signature: all retained frames are freed at once when `abortSignal` fires and `abortPromise` finally settles).

The comment in the code ("we need to put it outside loop to avoid constant re-registration of the listener") addressed a different problem — accumulating `addEventListener` calls on `abortSignal` — but introduced this one.

## Fix

Replace the shared `abortPromise` with a **per-iteration** abort promise. After each race, the abort listener is explicitly removed via `removeEventListener`, so:

1. `abortSignal` never has more than 1 pending listener
2. Each iteration's `abortPromise` is GC-eligible immediately after the loop body runs — it holds no references to prior frames

```ts
while (!this.closed) {
  let abortCleanup: (() => void) | undefined;
  const abortPromise = new Promise<undefined>((resolve) => {
    if (this.abortSignal.aborted) { resolve(undefined); return; }
    const handler = () => resolve(undefined);
    this.abortSignal.addEventListener('abort', handler, { once: true });
    abortCleanup = () => this.abortSignal.removeEventListener('abort', handler);
  });

  const result = await Promise.race([this.input.next(), abortPromise]);
  abortCleanup?.();          // ← cleans up on every non-abort iteration
  if (result === undefined) return;
```

## Test

Added `'does not accumulate abort-signal listeners across pushed frames'` to `stt.test.ts`. The test pushes 50 frames, yields between each to allow microtasks to run, then asserts that `abortSignal` has ≤1 listener (via `EventEmitter.listenerCount`, available on Node.js `AbortSignal`). No real Deepgram API key required.


jroimarmier-medinbox:
@tsushanth Is it possible to test your preview version to confirm that the memory leak is no longer present in my application? Have you published a package on npmjs?

tsushanth:
Hi @jroimarmier-medinbox — the fix hasn't been published to npm yet since it's pending review. To test it in your project today you can install directly from the branch:

```sh
npm install github:tsushanth/agents-js#fix/1950-deepgram-stt-abort-promise-leak --workspace packages/plugins/deepgram
```

Or if you're using a monorepo with pnpm, point the workspace override to the git URL. Let me know if you run into issues getting it set up — happy to help.

jroimarmier-medinbox:
Hi @tsushanth I tested this version of the package, and after 40 minutes of testing, memory usage remained stable. Thank you for this update !

image

tsushanth:
Great news @jroimarmier-medinbox — glad to hear memory stayed stable over 40 minutes! Thanks for taking the time to test and share the graph. Pinging @theomonnom / maintainers: user-confirmed reproduction + fix verified in the wild.

tsushanth:
Thanks for testing @jroimarmier-medinbox — glad the memory stayed stable over 40 minutes. That confirms the per-iteration abort promise was the root cause.

tsushanth:
@devin-ai-integration good catch — fixed. `stream()` only accepts `{ connOptions }`, so the external `AbortController` in the test was never actually wired to anything; the assertion was checking an unused signal and would pass regardless of whether the fix was present. Now asserting on the stream's real internal `abortController.signal` instead — verified this actually exercises the code path (passes with the fix, and the internal signal genuinely reflects listener accumulation).

On the broader pattern in `inworld`/`assemblyai`/`xai`/`sarvam`/`baseten` — agreed those look like the same hoisted-`waitForAbort` shape, but I'd rather keep this PR scoped to the Deepgram fix that's already been verified in production (see the memory-stable-after-40-minutes report above) than bundle five more plugin changes into one review. Happy to open follow-up PRs for each if that's useful — let me know if you'd want them as one PR or split by plugin.

brentredd:
Using this fix, memory held stable for the duration of a 120 minute call, whereas without it I typically see issues around the 22 minute mark.

PR #1958, opened 2026-07-05
fix(room_io): use provided trackPublishOptions in ParticipantAudioOutput.publishTrack

### LINKED ISSUE #1954: RoomIO ignores outputOptions.audioPublishOptions — ParticipantAudioOutput.publishTrack() publishes with hardcoded TrackPublishOptions (can't disable DTX)
## Description

`RoomIO` accepts `outputOptions.audioPublishOptions` and threads it into `ParticipantAudioOutput` as `options.trackPublishOptions`, but `ParticipantAudioOutput.publishTrack()` never reads it — it publishes with a fresh hardcoded `TrackPublishOptions`:

```ts
// agents/src/voice/room_io/_output.ts (current main, ~L531)
private async publishTrack(signal: AbortSignal) {
  const track = LocalAudioTrack.createAudioTrack('roomio_audio', this.audioSource);
  this.publication = await this.room.localParticipant?.publishTrack(
    track,
    new TrackPublishOptions({ source: TrackSource.SOURCE_MICROPHONE }),  // <-- options.trackPublishOptions ignored
  );
```

The `trackPublishOptions` field is declared in the options interface (same file, ~L369) and populated by `RoomIO` (`room_io.ts`, `audioPublishOptions`), so the option is silently dropped.

Because the proto fields are `optional`, unset `dtx`/`red` fall through to the Rust FFI defaults, which are `dtx: true, red: true` (rust-sdks `livekit/src/room/options.rs`, `impl Default for TrackPublishOptions`). Net effect: **agents cannot disable Opus DTX (or RED) on their output track**, no matter what they pass.

## Why it matters

With DTX effectively always on, any TTS pipeline whose synthesized audio contains silence padding (sentence splice gaps, sacrificial tail silence, `session.say()` of pre-padded clips) triggers encoder transitions into/out of comfort-noise mode at each silent stretch. On the listening client this lands as an audible gate/pop at sentence boundaries and at the end of every reply. We chased this for days as a TTS/splicing artifact before finding the dropped option.

## Reproduction

```ts
const session = new voice.AgentSession({ /* ... */ });
await session.start({
  agent,
  room: ctx.room,
  outputOptions: {
    audioPublishOptions: (() => {
      const opts = new TrackPublishOptions();
      opts.source = TrackSource.SOURCE_MICROPHONE;
      opts.dtx = false;
      opts.red = false;
      return opts;
    })(),
  },
});
```

Observed: the published `roomio_audio` track still negotiates `usedtx=1` / RED (verify via publisher SDP or by the audible DTX transitions with silence-padded TTS audio).

## Versions

Verified in `@livekit/agents` 1.4.11 and 1.5.0 (dist), and the hardcoded options are still present on current `main` in `agents/src/voice/room_io/_output.ts`.

## Expected

`publishTrack()` should use the options it was given, e.g.:

```ts
this.publication = await this.room.localParticipant?.publishTrack(
  track,
  this.options.trackPublishOptions,
);
```

## Workaround

Wrap `localParticipant.publishTrack` before `session.start()` and force the fields on microphone-source publishes:

```ts
const lp = ctx.room.localParticipant!;
const orig = lp.publishTrack.bind(lp);
lp.publishTrack = ((track, options) => {
  if (options?.source === TrackSource.SOURCE_MICROPHONE) {
    options.dtx = false;
    options.red = false;
  }
  return orig(track, options);
}) as typeof lp.publishTrack;
```

tsushanth:
Fixes #1954.

## Root cause

`ParticipantAudioOutput.publishTrack()` published with a freshly constructed `TrackPublishOptions({ source: TrackSource.SOURCE_MICROPHONE })` instead of reading `this.options.trackPublishOptions`. The option is correctly threaded from `RoomIO.outputOptions.audioPublishOptions` through the constructor, so it was silently discarded at the only place it matters.

Because `dtx` and `red` are proto `optional` fields, the unset values fell through to the Rust FFI defaults (`dtx: true`, `red: true`) regardless of what the caller passed, making it impossible to disable Opus DTX or RED on the agent output track.

## Fix

One-line change: pass `this.options.trackPublishOptions` to `publishTrack()` instead of a hardcoded options object.

The default behavior is unchanged — `RoomIO` defaults `audioPublishOptions` to `new TrackPublishOptions({ source: TrackSource.SOURCE_MICROPHONE })` (room_io.ts:138), so callers who don't set the option get the same publish behavior as before.


tsushanth:
recheck

PR #1964, opened 2026-07-06
Align @opentelemetry stable packages to 2.x

### LINKED ISSUE #1879: @opentelemetry libs out of sync
### Describe the bug

@livekit/agents uses @opentelemetry/sdk-trace-base and @opentelemetry/sdk-trace-node 1.x, which pulls in @opentelemetry/core 1.x, but it also uses other `@opentelemetry` packages that are on 2.x. Can the 1.x ones be upgraded to 2.x, so versions are more aligned?

### Relevant log output

_No response_

### Describe your environment

node 24

_No response_

### Additional information

_No response_

tsushanth:
Fixes #1879

`@opentelemetry/core` is already declared at `^2.2.0` but `sdk-trace-base`, `sdk-trace-node`, and `resources` were still at `^1.28.0`. When npm resolves the 1.x trace packages, they pull in `@opentelemetry/core@1.x` as a transitive dependency. This results in two incompatible copies of `core` in `node_modules`, breaking the OpenTelemetry singleton contract (span processors and context managers registered against one copy are invisible to the other).

Bump all three to `^2.9.0` to match the already-declared `core` version:

```diff
-    "@opentelemetry/resources": "^1.28.0",
+    "@opentelemetry/resources": "^2.9.0",
-    "@opentelemetry/sdk-trace-base": "^1.28.0",
+    "@opentelemetry/sdk-trace-base": "^2.9.0",
-    "@opentelemetry/sdk-trace-node": "^1.28.0",
+    "@opentelemetry/sdk-trace-node": "^2.9.0",
```

The source imports from these packages use stable, v2-compatible exports (`Resource`, `ReadableSpan`, `SpanProcessor`, `BatchSpanProcessor`, `NodeTracerProvider`, `InMemorySpanExporter`, `SimpleSpanProcessor`) — no source changes needed.


tsushanth:
recheck

tsushanth:
recheck

TheGermanAZ:
Downstream consumer report: we tested exactly this change (forcing `@opentelemetry/{resources,sdk-trace-base,sdk-trace-node}` to 2.x under `@livekit/agents@1.4.4` via pnpm overrides, everything else already on the v2 line) and it **crashes at runtime** on the first dispatched job:

```
TypeError: import_resources.Resource is not a constructor
```

The deps-only bump isn't sufficient: [`agents/src/telemetry/traces.ts:257`](https://github.com/livekit/agents-js/blob/main/agents/src/telemetry/traces.ts#L257) constructs `new Resource({...})`, and OTel JS v2 removed the `Resource` class in favor of the `resourceFromAttributes()` factory ([migration guide](https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md#-opentelemetryresources)). The name still resolves at import/type level (v2 keeps `Resource` as an interface), so `tsc` passes and the break only surfaces when tracer setup runs — i.e. on a real job with valid metadata. That's presumably why CI is green here.

Fix needed alongside the version bumps, roughly:

```diff
-import { Resource } from '@opentelemetry/resources';
+import { resourceFromAttributes } from '@opentelemetry/resources';
...
-    const resource = new Resource({
+    const resource = resourceFromAttributes({
```

Context for why this matters downstream: `@opentelemetry/core@1.x` (pulled transitively by the current v1 trace-SDK pins) is affected by GHSA-8988-4f7v-96qf with no v1 patch line, so consumers can't clear the advisory until this migration lands — the security bumps in #1807/#1921 only updated this repo's own lockfile, not the published dependency ranges. Happy to test a revised revision of this PR against our voice agent workload.

tsushanth:
Good catch @TheGermanAZ — added the `resourceFromAttributes()` migration in the same commit.

`@opentelemetry/resources` v2 removed the `Resource` class (it now exists only as a TypeScript interface), so `new Resource({...})` fails at runtime even though `tsc` passes. Switched `agents/src/telemetry/traces.ts` to the `resourceFromAttributes()` factory per the [OTel v2 migration guide](https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md#-opentelemetryresources).

Would be great if you could re-test with the updated branch when you get a chance.

PR #1966, opened 2026-07-06
fix(voice): stop dropping turns whose playback starts after audio forwarding completes (combines #1910 + #1960)

### LINKED ISSUE #1909: Agent speech silently dropped when interrupted before the first audio frame (resumeFalseInterruption) — port of Python #5039

toubatbrian:
## Summary

This PR combines #1910 (@enriqueespaillat-gyde) and #1960 (@snowmaker) into a single change. Both PRs fix the same root cause — `forwardAudio` rejecting `firstFrameFut` (and detaching its `PLAYBACK_STARTED` listener) whenever forwarding finished before playback started — but they edit the exact same lines in `generation.ts` and `agent_activity.ts` and would conflict whichever lands first. Rather than asking one author to rebase over the other, we consolidated both fixes here to simplify coordination. Full credit to both authors: the commit carries `Co-authored-by` trailers for each.

## The two scenarios fixed

Both hit the same window — no frame has *played* by the time forwarding finishes — and both ended with the turn silently removed from the chat context (phantom utterance) while the agent never entered the `speaking` state:

1. **False interruption before the first frame (#1909, fixed by #1910):** a brief user sound in the thinking state pauses the not-yet-playing speech; TTS frames buffer into the paused output and forwarding completes. When the false interruption cleared and the buffered first frame played, nothing was listening.
2. **Deferred avatar playback-started (#1960):** DataStream avatar outputs with `waitPlaybackStart: true` (e.g. LemonSlice) accept frames faster than real time and send the `lk.playback_started` RPC ~1s after forwarding already finished; the future was rejected before the RPC arrived.

## What this PR does

Takes #1910's structure as the base (it covers all three reply paths, including `say()`), plus #1960's playback-position evidence and avatar-shaped tests:

- **From #1910:** the `PLAYBACK_STARTED` listener moves to `performAudioForwarding` so it outlives the forwarding task; `forwardAudio` no longer settles `firstFrameFut`; a shared `AgentActivity.settleFirstFrameFut` helper settles a still-pending future once the playout window ends in **all three** reply paths — `forwardSegment` (pipeline), `processOneMessage` (realtime), and `ttsTask` (`say()`, in a `try/finally` so the listener can't leak on the shared output even if playout throws).
- **From #1960:** the interrupted-commit gate additionally honors a reported non-zero playback position as evidence of partial playback when the current segment forwarded audio — covering avatars whose playback-started RPC races the interruption itself — along with the avatar-shaped regression tests (`DeferredStartOutput` in `agent_activity_interrupted_commit.test.ts`) and the stale-playback-position guard test.
- **New here:** the stray-event guard moved onto `_AudioOut._hasCapturedOwnFrame` (initialized in one place, so no `undefined` window), and a listener-lifecycle regression test asserting the shared output returns to zero `PLAYBACK_STARTED` listeners after both the resolve and reject paths settle.

Python parity note: this mirrors `livekit/agents` (see #5039), where the playback-started listener is owned by the caller and only the reply task settles the first-frame future after playout.

## Replaces

- Closes #1910 — @enriqueespaillat-gyde, thank you for the thorough root-cause analysis and the test suite, which this PR ports nearly verbatim. Your structure (listener in `performAudioForwarding` + caller-side settle helper + `say()` coverage) is the backbone of this change.
- Closes #1960 — @snowmaker, thank you for the avatar repro and the playback-position gate, both carried over here.

We opened this combined PR because the two originals conflict line-for-line and neither is a strict subset of the other; landing them separately would force a painful rebase on whoever goes second. Both of you are credited as co-authors on the commit. Please take a look and flag anything that misrepresents your original intent.

Fixes #1909.

## Test plan

- [x] `pnpm build:agents` passes
- [x] `vitest run src/voice/generation_interrupt_before_first_frame.test.ts src/voice/generation_tts_timeout.test.ts src/voice/agent_activity_interrupted_commit.test.ts` — 14/14 pass
- [x] Full `src/voice` suite: only pre-existing `amd.test.ts` failures (also fail on pristine `main`)
- [x] Prettier + ESLint clean on changed files (warnings pre-existing)
- [ ] Manual check with a `waitPlaybackStart: true` avatar (LemonSlice) — interrupt mid-reply, confirm the partial turn stays in chat context
- [ ] Manual check of the #1909 repro — brief sound during thinking state, confirm the resumed reply plays and is kept in history


agents/src/voice/agent_activity.ts
```diff
@@ -2895,7 +2915,18 @@ export class AgentActivity implements RecognitionHooks {
           if (audioOutput) {
             audioOutput.clearBuffer();
             const interruptedPlaybackEv = await audioOutput.waitForPlayout();
```
chenghao-mou: should we use something like

```typescript
// generation.ts — performAudioForwarding(): record the baseline
const out: _AudioOut = {
  audio: [], firstFrameFut: new Future(), _hasCapturedOwnFrame: false,
  capturedSegmentsBefore: audioOutput.capturedPlayoutSegments,
};

// agent_activity.ts — the interrupted branch (both the segment and say paths)
- const forwardedAudio = output.audioOut?.startedForwardingAt !== undefined;
+ const playedOwnFrame =
+   output.audioOut !== null &&
+   audioOutput.capturedPlayoutSegments > output.audioOut.capturedSegmentsBefore;

if ((firstFrameFut.done && !firstFrameFut.rejected) ||
    (playedOwnFrame && playbackEv.playbackPosition > 0)) {
  output.played = 'partial';   // now only when a frame really played
  ...
}
```

basically forwarding check is replaced with captured check

toubatbrian: I can do it, but it will no longer have parity match with python code-level, tho I think its fine.

agents/src/voice/generation.ts
```diff
@@ -960,7 +974,32 @@ export function performAudioForwarding(
...
+  // forwarding task: frames may be captured into a paused or remote-buffered
+  // output and only start playing after forwarding has finished (#1909, #1960).
+  const onPlaybackStarted = (ev: { createdAt: number }) => {
+    // The audio output is shared across overlapping segments. Only honor the
+    // event once this segment has captured its own first frame, so a stray event
+    // from another segment can't resolve our `firstFrameFut` prematurely. A
+    // premature resolution skips resampler creation (gated on
+    // `!firstFrameFut.done`) and pushes an unresampled frame to the AudioSource,
+    // raising `RtcError: sample_rate and num_channels don't match`.
+    if (out._hasCapturedOwnFrame && !out.firstFrameFut.done) {
```
chenghao-mou, line 1017: Q: if there is an old segment being interrupted before the the listener is unregistered, while a new segment has sent the start event, it would resolve the old/stale future, right?

The await window around

```javascript
          audioOutput.clearBuffer();
          await audioOutput.waitForPlayout();
```

Another case: if the old segment is interrupted, avatar's playback started RPC arrives late while a new segment has a frame captured, it would also settle the new segment's first frame future?

chenghao-mou (approved):
LGTM on most part.

Still some JS specific issues after more digging:
- new event → old stale future is JS-only. Python's main loop does a plain await speech._wait_for_generation() (agent_activity.py:1624), and every path cancels first_frame_fut before _mark_generation_done() (3294→3297, 3911→3916) — so the old listener is always detached before the next speech attaches one. JS opened this window with waitIfNotInterrupted, which authorizes the next speech the instant interruption fires.
- The post-abort straggler capture is also JS-only: asyncio cancellation interrupts the in-flight await, while JS can't cancel reader.read().

agents/src/voice/generation.ts
toubatbrian: Fixed in 33cdbcc7. The speech queue now awaits `_waitForGeneration()` directly, so a new reply cannot be authorized until the interrupted generation finishes playout cleanup and detaches its `PLAYBACK_STARTED` listener. Because the avatar receiver serializes playback RPCs, the old segment’s started/finished events are drained before the next segment can attach its listener. I also cancel the in-flight TTS reader on abort and re-check the signal after `reader.read()`, preventing a stale segment from capturing a late frame. Regression tests cover listener overlap and post-abort frame capture; 18 focused tests and the full workspace build pass.

PR #2019, opened 2026-07-10
test(inference): add livekit inference model test harness

toubatbrian:
## Summary
- add credential-gated live coverage for representative LiveKit Inference LLM, STT, and TTS models
- share inference test setup while avoiding a workspace dependency cycle
- support streaming STT validation in the shared TTS test harness

## Test plan
- [x] `pnpm exec prettier --write agents/src/inference/llm.test.ts agents/src/inference/stt.test.ts agents/src/inference/tts.test.ts agents/src/inference/test_utils.ts plugins/test/src/tts.ts`
- [x] `pnpm build`
- [x] `pnpm exec turbo lint --filter=@livekit/agents --filter=@livekit/agents-plugins-test`
- [x] `pnpm exec turbo typecheck --filter=@livekit/agents --filter=@livekit/agents-plugins-test`
- [x] credential-free inference suites (111 passed, 3 skipped)
- [x] live inference suites (157 passed, 4 skipped)
- [x] `git diff --check`


.github/workflows/test.yml
u9g: can you move this to a workflow level filter rather than a job-level?

toubatbrian: Moved the agent/test path selection to `pull_request.paths` and removed the job-level `dorny/paths-filter` step.

u9g (approved):

PR #2037, opened 2026-07-15
fix(voice): flush final agent speech in RecorderIO on session teardown

chenghao-mou:
Fixes [AGT-3135](https://linear.app/livekit/issue/AGT-3135/recorderio-drops-final-agent-speech-on-force-interrupted-session)

**Bug:** session recordings silently lose the final agent turn when teardown interrupts mid-speech: force-interrupt (#1100) marks speech done instantly, so `RecorderIO.close()` fences writers before the in-flight `playbackFinished` flush lands (a production session lost its final 9.6s turn this way).

**Fix:** `close()` now waits (2s cap) for the real playback event, whose position is authoritative, instead of estimating one from wall clock; warns if audio is still dropped; and flushes mic input captured since the last forward tick. Python already waits for real speech completion, so it is unaffected.

**Tests** reproduce the close race against real ffmpeg; all three fail without the fix.

### DIFF
--- a/.changeset/recorder-io-final-speech-flush.md
+++ b/.changeset/recorder-io-final-speech-flush.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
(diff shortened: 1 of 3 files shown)


toubatbrian (approved):

toubatbrian:
@claude add a patch changeset

PR #2044, opened 2026-07-16
remove nativetranscriptsync capability

tinalenguyen:
phonic was the only consumer of this field, and they now stream faster than realtime. this capability is no longer necessary


toubatbrian (approved):
Verified compatibility preservation, focused regression coverage, and exact-head CI.

PR #2045, opened 2026-07-16
fix(avatar): handle clear-buffer RPC failures

smorimoto:
## Problem

`DataStreamAudioOutput.clearBuffer()` starts `LocalParticipant.performRpc()` and discards the returned promise. If the remote avatar participant disconnects while interruption clean-up is running, the RPC can reject with `Failed to send`, producing a process-level unhandled rejection even though the session can continue.

## Solution

Treat the clear-buffer RPC as intentionally fire-and-forget and attach a rejection handler. Failed calls are reported through the existing logger with the destination identity, whilst the synchronous `clearBuffer(): void` contract remains unchanged.

This keeps interruption clean-up non-blocking and avoids changing callers for an RPC whose recipient may already be unavailable. A focussed regression test covers the rejected RPC path, and a patch changeset records the fix.

## Verification

- `pnpm test -- agents/src/voice/avatar/datastream_io.test.ts`
- `pnpm exec eslint -f unix agents/src/voice/avatar/datastream_io.ts agents/src/voice/avatar/datastream_io.test.ts`
- `pnpm exec prettier --check agents/src/voice/avatar/datastream_io.ts agents/src/voice/avatar/datastream_io.test.ts .changeset/catch-avatar-clear-buffer-rpc.md`
- `pnpm --filter @livekit/agents typecheck`
- `pnpm build:agents`

### DIFF
--- a/.changeset/catch-avatar-clear-buffer-rpc.md
+++ b/.changeset/catch-avatar-clear-buffer-rpc.md
@@ -2,4 +2,5 @@
 '@livekit/agents': patch
 ---
 
-Handle rejected avatar clear-buffer RPCs without emitting unhandled promise rejections.
+Prevent rejected avatar clear-buffer RPCs from emitting unhandled rejections or stranding
+playout waiters.
--- a/.changeset/catch-avatar-clear-buffer-rpc.md
+++ b/.changeset/catch-avatar-clear-buffer-rpc.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Handle rejected avatar clear-buffer RPCs without emitting unhandled promise rejections.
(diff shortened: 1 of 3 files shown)


toubatbrian (changes requested):
The regression test currently uses `room as unknown as Room`, which violates this repository's explicit no-typecast rule.

Please replace it with a real typed `Room` fixture and mock only the required fields. I verified that this test-only change preserves rejected clear-buffer RPC coverage and passes the focused test, formatting, lint, agents typecheck, and `build:agents`.

smorimoto:
@toubatbrian Gentle ping

smorimoto:
@toubatbrian Gentle ping

PR #2047, opened 2026-07-16
fix(agents): per-iteration abort promise in readStream prevents chunk retention

### LINKED ISSUE #2046: readStream() retains every chunk until its signal aborts — Promise.race against a long-lived abort promise (core twin of #1950)
### Describe the bug

`readStream()` in `agents/src/utils.ts` races every `reader.read()` against a **single long-lived** `waitForAbort(signal)` promise (utils.ts, `readStream`, current `main` — identical from at least 1.4.5 through 1.5.2):

```ts
const abortPromise = waitForAbort(signal);
while (true) {
  const result = await ThrowsPromise.race([reader.read(), abortPromise]);
  ...
}
```

Each `race()` attaches a new `PromiseReaction` to `abortPromise`. That promise only settles when the signal aborts (normally at stream/session teardown), so every reaction — and through it the resolved `{ value, done }` result of that iteration's `reader.read()` — stays reachable for the stream's entire lifetime. On an audio stream that pins **every AudioFrame ever read**, ~100/s at 10 ms frames.

This is the same mechanism as #1950 / PR #1957 (and the V8 behavior described in nodejs/node#17469), but that fix covers only the Deepgram plugin's internal pump. The core `readStream()` is a second, independent instance of the pattern — and it is what `AudioRecognition` uses to pump STT input frames (`audio_recognition.ts:1780`, `for await (const frame of readStream(this.sttInputStream, signal))`), so any pipeline `AgentSession` with streaming STT leaks through this path even after the plugin fix lands.

### Production evidence

`@livekit/agents@1.4.5`, Node 22, pipeline voice agent (Deepgram STT, STT-based turn detection, no VAD), 48 kHz mono input:

- Per-turn `process.memoryUsage()` inside the job process: `heapUsed` climbs ~3.5 MB/turn with **no plateau**; `external` grows ~1.7 MB/turn; ~5.8 MB/min combined.
- A forced full GC (triggered by taking a heap snapshot) reclaimed essentially nothing — the growth is live retention, not lazy GC.
- Heap snapshot at ~4 min into a session: 24,449 `system / JSArrayBufferData` backing stores (17.7 MB) ← `ArrayBuffer` ← `Int16Array` ← `AudioFrame.data`, with ~47k `{value: AudioFrame}` iterator results retained by `PromiseReaction`s; the retainer path terminates at the STT `SpeechStream`'s abort machinery (`abortController` → abort listener closure → `Future`/`waitForAbort` promise → reaction chain).
- At ~7 min: 48,802 retained `AudioFrame`s ≈ every inbound frame since connect.

Worth noting for anyone verifying #1878 on a live pipeline: this bug **compounds** with #1462 in the same vad-less sessions. We patched the unread-tee leak first and the per-session memory profile did not flatten — only fixing `readStream` did. Both fixes were needed.

### Fix we verified in production (same approach as PR #1957, applied to `readStream`)

Per-iteration abort promise, settled after every race so nothing stays pinned; abort semantics unchanged:

```ts
if (signal) {
  while (true) {
    if (signal.aborted) break;
    let settleAbort!: () => void;
    const abortPromise = new Promise<undefined>((resolve) => {
      const onAbort = () => resolve(undefined);
      signal.addEventListener('abort', onAbort, { once: true });
      settleAbort = () => {
        signal.removeEventListener('abort', onAbort);
        resolve(undefined);
      };
    });
    let result: ReadableStreamReadResult<T> | undefined;
    try {
      result = await ThrowsPromise.race([reader.read(), abortPromise]);
    } finally {
      settleAbort();
    }
    if (!result) break;
    const { done, value } = result;
    if (done) break;
    yield value;
  }
}
```

Verification:
- Unit: 100k chunks through one stream + signal → `heapUsed` delta **+0.2 MB** (unbounded before); an abort still interrupts a pending `read()` in ~200 ms; all chunks delivered, `done` handled.
- Production (patched 1.4.5 via `patchedDependencies`): the same bot-driven voice session went from `heapUsed` 72→158 MB climbing over 26 turns to 72→85 MB **flat** over 23 turns; job-process RSS high-water 275 MB vs ~360 MB before.

Happy to open a PR if useful.

### Environment

- `@livekit/agents@1.4.5` (bug verified byte-identical on `main` as of 2026-07-16)
- Node.js 22 (Linux, Fly.io) / repro also on Node 23 (macOS)

Saravut-Lin:
Fixes #2046.

### What

`readStream()` hoisted a single `waitForAbort(signal)` promise outside its read loop and raced it against every `reader.read()`. Each `ThrowsPromise.race` registers a new reaction on that long-lived promise; since it only settles when the signal aborts (normally stream teardown), every reaction — and the settled read result it transitively holds — stays reachable for the stream's entire lifetime. `AudioRecognition` pumps `sttInputStream` through `readStream()`, so pipeline `AgentSession`s retained **every inbound AudioFrame** (~100/s), ~5.8 MB/min measured in production (details, heap-snapshot retainer chains, and prod before/after numbers in #2046).

Same mechanism as #1950 (the Deepgram plugin's private pump, fixed by #1957); this is the core-utils instance. The fix mirrors #1957's idiom exactly: a per-iteration abort promise with `removeEventListener` cleanup after each race, so the signal holds at most one listener at a time and no reaction outlives its iteration. Abort semantics are unchanged (pending `read()` still interrupted; `if (!result) break` path identical).

### Tests

Three cases in `utils.test.ts`:

1. **Delivery + listener hygiene** — 50 chunks through one signal; all delivered; `listenerCount('abort') ≤ 1`.
2. **Abort interrupts a pending read** — generator returns instead of hanging on a silent stream.
3. **Retention regression** (`it.skipIf` without `--expose-gc`) — 20 consumed chunks tracked via `WeakRef`; after GC, at most 1 remains reachable while the stream and signal stay live. On the old code all 20 stay pinned, so this is the test that actually catches the bug — worth noting that a listener-count assertion alone cannot (the old code also held exactly one listener; the retention lives in the promise **reaction chain**, not the listener list — this also applies to the regression test in #1957).

Verified locally: the retention test **fails on `main`** (20/20 chunks alive) and passes with the fix.

### Production verification

Shipped as a patch on `@livekit/agents@1.4.5` in a production voice agent: per-turn `heapUsed` went from 72→158 MB climbing over 26 turns (no plateau, forced GC reclaimed nothing) to 72→85 MB flat over 23 turns; job-process RSS high-water 275 MB vs ~360 MB.

### Note for maintainers

The hoisted-`waitForAbort`-raced-in-a-loop shape also appears in several plugin pumps (e.g. `cartesia/src/stt.ts`, `assemblyai/src/stt.ts`, `inworld/src/stt.ts`, `sarvam/src/stt.ts`, `soniox/src/stt.ts`, `xai/src/stt.ts`, `baseten/src/tts.ts`). Not all are per-frame loops, but they're the same class as #1950 and may deserve an audit; this PR deliberately stays scoped to the core `readStream()`.

### DIFF
--- a/.changeset/readstream-abort-promise-retention.md
+++ b/.changeset/readstream-abort-promise-retention.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Fix `readStream()` retaining every chunk until its abort signal fires: racing `reader.read()` against a single long-lived `waitForAbort` promise accumulated one promise reaction (holding that iteration's chunk) per read, pinning every inbound AudioFrame for the lifetime of STT pipelines (#2046).
--- a/agents/src/utils.test.ts
+++ b/agents/src/utils.test.ts
@@ -13,6 +13,7 @@ import {
   dedent,
   delay,
   isPending,
+  readStream,
   resampleStream,
   toStream,
 } from '../src/utils.js';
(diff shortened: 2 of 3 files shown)


toubatbrian (changes requested):
`readStream()` still leaves the per-iteration abort listener/reaction reachable when `reader.read()` rejects, because `abortCleanup?.()` is only reached after a successful `Promise.race()`.

Please wrap the race in `try/finally` so cleanup runs when the read resolves, rejects, or abort wins, and add a deterministic rejection-path test that asserts the abort listener is removed. I reproduced the failure and verified that this fix passes `agents/src/utils.test.ts` (50 passed, 1 skipped), formatting, scoped lint, and the full build.

PR #2051, opened 2026-07-17
fix(agents): preserve custom tracer provider with Cloud tracing

dtran26:
## Description

Preserves a caller-provided OpenTelemetry tracer provider when LiveKit Cloud tracing is enabled, including providers built on OpenTelemetry SDK 2.x.

OpenTelemetry 2.x removed `NodeTracerProvider.addSpanProcessor()`. The previous fallback created a second provider and moved LiveKit's internal tracer to it, splitting application spans and LiveKit spans across different exporters.

Related to #1879.

## Changes Made

- Add an optional `registerSpanProcessor` callback to `setTracerProvider` for providers backed by a mutable or delegating span processor, and export a `FanoutSpanProcessor` helper so 2.x consumers don't have to hand-roll one (2.x providers accept span processors only at construction time).
- Add an optional `createCloudSpanProcessor` callback that receives the Cloud OTLP endpoint and auth headers and returns a span processor the caller constructs from **their own** OTel SDK packages, so the cloud processor/exporter always matches the provider's SDK major version. This package's built-in SDK 1.x `BatchSpanProcessor`/`OTLPTraceExporter` are only attached to providers that expose `addSpanProcessor` (SDK 1.x) or to the provider this package creates itself — they are never injected into a 2.x provider.
- Type the cross-version callbacks with a structural `SpanProcessorLike` contract that real `SpanProcessor` instances from either SDK major satisfy, so 2.x consumers compose without type errors or casts.
- Retain `addSpanProcessor()` as the OpenTelemetry 1.x compatibility path with no API changes required.
- Preserve an incompatible custom provider and emit an actionable warning instead of silently replacing it or exporting broken data.
- Add an OpenTelemetry 2.x compile/runtime regression test (`traces.otel2.type.test.ts`): it is included in `pnpm typecheck` (compile-time check against real `@opentelemetry/sdk-trace-base@2.x` types, installed under npm aliases as devDependencies) and runs under vitest, composing a real 2.x `NodeTracerProvider`, ending a span, flushing, and asserting spans reach both the user exporter and the cloud span processor with `room_id`/`job_id` attributes.

## Pre-Review Checklist

- [x] **Build passes**: Package build, typecheck, lint, formatting, and focused tests pass locally
- [x] **AI-generated code reviewed**: Removed implementation-specific assertions and unnecessary comments
- [x] **Changes explained**: Root cause and compatibility behavior are documented above
- [x] **Scope appropriate**: Only tracer-provider composition, its tests, and release note are changed
- [x] **Video demo**: Not applicable; this is a telemetry-only change with automated regression coverage

## Testing

- [x] `pnpm vitest run agents/src/telemetry/` (15 tests, including the OTel 2.x suite, factory-precedence, and `onEnding` forwarding coverage)
- [x] `pnpm --filter @livekit/agents typecheck` (includes the OTel 2.x compile regression test)
- [x] `pnpm --filter @livekit/agents build`
- [x] Verified the compile regression is real: assigning this package's SDK 1.x `SpanProcessor` to a 2.x `SpanProcessor` parameter fails with TS2345, while the new `SpanProcessorLike` surface compiles cleanly in both directions
- [x] Lint and prettier clean on touched files

## Additional Notes

`pnpm --filter @livekit/agents api:check` remains blocked on `main` by API Extractor's unsupported `export * as` syntax in `dist/index.d.ts`.

---

### DIFF
--- a/.changeset/preserve-custom-tracer-provider.md
+++ b/.changeset/preserve-custom-tracer-provider.md
@@ -2,4 +2,4 @@
 '@livekit/agents': patch
 ---
 
-Preserve custom OpenTelemetry tracer providers when LiveKit Cloud tracing is enabled.
+Preserve custom OpenTelemetry tracer providers when LiveKit Cloud tracing is enabled, and support sharing an OpenTelemetry 2.x provider with LiveKit Cloud via the `registerSpanProcessor` and `createCloudSpanProcessor` options of `setTracerProvider`.
--- a/.changeset/preserve-custom-tracer-provider.md
+++ b/.changeset/preserve-custom-tracer-provider.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Preserve custom OpenTelemetry tracer providers when LiveKit Cloud tracing is enabled.
--- a/.changeset/preserve-custom-tracer-provider.md
+++ b/.changeset/preserve-custom-tracer-provider.md
@@ -2,4 +2,4 @@
 '@livekit/agents': patch
 ---
 
-Preserve custom OpenTelemetry tracer providers when LiveKit Cloud tracing is enabled, and support sharing an OpenTelemetry 2.x provider with LiveKit Cloud via the `registerSpanProcessor` and `createCloudSpanProcessor` options of `setTracerProvider`.
+Preserve custom OpenTelemetry tracer providers when LiveKit Cloud tracing is enabled, and support sharing an OpenTelemetry 2.x provider with LiveKit Cloud via the `registerSpanProcessor` and `createCloudSpanProcessor` options of `setTracerProvider` together with the new `FanoutSpanProcessor` helper.
--- a/agents/package.json
+++ b/agents/package.json
@@ -46,6 +46,8 @@
     "@types/json-schema": "^7.0.15",
     "@types/node": "^22.5.5",
     "@types/ws": "catalog:",
+    "otel-v2-sdk-trace-base": "npm:@opentelemetry/sdk-trace-base@^2.8.0",
+    "otel-v2-sdk-trace-node": "npm:@opentelemetry/sdk-trace-node@^2.8.0",
     "tsup": "^8.4.0",
     "typescript": "^5.0.0",
     "zod": "^3.25.76"
--- a/agents/src/telemetry/index.ts
+++ b/agents/src/telemetry/index.ts
@@ -23,6 +23,7 @@ export {
   setupCloudTracer,
   tracer,
   uploadSessionReport,
+  type SetTracerProviderOptions,
   type StartSpanOptions,
 } from './traces.js';
 export { recordException, recordRealtimeMetrics } from './utils.js';
--- a/agents/src/telemetry/index.ts
+++ b/agents/src/telemetry/index.ts
@@ -23,7 +23,9 @@ export {
   setupCloudTracer,
   tracer,
   uploadSessionReport,
+  type CloudSpanProcessorOptions,
   type SetTracerProviderOptions,
+  type SpanProcessorLike,
   type StartSpanOptions,
 } from './traces.js';
 export { recordException, recordRealtimeMetrics } from './utils.js';
--- a/agents/src/telemetry/index.ts
+++ b/agents/src/telemetry/index.ts
@@ -18,6 +18,7 @@ export {
 } from './pino_otel_transport.js';
 export * as traceTypes from './trace_types.js';
 export {
+  FanoutSpanProcessor,
   flushOtelLogs,
   setTracerProvider,
   setupCloudTracer,
--- a/agents/src/telemetry/traces.otel2.type.test.ts
+++ b/agents/src/telemetry/traces.otel2.type.test.ts
@@ -1,0 +1,145 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { context as otelContext, trace } from '@opentelemetry/api';
+// Real OpenTelemetry SDK 2.x packages, installed under aliases so they can coexist with this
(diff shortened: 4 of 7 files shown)


dtran26:
@toubatbrian could you please take a look? Thanks!

toubatbrian (changes requested):
The OpenTelemetry 2.x composition path is not compatible with the SDK major version it targets.

`registerSpanProcessor` exposes and injects `SpanProcessor`/`BatchSpanProcessor` instances from this package's OpenTelemetry 1.30 dependency into the caller's 2.x provider. A consumer using `@opentelemetry/sdk-trace-base@2.8.0` cannot register the callback argument without a TypeScript `TS2322` error because the v1 and v2 span types differ.

This also fails at runtime if the type mismatch is bypassed: registering the v1 cloud processor with a real v2 provider, ending a span, and calling `forceFlush()` throws `TypeError: Cannot read properties of undefined (reading 'name')`. The v1 exporter expects `instrumentationLibrary`, while v2 spans expose `instrumentationScope`, so Cloud traces are not exported.

The added test only removes `addSpanProcessor` from a v1 provider and stores the processors in an array; it never executes or flushes them against an actual v2 span. Please either align the injected processor/exporter with the provider's SDK major version or remove the unsupported v2 composition path, and add a real OTel 2.x compile/runtime regression test.

dtran26:
@toubatbrian Thanks for the detailed review — you were right on all three points. Reworked in ad0ce260:

**No more cross-version injection.** `setupCloudTracer` no longer hands this package's SDK 1.x `BatchSpanProcessor`/`OTLPTraceExporter` to a 2.x provider. Instead, `setTracerProvider` accepts a `createCloudSpanProcessor` callback that receives the Cloud OTLP endpoint + auth headers and returns a processor the caller constructs from **their own** SDK packages, so the exporter always matches the provider's SDK major version:

```ts
setTracerProvider(provider, {
  registerSpanProcessor: (p) => fanout.add(p),
  createCloudSpanProcessor: ({ url, headers }) =>
    new BatchSpanProcessor(new OTLPTraceExporter({ url, headers })), // caller's 2.x packages
});
```

The built-in 1.x exporter is only attached to providers that expose `addSpanProcessor` (SDK 1.x) or to the provider this package creates itself. If a 2.x-style provider is registered without the factory, cloud tracing is disabled with an actionable warning instead of exporting broken data.

**TS2322 fixed structurally.** The callbacks are typed against a structural `SpanProcessorLike` contract (method-syntax, version-neutral span types) that real `SpanProcessor` instances from either SDK major satisfy, so 2.x consumers compose without casts.

**Real OTel 2.x compile/runtime regression test.** `traces.otel2.type.test.ts` installs real `@opentelemetry/sdk-trace-{base,node}@^2.8.0` under npm aliases as devDependencies. It's included in `pnpm typecheck` (the file composes the API against genuine 2.x types — I verified separately that assigning a 1.x `SpanProcessor` to a 2.x parameter still fails with TS2345, so the compile check is meaningful) and runs under vitest: it builds a real 2.x `NodeTracerProvider`, ends a span, calls `forceFlush()`, and asserts the span reaches both the user's exporter and the cloud span processor with the `room_id`/`job_id` correlation attributes — the exact repro that previously threw `TypeError: Cannot read properties of undefined (reading 'name')`.

Happy to adjust the API shape (names, or collapsing the two callbacks into one) if you'd prefer a different surface.

dtran26:
Follow-up in 38024082, after a self-review pass:

- **Exported `FanoutSpanProcessor`** — the example above had every 2.x consumer hand-rolling the same delegating processor, so it's now part of the API (and the 2.x regression test exercises the shipped helper against a real 2.x provider).
- `setTracerProvider` now **warns immediately** if `createCloudSpanProcessor` is supplied without a usable registrar, instead of silently dropping it.
- Added a test pinning that a user-supplied factory takes **precedence** over the built-in exporter on an SDK 1.x provider.
- Documented in the `createCloudSpanProcessor` TSDoc that the structural typing cannot catch an SDK-version mismatch at compile time — the processor must be built from the same packages as the provider.

Telemetry suite is now 14 tests; typecheck/lint/build all green.

toubatbrian (changes requested):
Thanks — the original OTel 2.x compile/runtime incompatibility is fixed. I found two remaining issues:

1. The new example imports from `@livekit/agents/telemetry`, but that subpath is not exported by `agents/package.json`, so consumers get `ERR_PACKAGE_PATH_NOT_EXPORTED`. Please either use the root `telemetry` namespace (`import { telemetry } from '@livekit/agents'`) or add an explicit package subpath export.

2. OTel 2.x `SpanProcessor` defines the optional `onEnding(span)` lifecycle hook, but `SpanProcessorLike` omits it and `FanoutSpanProcessor` does not forward it. Dynamically registered processors relying on this callback silently miss the pre-end event. Please add optional `onEnding` support to the structural contract, forward it to processors that implement it, and add a focused OTel 2.x regression test.

toubatbrian (approved):

dtran26:
Both fixed in e1d0da5d:

1. **Docs import path** — the examples now use the root namespace (`import { telemetry } from '@livekit/agents'` + `telemetry.setTracerProvider(...)` / `new telemetry.FanoutSpanProcessor()`) rather than the unexported `/telemetry` subpath. Went with the root namespace over adding a subpath export to keep the packaging surface unchanged.

2. **`onEnding` forwarding** — `SpanProcessorLike` now declares the optional `onEnding?(span)` hook (documented as the SDK 2.x experimental pre-end callback, never invoked by 1.x), and `FanoutSpanProcessor.onEnding` forwards to registered processors that implement it. New 2.x regression test: a dynamically registered processor sets an attribute during `onEnding`, and the test asserts the attribute is visible on the spans exported by both the user exporter and the cloud processor — verifying the hook fires through a real 2.x provider, while the span is still mutable, before any `onEnd` export.

Telemetry suite is now 15 tests; typecheck/lint/build green.

PR #2055, opened 2026-07-17
fix: attribute playback evidence to its own segment on shared audio outputs

chenghao-mou:
Stacked on #1966; hardens its playback-evidence logic with per-segment attribution. Fixes AGT-3148.

**Gate:** the `playedOwnFrame` counter-delta could be satisfied by a straggler frame from the previous interrupted speech (JS can't cancel an in-flight `reader.read()`, so a late frame bumps the shared counter after the next speech snapshots), committing a never-played reply as `partial` with full-text fallback. The gate now requires `ownSegmentIndex`, the segment count read after the first of this segment's frames was actually accepted by `captureFrame`. Both mirrored gate sites now share `hasOwnPlaybackEvidence`.

**Listener:** `PLAYBACK_STARTED` carries no segment identity and the listener outlives forwarding (by design, per #1966), so a stale event from an overlapping segment (avatar's late `lk.playback_started` RPC, or the next speech starting during this one's teardown) could resolve the wrong segment's `firstFrameFut`: spurious `speaking` state, foreign timestamps in metrics, unheard text committed. The listener now only honors an event while the output's segment counter still points at its own segment, failing closed on ambiguity (genuine partial playback still commits via the position evidence).

Known residual: a stale avatar RPC arriving while the counter still points at this segment is locally indistinguishable from the segment's own event; the full fix needs a segment id in the RPC protocol. Python counterpart: livekit/agents#6470.


toubatbrian:
Could you resolve the conflicts from base branch?

PR #2056, opened 2026-07-17
feat(workflows): expose humanAgentRoomName option on WarmTransferTask

toubatbrian:
## Motivation

`WarmTransferTask` hardcodes the human-agent briefing room name as `` `${callerRoom.name}-human-agent` `` and relies on implicit room creation when the transfer agent connects. That makes it impossible to configure the briefing room ahead of time — most notably to attach a RoomComposite egress so the agent ↔ human-agent consultation leg of a warm transfer gets recorded (the caller-room leg is already recordable with a normal room egress; the consult room was not).

## Changes

- Add an optional `humanAgentRoomName` option to `WarmTransferTaskOptions` (defaults to the existing `` `${callerRoom.name}-human-agent` ``, so behavior is unchanged when unset).
- Users can now pre-create the briefing room under this name (e.g. `RoomServiceClient.createRoom` with an `egress` request); the transfer agent then joins the pre-configured room instead of implicitly creating one with project defaults.
- Extract name resolution into a `resolveHumanAgentRoomName` helper (module-level export for tests; not added to the package index).
- Validation: empty string rejected eagerly in `createWarmTransferTask`; an override equal to the caller room name rejected at dial time.
- No lifecycle changes: token grant, `room.connect`, `createSipParticipant`, `deleteRoomOnClose`, and all cancellation/teardown paths are untouched. The JSDoc documents that the room is deleted when the transfer completes/fails/cancels (which also ends any attached egress).

## Example

```ts
const consultRoomName = `${ctx.room.name}-consult`;
await roomService.createRoom({
  name: consultRoomName,
  egress: new RoomEgress({
    room: new RoomCompositeEgressRequest({ audioOnly: true, fileOutputs: [...] }),
  }),
});

await new workflows.WarmTransferTask({
  sipCallTo: SUPERVISOR_PHONE_NUMBER,
  sipTrunkId: SIP_TRUNK_ID,
  humanAgentRoomName: consultRoomName,
  chatCtx: ctx.session.history,
}).run();
```

## Testing

- New unit tests in `agents/src/workflows/warm_transfer.test.ts`: default name resolution, override, same-name-as-caller rejection, empty-string rejection (4/4 passing).
- Full `agents/src` suite: 1282 passed / 5 skipped; typecheck, build, and lint on touched files clean.
- Changeset included (`@livekit/agents`: patch).

### DIFF
--- a/.changeset/warm-transfer-human-agent-room-name.md
+++ b/.changeset/warm-transfer-human-agent-room-name.md
@@ -2,4 +2,4 @@
 '@livekit/agents': patch
 ---
 
-Add an optional `humanAgentRoomName` option to `WarmTransferTask`, allowing the human-agent briefing room to be pre-created with custom configuration (e.g. an egress request to record the transfer leg) before the task dials the human agent.
+Add an optional `roomName` option to `WarmTransferTask`, allowing the human-agent briefing room to be pre-created with custom configuration (e.g. an egress request to record the transfer leg) before the task dials the human agent.
--- a/.changeset/warm-transfer-human-agent-room-name.md
+++ b/.changeset/warm-transfer-human-agent-room-name.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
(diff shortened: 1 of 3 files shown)


davidzhao (approved):

theomonnom:
I'm wondering if the name should just be room_name?

PR #2069, opened 2026-07-19
fix(elevenlabs): stop publishing Buffer pool memory as audio

### LINKED ISSUE #2068: ElevenLabs TTS publishes Buffer pool memory as audio, producing bursts of full-scale noise
### Describe the bug

`plugins/elevenlabs/src/tts.ts` hands `AudioByteStream.write()` the raw `ArrayBuffer` behind each decoded audio chunk instead of the `Buffer` view:

```ts
const audioData = Buffer.from(data.audio as string, 'base64'); // line 574
...
for (const frame of bstream.write(audioData.buffer)) {         // line 1101
```

and the same pattern on the non-streaming path:

```ts
for (const frame of bstream.write(value.buffer)) {             // line 897
```

Node serves `Buffer`s below `Buffer.poolSize >>> 1` out of a shared 8 KB pool, so a small chunk is a window into a larger `ArrayBuffer`, described by `byteOffset` and `byteLength`. Passing `.buffer` discards both. `AudioByteStream.write` only honours them for a value that `ArrayBuffer.isView` accepts:

```ts
const bytes = ArrayBuffer.isView(data)
  ? new Int8Array(data.buffer, data.byteOffset, data.byteLength)
  : new Int8Array(data); // an ArrayBuffer lands here, so the whole pool is read
```

So any audio chunk small enough to be pooled is published as the entire 8,192 byte pool: the real audio plus up to 7 KB of unrelated adjacent memory, interpreted as PCM. At the plugin's default `pcm_22050` that is 185.8 ms of audio in place of the chunk.

What it sounds like on a call is a burst of loud broadband noise spliced into the middle of a sentence. It clips at full scale, so it is substantially louder than the speech around it.

`autoMode: true` makes this materially more likely, since letting the provider choose generation triggers produces small chunks routinely, particularly at generation boundaries. In our production telephony traffic it landed most often on short scripted lines.

Two notes on scope:

- `agents/src/inference/tts.ts` looks similar but is **not** affected, because it copies first: `new Int8Array(Buffer.from(serverEvent.audio, 'base64'))` produces an exact size buffer, so its `.buffer` is just the chunk. That contrast is probably why this went unnoticed, and it matches what we measured: across 175 production recordings, every occurrence was on the ElevenLabs plugin and none were on Cartesia via Inference.
- The other `bstream.write(x.buffer)` call sites found by search operate on `AudioFrame.data`, which is a different situation and not something we have investigated.

### Relevant log output

Nothing is logged. The corruption is silent: no error, no warning, and the frames are structurally valid audio.

### Describe your environment

- macOS 26.3.1, Node v22.21.0
- `@livekit/agents` 1.5.2
- `@livekit/agents-plugin-elevenlabs` 1.5.2, also confirmed present in 1.5.3
- Model `eleven_flash_v2_5`, default `pcm_22050` encoding, `autoMode: true`

Runs against the real `AudioByteStream`, no network or API key needed:

```js
import { AudioByteStream, initializeLogger } from '@livekit/agents';

initializeLogger({ pretty: false });

const sampleRate = 22050;
const samples = 500; // 1000 bytes, comfortably inside Node's pool threshold
const pcm = Buffer.alloc(samples * 2);
for (let i = 0; i < samples; i++) {
  pcm.writeInt16LE(Math.round(Math.sin((2 * Math.PI * 440 * i) / sampleRate) * 3000), i * 2);
}

// exactly what the plugin does with an incoming websocket audio message
const audioData = Buffer.from(pcm.toString('base64'), 'base64');
console.log('pooled:', audioData.buffer.byteLength !== audioData.length); // true

const drain = (written) => {
  const stream = new AudioByteStream(sampleRate, 1);
  const frames = [...stream.write(written), ...stream.flush()];
  return {
    samples: frames.reduce((n, f) => n + f.data.length, 0),
    peak: Math.max(...frames.flatMap((f) => [...f.data].map(Math.abs))),
  };
};

console.log('current :', drain(audioData.buffer)); // { samples: 4096, peak: ~32767 }
console.log('fixed   :', drain(audioData));        // { samples: 500,  peak: 3000  }
```

Output:

```
pooled: true
current : { samples: 4096, peak: 32731 }
fixed   : { samples: 500, peak: 3000 }
```

185.8 ms emitted for a 22.7 ms chunk, peaking at full scale int16, roughly 17 dB above the tone it was supposed to be carrying.

### Additional information

The fix is to pass the view and let `AudioByteStream` apply the offset and length it already handles:

```diff
-        for (const frame of bstream.write(value.buffer)) {
+        for (const frame of bstream.write(value)) {
-          for (const frame of bstream.write(audioData.buffer)) {
+          for (const frame of bstream.write(audioData)) {
```

Happy to open a PR for this.

For anyone hitting the same thing before it is released, the artifact is identifiable in recordings by its spectrum rather than its level: it is near flat from DC up to the stream's Nyquist edge, which is 11.025 kHz at `pcm_22050`, and it clips. Searching for loud events alone is misleading, because the loudest moments in most call recordings are answering machine beeps and ringback tones.

sunnyysetia:
Fixes #2068

### Problem

`plugins/elevenlabs/src/tts.ts` hands `AudioByteStream.write()` the raw `ArrayBuffer` behind each decoded audio chunk instead of the `Buffer` view, on both the streaming and non streaming paths:

```ts
const audioData = Buffer.from(data.audio as string, 'base64');
...
for (const frame of bstream.write(audioData.buffer)) {
```

Node serves `Buffer`s below `Buffer.poolSize >>> 1` out of a shared 8 KB pool, so a small chunk is a window into a larger `ArrayBuffer`, described by `byteOffset` and `byteLength`. Passing `.buffer` discards both, and `write` honours them only for a value `ArrayBuffer.isView` accepts:

```ts
const bytes = ArrayBuffer.isView(data)
  ? new Int8Array(data.buffer, data.byteOffset, data.byteLength)
  : new Int8Array(data); // an ArrayBuffer lands here, so the whole pool is read
```

So any pooled chunk is published as the entire 8,192 byte pool: the real audio plus up to 7 KB of unrelated adjacent memory, interpreted as PCM. At the plugin's default `pcm_22050` that is 185.8 ms of audio emitted in place of the chunk.

On a live call it is a burst of loud broadband noise spliced into the middle of a sentence. It clips at full scale, so it is much louder than the speech around it, and nothing is logged.

`autoMode: true` makes it materially more likely, since letting the provider choose generation triggers produces small chunks routinely, especially at generation boundaries.

We hit this in production telephony. Across 175 call recordings every occurrence was on this plugin, and none were on Cartesia through `agents/src/inference/tts.ts`, which is not affected because it copies first: `new Int8Array(Buffer.from(...))` produces an exact size buffer, so its `.buffer` is just the chunk.

### Fix

Pass the view and let `write` apply the offset and length it already handles. Two call sites, no signature change: `write` already accepts `ArrayBufferLike | ArrayBufferView`.

### Verification

Against the real `AudioByteStream`, a 22.7 ms chunk:

| | samples emitted | duration | peak |
| --- | --- | --- | --- |
| before | 4096 | 185.8 ms | 32731 (full scale int16) |
| after | 500 | 22.7 ms | 3000 (the tone that was encoded) |

Repro script is in #2068 and needs no API key.

Also run:

- `pnpm -w format:write` and `pnpm -w lint:fix`, both clean. Note these two commands also reformat `agents/src/voice/audio_recognition.ts` and `agents/src/voice/tool_executor.ts` on a clean checkout of `main`; I reverted those so this diff stays focused.
- `@livekit/agents` and `@livekit/agents-plugin-elevenlabs` build, both pass.
- `plugins/elevenlabs/src/tts.test.ts`, 7 passed and 1 skipped.
- `api:check` for the plugin fails, but it fails identically on a clean checkout of `main`, so it looks unrelated to this change.

Added a patch changeset. Happy to drop it if maintainers would rather add release notes themselves, per CONTRIBUTING.

### DIFF
--- a/.changeset/elevenlabs-buffer-pool-audio.md
+++ b/.changeset/elevenlabs-buffer-pool-audio.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-elevenlabs': patch
+---
+
+Pass the decoded audio chunk to `AudioByteStream` as a view rather than as its underlying `ArrayBuffer`. Node serves small `Buffer`s out of a shared pool, so the previous code discarded the chunk's `byteOffset` and `byteLength` and published the whole 8 KB pool as PCM, producing bursts of full-scale noise in the middle of synthesized speech.
--- a/plugins/elevenlabs/src/tts.ts
+++ b/plugins/elevenlabs/src/tts.ts
@@ -894,7 +894,7 @@ export class ChunkedStream extends tts.ChunkedStream {
         const { done, value } = await reader.read();
         if (done) break;
 
-        for (const frame of bstream.write(value.buffer)) {
+        for (const frame of bstream.write(value)) {
           if (lastFrame) {
             this.queue.put({ requestId, segmentId: requestId, frame: lastFrame, final: false });
           }
@@ -1098,7 +1098,7 @@ export class SynthesizeStream extends tts.SynthesizeStream {
         // Process audio queue
         while (this.#audioQueue.length > 0) {
           const audioData = this.#audioQueue.shift()!;
-          for (const frame of bstream.write(audioData.buffer)) {
+          for (const frame of bstream.write(audioData)) {
             sendLastFrame(false);
             lastFrame = frame;
           }


chenghao-mou (approved):
lgtm. Thanks for the PR! Could you sign the CLA before we can merge this?

sunnyysetia:
> lgtm. Thanks for the PR! Could you sign the CLA before we can merge this?

All done :)

PR #2085, opened 2026-07-22
fix(cartesia): reuse one TTS websocket across generations

sunnyysetia:
### Problem

`plugins/cartesia/src/tts.ts` opens a new Cartesia WebSocket for every streaming synthesis and force-terminates it when the generation ends. `SynthesizeStream.run()` connects per call:

```ts
ws = await connectCartesiaWebSocket({ url, headers, timeoutMs: this.connOptions.timeoutMs, abortSignal: this.abortSignal });
await Promise.all([inputTask(), sentenceStreamTask(ws), recvTask(ws)]);
```

closes it on the normal `done` path:

```ts
if (segmentId === requestId) {
  closing = true;
  clearTTSChunkTimeout();
  ws.close();
  break;
}
```

and tears it down again in `finally`:

```ts
} finally {
  if (ws && ws.readyState !== WebSocket.CLOSED) {
    safeTerminateWebSocket(ws);
  }
}
```

The `TTS` instance holds no connection state, so every turn repays the TCP, TLS, and WebSocket handshake. Cartesia recommends the opposite: keep one socket open and run many generations over it, because a fresh connection costs tens to low hundreds of milliseconds per turn (https://docs.cartesia.ai/use-the-api/compare-tts-endpoints). The Python plugin already follows that advice with a pooled connection (`livekit-plugins-cartesia`, `utils.ConnectionPool`, `max_session_duration=300`), and the JS `recvTask` even carried a `// IMPORTANT: Remove listeners so connection can be reused` comment that was never wired to a pool.

Everything Cartesia needs per generation (model, voice, encoding, sample rate, speed, emotion, volume, language) already travels in-band on each message, and a fresh `context_id` is generated per `run()`, so the socket is safe to reuse as-is. Only the auth and version headers are set at handshake time.

### Fix

Hold a `ConnectionPool<WebSocket>` on the `TTS` instance, the same pattern `plugins/fishaudio`, `plugins/inworld`, and `plugins/xai` already use, and run each generation inside `pool.withConnection`:

```ts
await this.#pool.withConnection(
  async (ws) => {
    if (ws.readyState !== WebSocket.OPEN) {
      throw new APIConnectionError({ message: 'Cartesia pooled websocket is not open' });
    }
    await Promise.all([inputTask(), sentenceStreamTask(ws), recvTask(ws)]);
  },
  { timeout: this.connOptions.timeoutMs, signal: this.abortSignal },
);
```

- A generation no longer closes the socket on `done`; it just breaks the receive loop, so `withConnection` returns the socket to the pool and the next turn skips the handshake.
- The chunk-timeout watchdog now poisons its socket and throws a retryable `APITimeoutError`, so `withConnection` removes that socket from the pool instead of returning a stuck one.
- `updateOptions` invalidates the pool only when a handshake input actually changes (`apiKey`, `apiVersion`, `baseUrl`). Model, voice, and the generation controls are sent per message, so a pooled socket serves the new values without reconnecting.
- Adds `TTS.prewarm()` to open the socket before the first turn and `TTS.close()` to drain the pool. `maxSessionDuration` is 300000 ms with `markRefreshedOnGet`, matching the Python plugin.

This mirrors the Python plugin's behaviour and keeps the existing IPv4 happy-eyeballs connect retry, chunk-timeout, and error-propagation logic intact.

### Verification

Added a local-WebSocketServer test suite in `plugins/cartesia/src/tts.test.ts` that needs no API key (mirrors `plugins/fishaudio/src/tts.test.ts`):

- reuses one websocket across sequential turns (`connectionCount === 1`)
- prewarms and reuses the ready websocket
- discards a poisoned websocket after a failure, then reconnects
- replaces a websocket that closed while idle
- closes the pooled websocket when the TTS closes

`vitest run plugins/cartesia/src/tts.test.ts`: 5 passed, 1 skipped (the key-gated integration test). Build, `typecheck`, `lint`, and `prettier` are clean. `api:check` for this plugin fails, but it fails identically on a clean checkout of `main` (there is no checked-in `etc/*.api.md` report for the plugin), so it is unrelated to this change. Added a `minor` changeset.

I also measured the actual patched plugin against direct Cartesia Sonic from a hosted us-east worker, 11 sequential turns, first-audio latency:

| Path | first-audio p50 | first-audio p95 |
| --- | ---: | ---: |
| Reconnect per turn (current behaviour) | 364 ms | 757 ms |
| Pooled, warm (this change) | 262 ms | 300 ms |

Pooling removes about 102 ms at p50 and collapses the tail, because the reconnect p95 is a handshake landing on a turn while the pooled path has already paid it once. A prewarmed first turn measured 274 ms, within the warm band.

### DIFF
--- a/.changeset/cartesia-persistent-tts-websocket.md
+++ b/.changeset/cartesia-persistent-tts-websocket.md
@@ -1,5 +1,5 @@
 ---
-'@livekit/agents-plugin-cartesia': minor
+'@livekit/agents-plugin-cartesia': patch
 ---
 
 Reuse a single Cartesia TTS WebSocket across generations instead of opening and closing one per synthesis. The plugin now holds a `ConnectionPool` on the `TTS` instance (matching the Python plugin and the fishaudio/inworld/xai plugins), so only the first turn pays the connect and later turns skip the TCP/TLS and WebSocket handshake. Adds `TTS.prewarm()` to open the socket before the first turn and a `TTS.close()` that drains the pool.
--- a/.changeset/cartesia-persistent-tts-websocket.md
+++ b/.changeset/cartesia-persistent-tts-websocket.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-cartesia': minor
+---
+
+Reuse a single Cartesia TTS WebSocket across generations instead of opening and closing one per synthesis. The plugin now holds a `ConnectionPool` on the `TTS` instance (matching the Python plugin and the fishaudio/inworld/xai plugins), so only the first turn pays the connect and later turns skip the TCP/TLS and WebSocket handshake. Adds `TTS.prewarm()` to open the socket before the first turn and a `TTS.close()` that drains the pool.
--- a/plugins/cartesia/src/tts.test.ts
+++ b/plugins/cartesia/src/tts.test.ts
@@ -188,7 +188,16 @@ describe('Cartesia streaming pool', () => {
     try {
       expect(await synthesizeTurn(cartesia, 'first turn.')).not.toHaveLength(0);
       await waitFor(firstClosed);
-      expect(await waitFor(synthesizeTurn(cartesia, 'second turn.'))).not.toHaveLength(0);
+      // Let the client observe the close so the idle handler removes the socket
+      // before the next checkout, making the maxRetry: 0 assertion deterministic.
+      await new Promise((resolve) => setTimeout(resolve, 100));
+      // maxRetry: 0 proves the idle-closed socket was dropped from the pool, not
+      // handed back to burn the turn's only attempt.
+      expect(
+        await waitFor(
+          synthesizeTurn(cartesia, 'second turn.', { ...DEFAULT_API_CONNECT_OPTIONS, maxRetry: 0 }),
+        ),
+      ).not.toHaveLength(0);
       expect(server.connectionCount()).toBe(2);
     } finally {
       await cartesia.close();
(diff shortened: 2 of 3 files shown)


sunnyysetia:
Thanks, both are fair. Addressed in 00780c4:

**Idle-closed sockets burning a retry.** When a pooled socket is created it now gets an idle `close` handler that removes it from the pool, plus a no-op `error` handler so an idle error does not crash the process. A socket that closes between turns is dropped from `available` instead of being handed to the next turn, so the `readyState` guard no longer spends a retry to discard it and a `maxRetry: 0` turn recovers. The idle-close test now runs the recovery turn at `maxRetry: 0` to lock this in.

**Handshake options read live during connect.** `#connectWebSocket` now snapshots `apiKey`, `apiVersion`, and `baseUrl` before connecting and, if one changed while the connect was in flight, closes the socket and reconnects on the new value, so a pooled socket is never built on stale credentials. This mirrors the fishaudio plugin's post-connect model re-check, and it composes with the existing `updateOptions` invalidate so a handshake change never leaves a stale-credential socket in the pool.

sunnyysetia:
Addressed the newer feedback in the latest commits (e590394):

**Mid-generation socket drop (Devin).** Good catch. `recvTask` now tracks whether the generation completed, and `onClose`/`onError` capture an early drop as a pending error that is thrown after the receive loop. So a socket that dies before `done` fails the turn (the framework retries) instead of ending it silently mid-speech, and withConnection discards the dead socket rather than pooling it. Added a test that emits one chunk, drops the socket without `done`, and asserts fail-over plus a fresh reconnect.

**Idle-closed socket / readyState guard (Codex, re-flag).** The common idle-close path no longer reaches that guard: a pooled socket now carries a `close` handler that removes it from the pool the moment it closes while idle, so the next checkout opens a fresh one and does not spend a retry. The `readyState` guard remains only as a defense-in-depth backstop for the narrow window where a socket starts closing between checkout and use. Combined with the mid-generation fix above, a dead socket is never returned as a normal result.

toubatbrian (approved):

PR #2100, opened 2026-07-23
feat(telemetry): tag content-bearing trace and log keys with lk.pii. for redaction AGT-3074

chenghao-mou:
**Why:** Cloud redaction only drops sensitive telemetry when it is tagged; the JS SDK still exposed content through legacy keys and log bodies.

**Behavior:** Ports livekit/agents#6356: trace and structured-log content now uses `lk.pii.*`, exception recording honors job redaction, and guard tests classify future keys. Dashboards must migrate from legacy keys.

**Validation:** Format, lint, typecheck, build, and relevant tests pass. Full-suite credential/network failures and the API Extractor TypeScript parser mismatch remain.

Fixes AGT-3074


toubatbrian (approved):
LG!

davidzhao (approved):

PR #2105, opened 2026-07-24
update to otel core 2.8

davidzhao:
updating dependencies and their usage

## OpenTelemetry 2.x migration

- Requires OpenTelemetry JS SDK 2.x and experimental packages 0.2xx.
- Configure span processors through the provider's `spanProcessors` constructor option.
- Custom providers must include `telemetry.FanoutSpanProcessor` and pass `registerSpanProcessor` to `telemetry.setTracerProvider`.
- Provider shutdown now flushes pending spans and releases exporter resources.
- The changeset is a minor release because SDK 1.x compatibility is intentionally removed.

## Verification

Verified at `521730e4df1ea6ecfc5015ae9340000f4010e988`:

- 27 focused telemetry and voice-telemetry tests passed.
- Full workspace typecheck and build passed.
- Frozen lockfile install and formatting passed.
- Changed-file lint passed with no new errors.
- Built-package runtime smoke verified OTel 2 log metadata and dual trace export.
- Independent spec-compliance and code-quality reviews approved.

Cue is not applicable because this change affects OTLP/log export rather than the voice event path; the direct OTel runtime boundary was verified instead.


u9g (commented):
Overrides look correct, just a note about the one override, but it's not breaking anything.

pnpm-workspace.yaml
```diff
@@ -19,12 +19,31 @@ minimumReleaseAgeExclude:
...
   '@changesets/assemble-release-plan': 'patches/@changesets__assemble-release-plan.patch'
 
+# Force patched versions of transitive dependencies flagged by Dependabot. Each override is
+# scoped to the vulnerable range, so it becomes inert once the parent dependency updates.
+overrides:
+  'adm-zip@<0.6.0': ^0.6.0
+  'uuid@<11.1.1': ^11.1.1
+  'lodash@<4.18.0': ^4.18.0
+  'protobufjs@<7.6.5': ^7.6.5
+  'ws@>=8.0.0 <8.21.0': ^8.21.0
```
u9g, line 35: `8.20.1` sits inside this override's range and gets bumped to `≥8.21.0`, so the `ws@8.20.1` entry in `minimumReleaseAgeExclude` (line 12) is now dead config — safe to delete along with its comment.

u9g (approved):

PR #2113, opened 2026-07-24
fix: settle recorder playout waits when the downstream output drops a frame

chenghao-mou:
**Bug:** the recorder output wrapper registers its playback segment *after* forwarding the frame downstream, so a `playback_finished` landing during the opening capture was discarded by the base class and `waitForPlayout()` never settled. A frame the downstream output drops (`ParticipantAudioOutput` when interrupted-while-paused) left the same dangling segment, wedging the interrupted turn and `RecorderIO.close()`.

**Fix:** buffer finishes that arrive mid-capture and replay them once the segment is registered, and settle segments the downstream output never counted with a synthetic interrupted finish once it drains, revoked if the output starts accepting frames mid-segment.

Fixes AGT-3177

### DIFF
--- a/.changeset/steady-recorders-finish.md
+++ b/.changeset/steady-recorders-finish.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Prevent recorder output wrappers from deadlocking playout waits when interruption finishes during first-frame capture.
--- a/agents/src/voice/recorder_io/recorder_io.test.ts
+++ b/agents/src/voice/recorder_io/recorder_io.test.ts
@@ -5,12 +5,12 @@ import { AudioFrame } from '@livekit/rtc-node';
 import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
-import { beforeAll, describe, expect, it } from 'vitest';
+import { beforeAll, describe, expect, it, vi } from 'vitest';
 import { initializeLogger } from '../../log.js';
 import { type StreamChannel, createStreamChannel } from '../../stream/stream_channel.js';
 import { Future, isWritableStreamClosedError } from '../../utils.js';
 import type { AgentSession } from '../agent_session.js';
-import { AudioInput, AudioOutput } from '../io.js';
+import { AudioInput, AudioOutput, type PlaybackFinishedEvent } from '../io.js';
 import { RecorderIO } from './recorder_io.js';
 
 class FakeAudioInput extends AudioInput {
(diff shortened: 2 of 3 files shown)


toubatbrian:
Confirmed this fixes the reported deadlock — I reproduced the hang 5/5 and verified your patch resolves it.

The sequence I captured: the interrupted `playback_finished` lands inside `RecorderAudioOutput.captureFrame`'s `await this.nextInChain.captureFrame(frame)`, before `super.captureFrame` has registered the segment. `AudioOutput.onPlaybackFinished` discards it as surplus, and the segment registered a moment later has nothing left to settle it. Your `captureDeferredPlaybackFinished` window covers exactly that case, so it holds.

One framing note: the frame is *released and accepted* downstream in every run I captured, never dropped — `openedRecorderSegment && !downstreamAccepted` was zero across all runs, including one with 356 gate-dropped frames. So the dropped-frame half of this PR is defense-in-depth rather than the reported cause.

Where I ended up differently is scope rather than correctness. Because playout is still tracked by a single global counter and recorded audio by a single shared `accFrames` buffer, the wrapper can't attribute a finish to a specific segment, and a few adjacent interleavings stay broken. These were verified by execution against this branch, not by inspection:

- **Still deadlocks**: when the sink counts a segment (`room_io/_output.ts:441`) and the capture then rejects (`:450`, reachable when the track or room is gone), `waitForPlayout` parks on the downstream wait forever.
- **Silently loses recorded audio**: `accFrames = []` discards a newer segment's buffered frames when an older overlapping segment's finish is processed.
- **Misreports playout state**: `waitForPlayout` returns the base class's `lastPlaybackEvent`, so a caller can be told `interrupted: true` for a segment that actually played to completion.

I opened #2114 with a per-segment version that closes these — it passes 15 of the 16 tests here. Your timeline is worth keeping as a regression test either way, and I'm happy to port it across.

Also worth flagging for both PRs: this is only reachable when `TranscriptSynchronizer` is out of the output chain, since its audio output already does per-frame `segmentAccepted` tracking plus `settleDriftFinish()` reconciliation. With `useTtsAlignedTranscript` defaulting to true, the default chain absorbs the drift — so the real exposure is `useTtsAlignedTranscript: false`, realtime models, and avatar or custom sinks.

toubatbrian:
Following up on the 15/16, since I should name the one that doesn't pass rather than leave it hanging: `keeps an owed completion reported during the capture of a dropped frame`.

It sets up its scenario by calling `downstream.flush()` directly, which leaves the sink owing a finish for a segment the recorder never registered. My version doesn't track downstream segment indices, so it has nothing to attribute that finish to.

I don't think that state is reachable in-tree, though. Segment boundaries only come from `flush()`, and both `SyncedAudioOutput.flush` (`transcription/synchronizer.ts:807`) and `ParticipantAudioOutput.flush` (`room_io/_output.ts:496`) are only ever reached by propagation from the recorder's own `flush()` — so a sink can end up with fewer segments than the recorder, never more.

I did check the reachable form of the same interleaving (an owed real finish for an older recorder segment arriving during a newer segment's dropped capture, with no independent downstream flush) and that one passes. Happy to be wrong if you know a path that gets the sink ahead of the recorder.

chenghao-mou:
closing in favor of https://github.com/livekit/agents-js/pull/2114

PR #2121, opened 2026-07-26
fix(interruption): surface a duplicated @livekit/rtc-node instead of dropping all audio

toubatbrian:
## The failure

A user's adaptive interruption returned `backchannel` for every overlap, forever, with no error and no failover. Instrumented counters from their live session:

```
broadcastFrames      2945   room audio frames reaching the source
forwardedAudio          0   frames the forward loop classified as audio
forwardedSentinels   2960   = 2945 audio frames + 15 real sentinels
```

Two copies of `@livekit/rtc-node` were loaded in one process, so the `AudioFrame` constructor that built the frames was not the one `@livekit/agents` compared against. Every `frame instanceof AudioFrame` went false:

1. `InterruptionStreamBase.pushFrame` took the `!(frame instanceof AudioFrame)` branch and wrote the raw frame into `inputStream` as if it were a sentinel, skipping resampling.
2. The transform's `if (chunk instanceof AudioFrame)` also failed, so it fell through to the `chunk.type === ...` comparisons. A real `AudioFrame` has no `type`, so every branch missed and the frame was **silently discarded** — no accept, no drop, no counter, no log.
3. Adaptive interruption had nothing to classify and fell back to a backchannel verdict on every overlap.

## What this changes

**Audio is recognised by shape, not by constructor identity.** A frame from another copy of the module is rebuilt as a local `AudioFrame` over the same sample buffer (no copy) and keeps flowing. That is not cosmetic: `AudioResampler.push()` calls `frame.protoInfo()`, which resolves pointers through the frame's *own* `FfiClient` singleton, so passing a foreign frame to a local resampler would mix two FFI runtimes.

**The first such frame raises a once-per-process `error`** naming the duplicate dependency and how to find and collapse it, so the user is told the cause instead of inferring it from a classifier that only ever says "backchannel". Accepting the frame rather than throwing is deliberate — hard-failing a live voice session over a dependency-resolution problem trades a silent degradation for a dropped call, and the diagnostic is loud enough that this cannot pass unnoticed.

**The sentinel dispatch is now an exhaustive `switch`.** Its `default` is both a compile-time `never` check (so a new sentinel cannot be added without handling it) and a real runtime guard: anything the stream cannot classify is reported at `error`, rate limited to the first occurrence plus one per 10s, carrying a running count of discarded chunks.

**Two other cross-package `instanceof` checks that fail the same way are now structural.** `ParticipantAudioInputStream.setParticipant` discriminates on `typeof participant === 'string'` (matching the existing `BaseParticipantTranscriptionOutput.setParticipant`) instead of `instanceof RemoteParticipant`, which would have stored an object as the identity and subscribed to nothing; `waitForParticipantAttribute` matches on identity alone, which is already unique within a room, instead of hanging forever.

## Audit of `instanceof` against `@livekit/rtc-node` types in `agents/src`

| Site | Failure mode on a false negative | Action |
| --- | --- | --- |
| `interruption/interruption_stream.ts` transform | **Silent** — frame matched no branch and vanished | Fixed |
| `interruption/interruption_stream.ts` `pushFrame` | **Silent** — frame written through as a sentinel, resampling skipped | Fixed |
| `voice/room_io/_input.ts` ×2 (`RemoteParticipant`) | **Silent** — object stored as identity, no track subscribed, no audio input | Fixed |
| `utils.ts` `waitForParticipantAttribute` (`RemoteParticipant`) | **Silent** — future never resolves, caller hangs | Fixed |
| `llm/utils.ts` (`VideoFrame`) | **Loud** — falls to `throw new Error('Unsupported image type')` | Left alone |

`room_io.ts` (`RealtimeModel`, `ParticipantTranscriptionOutput`) and the `LLM` / `STT` / `TTS` / `VAD` / `ChatMessage` checks are against `@livekit/agents`' own classes, not `@livekit/rtc-node`, and `worker.ts` / `eot/transports.ts` check JS builtins.

## Tests

`agents/src/inference/interruption/interruption_dual_copy.test.ts` drives the real pipeline (room audio into the interruption stream channel, through the audio transformer, out to a mocked WS transport) with frames from a locally-declared class of identical shape, so `instanceof AudioFrame` is false exactly as it is under a duplicated dependency.

Verified red before green. Against the pre-fix transform:

```
 ❯ src/inference/interruption/interruption_dual_copy.test.ts (3 tests | 2 failed)
   × keeps user audio flowing and says why, instead of discarding every frame
   × reports a chunk it cannot classify instead of dropping it in silence

AssertionError: expected 0 to be greater than 0
 ❯ src/inference/interruption/interruption_dual_copy.test.ts:256:35
    256|     expect(totalAudioSendCount()).toBeGreaterThan(0);
```

Zero frames reached the gateway — the reported symptom reproduced exactly. After the fix all 8 tests across the two new files pass, and the full `agents` suite is green at 1491 passed / 5 skipped across 112 files.

## Test plan

- [x] `pnpm build` exits 0
- [x] `pnpm vitest run` in `agents/` — 1491 passed, 5 skipped, 0 failed
- [x] `pnpm lint` exits 0 with no new warnings
- [x] `pnpm format:check` clean
- [ ] `pnpm api:check` — fails identically on `main`, not evaluated

### DIFF
--- a/.changeset/interruption-duplicate-rtc-node.md
+++ b/.changeset/interruption-duplicate-rtc-node.md
@@ -1,0 +1,29 @@
+---
+'@livekit/agents': patch
+---
+
+fix(interruption): stop silently dropping all audio when `@livekit/rtc-node` is loaded twice
+
+When an application and `@livekit/agents` resolve two different installs of `@livekit/rtc-node` —
+easily done by linking a local checkout, or by a transitive version split — room audio is built by
+one `AudioFrame` constructor and type-checked against the other. Every `frame instanceof AudioFrame`
+went false, so audio was pushed into the interruption stream as if it were a control sentinel,
+matched none of the sentinel branches in the transform, and disappeared: no accept, no drop, no
+counter, no log. Adaptive interruption had nothing left to classify and answered **every** overlap
+with the default backchannel verdict, so the agent never yielded to the user — with no error, no
+warning and no failover to give the cause away.
+
+- Audio is now recognised by shape rather than by constructor identity. A frame from another copy
+  of the module is rebuilt as a local `AudioFrame` (over the same sample buffer, so no copy) and
+  keeps flowing, which also stops a foreign frame being handed to a resampler backed by a different
+  FFI runtime.
+- The first such frame raises a once-per-process `error` naming the duplicate dependency and how to
+  find and collapse it, instead of leaving the failure to be inferred from a classifier that only
+  ever says "backchannel".
+- The transform's sentinel dispatch is now an exhaustive `switch`, and anything it cannot classify
+  is reported at `error` (rate limited, with a running count of discarded chunks) rather than
+  falling off the end of the chain.
+
+Two other cross-package `instanceof` checks that would fail the same way are now structural:
+`RoomIO`'s participant audio input no longer mistakes a `RemoteParticipant` for an identity string
+and subscribes to nothing, and `waitForParticipantAttribute` no longer waits forever.
--- a/agents/src/audio_frame_guard.test.ts
+++ b/agents/src/audio_frame_guard.test.ts
@@ -1,0 +1,111 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { AudioFrame } from '@livekit/rtc-node';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { asAudioFrame, describeUnknownChunk, isAudioFrameShaped } from './audio_frame_guard.js';
+import { initializeLogger, log } from './log.js';
+
+initializeLogger({ pretty: false, level: 'silent' });
+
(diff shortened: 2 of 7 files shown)


toubatbrian:
Closing note, for the record: the diagnosis behind this PR was right, but the remedy was aimed at the wrong layer.

The duplicated `@livekit/rtc-node` was real — it came from the Krisp plugin reaching the Cloud backend through `createRequire`, which resolves the internal package's `require` condition and pulls in the CJS build alongside the framework's ESM one. What this PR proposed was to detect the duplication at the consumption sites and rebuild the foreign frame. Fixing it at the source is better on every axis, and that is what shipped:

- #2130 (consumer-side guard in the krisp plugin — still needed while the two below are unreleased and `rtc-node` is tracked by range rather than pin)
- livekit/plugin-krisp-viva-internal#13 (producer side: construct with the caller's constructor) — merged
- livekit/node-sdks#699 (root: `Symbol.hasInstance` branding on the data classes, so `instanceof` holds across copies) — merged

Issue #1124, opened 2026-03-12
Speech handle stuck in limbo after tool call interruption — agent enters unrecoverable silence loop


enriqueespaillat-gyde:
### Describe the bug

When a speech handle is interrupted during the `waitForScheduled()` / `waitForAuthorization()` gate after a tool call, the speech handle appears to never be finalized. It seems to remain the "active" speech indefinitely, blocking all subsequent speech handles from being authorized. The agent enters a silence loop: LLM generates responses, TTS synthesizes audio, but nothing is ever forwarded to the client. The agent does not appear to recover — the loop continues until the user disconnects.

This seems related to #1085 but is a different manifestation — rather than dropped transcripts, the issue here appears to be **state corruption** of the speech handle lifecycle. We believe the interrupted speech handle is never moved to a terminal state, so the pipeline cannot advance.

### Suspected root cause

In `agent_activity.ts`, `mainTask` calls `_authorizeGeneration()` and then `await _waitForGeneration()`. If the speech handle is interrupted after authorization but before the reply task calls `_markGenerationDone()`, `_waitForGeneration()` blocks indefinitely because nothing resolves its underlying `Future`. `mainTask` never moves on to the next queued handle.

1. Tool call completes → pipeline re-enters → LLM/TTS generate a response
2. Audio forwarding is gated behind `waitForScheduled()` / `waitForAuthorization()`
3. User speaks during the gate → speech handle is interrupted
4. `mainTask` is stuck on `await _waitForGeneration()` for the interrupted handle
5. Subsequent speech handles queue up but are never popped or authorized
6. Every new interrupt targets the same zombie handle (`_currentSpeech` never advances)

This looks like the same root cause as #836 (`_authorizeGeneration` doesn't handle interrupted speech handles), but manifesting as an unrecoverable loop rather than a one-time hang.

### Evidence from production logs

Timeline from a real call where a tool call completed successfully, but the agent went silent and did not recover:

```
T+0.0s   Tool call executes successfully
         Pipeline re-enters → LLM generates response → TTS synthesizes audio
         Audio forwarding gated behind waitForScheduled/waitForAuthorization

T+9.0s   "speech interrupted by audio activity" (speech_AAA)
         User heard silence for 9s and spoke
T+9.5s   "speech interrupted, new user turn detected" (speech_AAA)

T+15s    User: "Did you get stuck?" (confidence 0.999)
         "starting preemptive generation"
[...]

PR #2122, opened 2026-07-26
fix(voice): stop a confirmed interruption from resuming the interrupted speech

toubatbrian:
## Problem

When the adaptive model rules an overlap a genuine barge-in, the agent pauses — and then puts the
interrupted speech back on the wire, continuing from exactly where it stopped. Reported from a live
session: "when I interrupt after asking for a long story, it does pause, but then it continues the
audio from where it's been cut off."

Two independent defects produce that.

**1. The sink gate is opened before the interruption reaches the sink.**

`ParticipantAudioOutput.captureFrame` parks frames at a pause gate and races them against
`interruptedFuture`, which `clearBuffer()` resolves:

```ts
if (!this.playbackEnabledFuture.done) {
  this.audioSource.clearQueue();
  await Promise.race([this.playbackEnabledFuture.await, this.interruptedFuture.await]);
  if (this.interruptedFuture.done) {
    return;
  }
}
```

`cancelSpeechPause` interrupts the paused speech and then calls `audio.resume()` without calling
`clearBuffer()`. That resume is only meant to admit the *next* speech, but it also releases every
frame parked during the pause. Whether you hear the interrupted speech continue depends on which of
`clearBuffer()` (async, via the playout task) and `resume()` (sync, at the end of
`cancelSpeechPause`) wins — so it reproduces intermittently and needs no false-interruption timer,
which is why affected sessions show no `agent_false_interruption` event.

**2. A confirmed interruption was left resumable.**

`onInterruption` only parked the speech through `interruptByAudioActivity`, leaving the
false-interruption timer free to resume it after its timeout. That timer exists for overlaps nobody
has ruled on yet, so a confirmed verdict now ends the pause outright.

## Runtime evidence

Instrumented barge-in run, consecutive events:

| event | key fields |
| --- | --- |
| `_output.ts:pause` | gate closed, `interruptedAlready: false`, 1 pending segment |
| `cancelSpeechPause:ungateSink` | still 1 pending segment |
| `_output.ts:resume` | **`clearBufferAlreadyFired: false`** |
| `captureFrame` past the gate | **a frame parked during the pause reached the wire** |

Over that run 14 frames were parked: 13 dropped (clearBuffer won) and 1 released (resume won).

After the fix, across three runs every resume that actually opened a closed gate saw
`clearBufferAlreadyFired: true`, with zero violations and zero released frames.

## Changes

- `cancelSpeechPause` calls `clearBuffer()` before `resume()` when it interrupted the paused speech.
- `onInterruption` commits the interruption instead of leaving it parked. Side benefit: the agent
  stops roughly 700ms sooner, since it no longer waits for the final transcript.

## Test plan

- [x] New unit test asserts the `clearBuffer` → `resume` ordering; fails on `main`
      (`expected [ 'resume' ] to deeply equal [ 'clearBuffer', 'resume' ]`).
- [x] New unit test asserts a confirmed interruption is not resumable by the timer.
- [x] `agents/src/voice` + `agents/src/inference/interruption`: 557 passed.
- [x] cue-cli barge-in scenario, 3 iterations: agent stops and answers the new question.
- [ ] Live confirmation that the interrupted story no longer resumes. The cue-cli runs never parked
      a frame at the gate post-fix, so they exercise the ordering invariant but not the release path.

### DIFF
--- a/.changeset/commit-confirmed-interruption.md
+++ b/.changeset/commit-confirmed-interruption.md
@@ -1,0 +1,12 @@
+---
+'@livekit/agents': patch
+---
+
+Commit a confirmed interruption instead of leaving it resumable
+
+When adaptive interruption ruled an overlap a genuine barge-in, `onInterruption` only
+parked the speech through `interruptByAudioActivity`, leaving the false-interruption
+timer free to put the interrupted audio back on the wire once its timeout elapsed. That
+timer exists for overlaps nobody has ruled on yet, so a confirmed verdict now ends the
+pause outright. This also stops the agent roughly 700ms sooner, since committing the
+interruption no longer waits for the final transcript to arrive.
--- a/.changeset/commit-confirmed-interruption.md
+++ b/.changeset/commit-confirmed-interruption.md
@@ -2,11 +2,18 @@
 '@livekit/agents': patch
 ---
 
-Commit a confirmed interruption instead of leaving it resumable
+Stop a confirmed interruption from resuming the interrupted speech
 
-When adaptive interruption ruled an overlap a genuine barge-in, `onInterruption` only
-parked the speech through `interruptByAudioActivity`, leaving the false-interruption
-timer free to put the interrupted audio back on the wire once its timeout elapsed. That
-timer exists for overlaps nobody has ruled on yet, so a confirmed verdict now ends the
-pause outright. This also stops the agent roughly 700ms sooner, since committing the
-interruption no longer waits for the final transcript to arrive.
+Two defects let audio the model had already ruled a genuine barge-in get back on the wire.
+
+`cancelSpeechPause` un-gated the audio sink without first signalling the interruption to it.
+Frames of the speech it had just interrupted were parked at the sink's pause gate, so opening
+that gate — meant only to admit the next speech — released them and the interrupted speech
+continued from exactly where it stopped. The interruption is now signalled before the gate
+opens, so those frames are dropped.
+
+`onInterruption` also only parked the speech through `interruptByAudioActivity`, leaving the
+false-interruption timer free to resume it once its timeout elapsed. That timer exists for
+overlaps nobody has ruled on yet, so a confirmed verdict now ends the pause outright. This
+also stops the agent roughly 700ms sooner, since committing the interruption no longer waits
+for the final transcript to arrive.
--- a/agents/src/voice/agent_activity.test.ts
+++ b/agents/src/voice/agent_activity.test.ts
@@ -413,7 +413,12 @@ describe('AgentActivity - mainTask', () => {
     const handle = SpeechHandle.create({ allowInterruptions: true });
     handle._authorizeGeneration();
 
-    const audioOutput = { canPause: true, pause: vi.fn(), resume: vi.fn() };
+    const audioOutput = {
+      canPause: true,
+      pause: vi.fn(),
+      resume: vi.fn(),
+      clearBuffer: vi.fn(),
+    };
     const fakeActivity = {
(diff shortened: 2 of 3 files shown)


toubatbrian:
**Do not merge this without #2124.**

#2124 fixes a pre-existing bug where `ParticipantAudioOutput`'s interrupt signal is level-triggered and only reset inside `flush()`, so a stale `clearBuffer()` from a previous turn makes every frame parked at the pause gate bail. This PR widens the exposure: on the path where a reply completes its playout while paused and the user's transcript then finalizes, `main` never calls `clearBuffer()` at all, whereas this PR does.

Measured with a real `AgentSession` and a real `ParticipantAudioOutput`, on the next reply:

| | flag left set after `cancelSpeechPause` | next reply delivered |
| --- | --- | --- |
| `main` | no | 20/20 frames |
| this PR | yes | 2/20 frames, 18 lost |

That is a reply that stops dead after ~40ms of 400ms, which the session still commits to chat context as fully spoken and not interrupted.

This PR needs no changes. #2124 resolves it, and the two were verified together: all scenarios deliver 20/20 with both applied, and this PR's own behaviour is preserved — its parked frames are still dropped, identically with and without #2124. Either merge order is fine as long as both land together.

toubatbrian:
The first commit — `onInterruption` committing a confirmed interruption instead of leaving it parked — is carried into #2126.

Verified there against a real `AgentSession` and a real `ParticipantAudioOutput`, with only the model verdict synthesized:

```
origin/main: interrupted=false  deliveredAfterVerdict=35  falseInterruptionEvents=[true]
#2126:       interrupted=true   deliveredAfterVerdict=1   falseInterruptionEvents=[]
```

The second commit — the `clearBuffer()`-before-`resume()` ordering inside `cancelSpeechPause()` — is **not** carried over. Its test asserts a call sequence rather than any observable behaviour, and with #2124 landed alongside it the parked frames bail at the sink's segment-scoped gate anyway. It is not quite a no-op, though: in the reproduction above the branch delivers one frame (20ms) after the verdict rather than zero, because `resume()` lands about a millisecond ahead of the interrupted speech's `clearBuffer()`. That is worth a small standalone PR with its own evidence rather than riding along here.

#2126 also asks reviewers to confirm the product judgement this change makes: committing the verdict removes the STT final-transcript safety net, so a false positive from the interruption model now permanently ends the speech. Python's `on_interruption` behaves like the unfixed JS, so it is a framework-wide gap.

Closing in favour of #2126.

toubatbrian:
Redirect: **half** of this PR now lives in #2134 (stack 5/5, based on #2133). The other half is not being shipped.

**Kept** — `cancelSpeechPause()` clears the sink before reopening its gate. That is parity with Python, not a divergence: `_cancel_speech_pause` (`agent_activity.py:4230`) awaits `_wait_for_generation()` before `resume()`, so the interrupted reply task has already run `clear_buffer()` by the time Python reopens the gate. JS races that same await against the interrupt (deliberately, per #1124), so it returns immediately and has to signal the sink explicitly.

**Dropped** — having `onInterruption` call `cancelSpeechPause()` to commit the verdict outright. [@chenghao-mou pointed out](https://github.com/livekit/agents-js/pull/2126#discussion_r3653386785) that leaving the speech parked is by design, and Python agrees (`on_interruption`, `agent_activity.py:2066-2076`).

The cost of dropping it is measured in #2134: on the path where no STT final transcript arrives, the stack is identical to `main` — the whole tail of the interrupted reply goes back on the wire. So the 'agent pauses, then resumes mid-sentence' symptom this PR was opened for is only prevented by a transcript landing inside `falseInterruptionTimeout`. That is Python's contract, and accepting it is the explicit trade-off of merging the stack.

PR #699, opened 2025-09-15
Update realtime_model.ts types

yaniv-peretz:
seen this type error locally, and proposing a fix

PR #2123, opened 2026-07-26
fix(krisp): adopt processed frames into the local rtc-node binding

toubatbrian:
## Summary

Enabling Krisp noise cancellation made the agent impossible to interrupt: every barge-in was classified as a backchannel.

The Cloud backend is reached through `createRequire`, which resolves the internal package's `require` condition and so loads the **CJS** build of `@livekit/rtc-node` alongside the ESM one the framework uses. Frames returned by that backend were instances of the CJS copy's `AudioFrame`, so every `instanceof AudioFrame` downstream failed. `AudioRecognition` dropped them as unrecognised sentinels, and the adaptive interruption stream — fed no audio at all — fell back to `isInterruption: false` for every overlap.

Nothing logged, because a frame that matches neither `AudioFrame` nor any known sentinel simply falls off the end of the dispatch chain.

Frames are now adopted into the local binding before leaving the filter. `AudioFrame` is a plain data holder, so this shares the samples rather than copying them.

## Runtime evidence

Instrumented `interruption_stream.ts`, ran `basic_agent.ts` (which uses `krisp.voiceIsolation()`) against cue-cli, asking for a long story and then interrupting.

Before:

```
ctorName: "AudioFrame", sampleRate: 24000, isCjsAudioFrame: true, esmCtorIdentical: false
framesSeen: 0   unclassifiable: 1041   slicesEnqueued: 0   numRequests: 0
```

`ctorName` is `AudioFrame` and every field is present, yet `instanceof` is false — the frames come from the other copy.

After:

```
framesSeen: 1 → 3000 (climbing)   sampleRate: 16000   slicesEnqueued: >0   unclassifiable: 0
```

Two consecutive runs of the same scenario, each with two overlaps, produced `isInterruption: true` for all of them. Before the fix the verdict was always a backchannel with zero inference requests.

## Test plan

- [x] `pnpm vitest run plugins/krisp` — new `_frame_identity.test.ts`, 2/2 pass. It uses a same-shaped, same-named lookalike class to stand in for the second copy, asserts the adopted frame is a real `AudioFrame`, and asserts the sample buffer is shared rather than copied.
- [x] `pnpm --filter @livekit/agents-plugin-krisp build`
- [x] `pnpm exec eslint plugins/krisp/src/`
- [x] Live verification with cue-cli as described above.

## Note on follow-up

The silent-discard behaviour is what made this take so long to find: an unrecognised chunk is dropped with no counter and no log. Worth making that dispatch exhaustive and loud in a separate change, so the next mismatch of this kind surfaces immediately instead of looking like a model that always answers "backchannel".

### DIFF
--- a/.changeset/fix-krisp-frame-identity.md
+++ b/.changeset/fix-krisp-frame-identity.md
@@ -1,0 +1,13 @@
+---
+'@livekit/agents-plugin-krisp': patch
+---
+
+Fix Krisp-processed audio being invisible to the rest of the pipeline
+
+The LiveKit Cloud backend is reached through `createRequire`, which resolves the internal
+package's `require` condition and so loads the CJS build of `@livekit/rtc-node` next to the
+ESM one the framework uses. Frames returned by that backend were instances of the CJS copy's
+`AudioFrame`, so every `instanceof AudioFrame` downstream failed. Adaptive interruption saw
+zero audio and classified every barge-in as a backchannel, making it impossible to interrupt
+an agent that had noise cancellation enabled. Frames are now adopted into the local binding
+before leaving the filter, sharing their samples rather than copying them.
--- a/plugins/krisp/src/_frame_identity.test.ts
+++ b/plugins/krisp/src/_frame_identity.test.ts
@@ -1,0 +1,51 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { AudioFrame } from '@livekit/rtc-node';
+import { describe, expect, it } from 'vitest';
+import { adoptLocalAudioFrame } from './_frame_identity.js';
+
+/**
+ * Stands in for the `AudioFrame` of a second copy of `@livekit/rtc-node`: identical shape, same
+ * constructor name, unrelated identity. That is exactly what the CJS build produces when the
+ * Cloud backend pulls it in through `createRequire`.
+ */
+class ForeignAudioFrame {
+  constructor(
+    readonly data: Int16Array,
+    readonly sampleRate: number,
+    readonly channels: number,
+    readonly samplesPerChannel: number,
+    private readonly _userdata: Record<string, unknown> = {},
+  ) {}
+
+  get userdata(): Record<string, unknown> {
+    return this._userdata;
+  }
+}
+
+describe('adoptLocalAudioFrame', () => {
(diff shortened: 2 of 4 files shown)


toubatbrian:
Upstream fixes for this are now open on both sides:

- livekit/node-sdks#699 — the root fix. Brands rtc-node's pure data classes so `instanceof` survives the ESM/CJS split for every consumer. It also found a second instance of the same failure: `LocalParticipant.handleRpcMethodInvocation` does `error instanceof RpcError` on an error thrown by a user-supplied handler, and silently degrades the caller's error code to `APPLICATION_ERROR` when that check fails across copies.
- livekit/plugin-krisp-viva-internal#13 — the producer side, building the returned frame from the caller's constructor.

This PR stays useful as a guard against any dual-format dependency doing the same thing. Note that once #699 lands and rtc-node is bumped here, `adoptLocalAudioFrame` becomes a no-op on this path — foreign frames will pass `instanceof` directly — so it could be dropped at that point if the guard is judged not worth keeping.

toubatbrian:
Carried into #2126 in full.

Verified there end to end with the real second copy of `@livekit/rtc-node` — `createRequire` rooted at `plugins/krisp` resolves the package's `require` condition to `dist/index.cjs`, exactly as the Cloud backend does — pushed through the real interruption pipeline:

```
cjs constructor=AudioFrame instanceof(esm AudioFrame)=false
WITHOUT adoption -> framesSentToGateway=0  verdicts=[{isInterruption:false,numRequests:0}]  onInterruptionCalls=0
WITH adoption    -> framesSentToGateway=1  verdicts=[{isInterruption:true,numRequests:1}]   onInterruptionCalls=1
```

This is the fix for the originally reported symptom. It still earns its place even though the upstream producer-side fix (livekit/plugin-krisp-viva-internal#13) has merged: that is unreleased, and this plugin tracks `rtc-node` through a version range rather than a pin.

Closing in favour of #2126.

toubatbrian:
Redirect: this fix now lives in #2130, at the base of a five-PR stack.

It was folded into #2126 for a while; that consolidation is closed and split back apart, because bundling five independent defects made each one hard to judge. #2130 carries this change unmodified, plus the end-to-end evidence — frames built by the real second copy of `@livekit/rtc-node` reach the gateway only with the adoption in place (0 → 1 frames, backchannel → bargein).

Worth restating there and here: **of everything in that investigation, this is the defect that caused the reported symptom.** The other four are real, but none of them is what the user hit.

PR #2124, opened 2026-07-26
fix(voice): stop a finished interruption from dropping the next reply

toubatbrian:
## Symptom

The agent produces a full reply, the user hears only the first fraction of a second of it, and the rest is silence. Nothing in the session reports a problem: no error, no `agent_false_interruption`, and the chat context records the reply as **fully spoken and not interrupted**. The next turn behaves normally, so it looks like a one-off TTS glitch.

It needs the default configuration and nothing else — `resumeFalseInterruption: true`, `falseInterruptionTimeout: 2000` and a `ParticipantAudioOutput` (`canPause: true`), which is what every RoomIO agent gets.

## Mechanism

`ParticipantAudioOutput.clearBuffer()` resolves an `interruptedFuture`. `captureFrame` consults it on the paused path so a frame parked at the pause gate can bail when an interruption lands:

https://github.com/livekit/agents-js/blob/c92d739b7d1deefb412255df4f83a993405fe747/agents/src/voice/room_io/_output.ts#L425-L435

The signal is level-triggered and was reset in exactly one place — `waitForPlayoutTask`, i.e. inside `flush()`:

https://github.com/livekit/agents-js/blob/c92d739b7d1deefb412255df4f83a993405fe747/agents/src/voice/room_io/_output.ts#L453-L458

`flush()` runs at the *end* of a segment (`forwardAudio`'s `finally` in `generation.ts`), after every frame of that segment has already been captured. So the reset always lands too late to protect the segment it opens, and the window where the signal is stale spans an entire reply:

1. A genuine barge-in ends turn N. The reply task calls `clearBuffer()` after `flush()` (`agent_activity.ts:3021`, `3606`, `2658`; `generation.ts:1049`), leaving `interruptedFuture` resolved.
2. Turn N+1 starts. `forwardAudio` calls `resume()`, so early frames flow normally and nothing looks wrong.
3. The user backchannels mid-reply. `interruptByAudioActivity` calls `audioOutput.pause()` (`agent_activity.ts:1531`) and the gate closes.
4. Every remaining frame now reads the turn-N signal, bails, and is discarded. Because bailing returns immediately, the forwarding loop drains the entire TTS stream at full speed instead of blocking — the whole rest of the reply is lost in milliseconds, not just the frames that overlap the pause.
5. The false-interruption timer then finds the handle already `done()` and emits `agent_false_interruption { resumed: false }` without resuming. `pushedDuration` reflects only the pre-pause frames, so `waitForPlayoutTask` reports `interrupted: false` and the turn is committed as complete.

Without the stale signal, step 4 parks one frame at the gate, the loop blocks, and the buffered audio is delivered on resume. That is what the control below shows.

## Evidence

Reproduction built on real in-tree components: a real `AgentSession` (`FakeLLM`, real `AgentActivity`, real recognition hooks), a real `ParticipantAudioOutput` and a real `AudioSource`. Only track publishing is stubbed, since there is no LiveKit server in the test. Frames are counted where they enter `AudioSource.captureFrame` — the wire — so "frames were lost" is measured, not inferred. Each reply is 20 frames x 20ms = 400ms.

`main`:

```
[control] interruptedFuture set before turn = false
[control] pausedMidTurn=true produced=20 delivered=20 lost=0
[control] falseInterruption events = [{"resumed":true}]

[bug] turn1 (genuine barge-in) produced=4 delivered=2
[bug] interruptedFuture set at start of turn 2 = true
[bug] pausedMidTurn=true produced=20 delivered=2 lost=18
[bug] falseInterruption events = [{"resumed":false}]
[bug] assistant chat ctx = [{"text":"first reply","interrupted":true},
                            {"text":"second reply","interrupted":false}]
```

18 of 20 frames — 360ms of a 400ms reply — never reached the wire, and the session recorded `second reply` as fully spoken. The control is the *same* false interruption on an output that has not seen a prior `clearBuffer()`: 0 lost. The only difference between the two is the leftover signal.

With this PR:

```
[bug] pausedMidTurn=true produced=20 delivered=20 lost=0
[bug] falseInterruption events = [{"resumed":true}]
```

Turn 1 is unchanged (`produced=4 delivered=2` — the barge-in still drops its parked frames).

## Fix

Scope the gate to the segment being captured. `clearBuffer()` bumps a monotonic `interruptCount`; the first frame after a flush boundary snapshots it; a frame bails only when the count has advanced past its own segment's snapshot. An interruption that belongs to a speech that is already over no longer applies to the next one.

The pre-existing deadlock guard is preserved and strengthened. A frame parked at the gate when an interruption arrives must still bail rather than hang, and it now waits on a per-frame signal that `clearBuffer()` resolves for every parked frame. Previously a concurrent `flush()` could replace `interruptedFuture` while a frame was parked on the old one, and the interruption that followed would resolve the new instance and never reach it — `_output.test.ts` has a regression test for that (it fails on `main`).

`waitForPlayoutTask`'s own use of `interruptedFuture` is untouched.

## Interaction with #2122

#2122 makes `cancelSpeechPause()` call `clearBuffer()` before `resume()` so that frames parked at the gate are dropped when an interruption is confirmed. That is a deliberate `clearBuffer()` while paused, and it **materially widens this window**. Measured on that branch with the same harness, for the case where a reply completes its playout while paused and the user's transcript then finalizes:

| | `interruptedFuture` after `cancelSpeechPause` | next reply |
| --- | --- | --- |
| `main` | not set | 20/20 delivered |
| #2122 | set | 2/20 delivered, **18 lost** |

On `main` that path never calls `clearBuffer()` at all — the reply task finished cleanly — so it is exposure #2122 introduces. It is a common shape: the agent finishes its sentence while the user talks over the tail, then the transcript finalizes.

This PR fixes both. Verified by applying it on top of #2122: all three scenarios deliver 20/20, and #2122's own behaviour is preserved (its parked frames of the confirmed barge-in are still dropped, `produced=4 delivered=2`, identical with and without this change). #2122 needs no modification — the two are compatible in either merge order. The third test added here (`a paused reply completed and was then cancelled`) passes on `main` today and guards that path from regressing.

Note for #2114, which adds `AudioOutput.abandonOpenSegment()`: that resets the base class's `_capturing` and should also clear `segmentOpen` here, so the two notions of "a segment is open" stay in sync. Not doing so is conservative (a stale snapshot is kept rather than refreshed), so there is no ordering hazard, but it is worth wiring up when both land.

## Test plan

- [x] `pnpm vitest run agents/src/voice` — 534 passed, 52 files
- [x] `npx vitest run agents/src` — 1463 passed, 5 skipped
- [x] New end-to-end test fails on `main` (2/20 frames) and passes here; run 5x for flakiness
- [x] New `_output.test.ts` cases fail on `main` for the right reasons (frame dropped instead of parked; parked frame stranded by a concurrent flush)
- [x] Existing pause-gate tests still pass, including the #1662 segment-counter guard
- [x] `pnpm build:agents`, eslint and prettier clean on touched files
- [x] `pnpm api:check` fails identically with and without this change (pre-existing `export * as` tooling issue); no public API surface changed — all new members are private

### DIFF
--- a/.changeset/stale-interruption-pause-gate.md
+++ b/.changeset/stale-interruption-pause-gate.md
@@ -1,0 +1,20 @@
+---
+'@livekit/agents': patch
+---
+
+Stop a finished interruption from silently discarding the next reply
+
+`ParticipantAudioOutput.clearBuffer()` resolves an `interruptedFuture` that frames parked at
+the pause gate consult to decide whether to bail. Nothing reset that signal until the _next_
+segment's `flush()`, which only runs after that segment's frames have all been captured — so
+for the whole of the following reply the signal still described an interruption that was
+already over. If the output was paused mid-reply during that window (an ordinary
+false-interruption pause, on by default), every remaining frame bailed at the gate and never
+reached the wire. The session still reported the reply as fully spoken and committed it to the
+chat context, so the loss was silent: in a reproduction with a real `AgentSession` and a real
+`ParticipantAudioOutput`, 18 of 20 frames — 360ms of a 400ms reply — were dropped.
+
+The gate is now scoped to the segment being captured: a frame bails only for an interruption
+raised at or after its own segment began. Frames parked at the gate are also woken by a
+per-frame signal, so a `flush()` that replaces `interruptedFuture` can no longer strand one
+there.
--- a/agents/src/voice/false_interruption_audio_loss.test.ts
+++ b/agents/src/voice/false_interruption_audio_loss.test.ts
@@ -1,0 +1,258 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { AudioFrame, type Room, TrackPublishOptions, TrackSource } from '@livekit/rtc-node';
+import { ReadableStream } from 'node:stream/web';
+import { describe, expect, it } from 'vitest';
+import { initializeLogger } from '../log.js';
+import { VADEventType } from '../vad.js';
+import { Agent } from './agent.js';
+import { AgentSession } from './agent_session.js';
+import { AgentSessionEventTypes } from './events.js';
+import { ParticipantAudioOutput } from './room_io/_output.js';
+import { FakeLLM } from './testing/fake_llm.js';
+
+const SAMPLE_RATE = 24000;
+const FRAME_MS = 20;
+const FRAMES_PER_REPLY = 20;
+const FALSE_INTERRUPTION_TIMEOUT = 400;
+
+function frame(): AudioFrame {
+  const samples = (SAMPLE_RATE * FRAME_MS) / 1000;
+  return new AudioFrame(new Int16Array(samples), SAMPLE_RATE, 1, samples);
+}
+
+function vadEvent(type: VADEventType, speechDuration = 0, silenceDuration = 0) {
+  return {
+    type,
+    samplesIndex: 0,
+    timestamp: Date.now(),
+    speechDuration,
+    silenceDuration,
+    frames: [],
+    probability: 1,
+    inferenceDuration: 0,
+    speaking: type === VADEventType.START_OF_SPEECH,
+    rawAccumulatedSilence: 0,
+    rawAccumulatedSpeech: 0,
+  } as never;
+}
+
+function sttFinal(text: string) {
+  return {
+    type: 'final_transcript',
+    alternatives: [{ text, language: 'en', startTime: 0, endTime: 0, confidence: 1 }],
+  } as never;
+}
+
+/** The real ParticipantAudioOutput; only track publishing is skipped (no LiveKit server here). */
+class TestParticipantAudioOutput extends ParticipantAudioOutput {
+  constructor() {
+    super({} as Room, {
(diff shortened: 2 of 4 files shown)


toubatbrian:
Carried into #2126 in full.

Verified there before and after with `false_interruption_audio_loss.test.ts`, which drives a real `AgentSession` against a real `ParticipantAudioOutput` and counts frames at the `AudioSource` boundary. On `origin/main`:

```
FAIL  delivers the whole reply when the previous turn ended in a barge-in
AssertionError: expected 2 to be 20
```

18 of 20 frames — 360ms of a 400ms reply — dropped, while the session still reported the reply as fully spoken. All three tests plus the two new `_output.test.ts` unit tests (both of which also fail on `main`) pass on the consolidated branch.

Closing in favour of #2126.

toubatbrian:
Redirect: this fix now lives in #2132 (stack 3/5, based on #2131).

It was folded into #2126 for a while; that consolidation is closed and split back apart, because bundling five independent defects made each one hard to judge. #2132 carries the segment-scoped pause gate unmodified, along with the `false_interruption_audio_loss.test.ts` evidence (18 of 20 frames dropped on `main`) and the Python comparison explaining why the equivalent is self-healing there.

PR #2130, opened 2026-07-27
fix(krisp): adopt processed frames into the local rtc-node binding

toubatbrian:
**Stack 1/5** — split out of #2126. Base: `main`.

This is the one that caused the reported symptom. Everything above it in the stack is a real defect found on the way, but none of them is what the user hit.

## The defect

The Krisp plugin reaches the LiveKit Cloud backend through `createRequire`, which resolves the internal package's `require` condition and loads the **CJS** build of `@livekit/rtc-node` alongside the ESM one the framework uses. Frames coming back out of the filter are instances of that second copy's `AudioFrame`, so `instanceof AudioFrame` fails everywhere downstream. The adaptive interruption stream treats an unrecognised object as a sentinel and drops it, so the classifier is fed **zero** audio — and with no audio it can only ever fall back to `backchannel`. Every barge-in was ruled a backchannel.

## Evidence

Driven end to end: frames built by the *real* second copy of `@livekit/rtc-node` (resolved with `createRequire` rooted at `plugins/krisp`, exactly as the Cloud backend resolves it) pushed through the real interruption pipeline as `AudioRecognition` wires it, with a mock gateway that rules the first request a barge-in.

```
[repro-2123] cjs constructor=AudioFrame instanceof(esm AudioFrame)=false
[repro-2123] WITHOUT adoption -> {"framesSentToGateway":0,"verdicts":[{"isInterruption":false,"numRequests":0}],"onInterruptionCalls":0}
[repro-2123] WITH adoption    -> {"framesSentToGateway":1,"verdicts":[{"isInterruption":true,"numRequests":1}],"onInterruptionCalls":1}
```

The committed unit test pins the adoption function itself.

## Why this is still needed after the upstream fix

The producer-side fix (livekit/plugin-krisp-viva-internal#13) has merged, but it is unreleased, and the krisp plugin tracks `rtc-node` through a version range rather than a pin — nothing prevents the duplicate resolution from coming back. The root fix (livekit/node-sdks#699, `Symbol.hasInstance` branding) has also merged and is likewise unreleased.

## Python parity

Structurally impossible in Python. `viva_filter.py:262` forwards the backend's frame with no adoption and is correct, because Python resolves through the process-global `sys.modules` — one `livekit.rtc`, one `AudioFrame` per interpreter, no ESM/CJS dual build.

Closes #2123.

### DIFF
--- a/.changeset/fix-krisp-frame-identity.md
+++ b/.changeset/fix-krisp-frame-identity.md
@@ -1,0 +1,13 @@
+---
+'@livekit/agents-plugin-krisp': patch
+---
+
+Fix Krisp-processed audio being invisible to the rest of the pipeline
+
+The LiveKit Cloud backend is reached through `createRequire`, which resolves the internal
+package's `require` condition and so loads the CJS build of `@livekit/rtc-node` next to the
+ESM one the framework uses. Frames returned by that backend were instances of the CJS copy's
+`AudioFrame`, so every `instanceof AudioFrame` downstream failed. Adaptive interruption saw
+zero audio and classified every barge-in as a backchannel, making it impossible to interrupt
+an agent that had noise cancellation enabled. Frames are now adopted into the local binding
+before leaving the filter, sharing their samples rather than copying them.
--- a/plugins/krisp/src/_frame_identity.test.ts
+++ b/plugins/krisp/src/_frame_identity.test.ts
@@ -1,0 +1,51 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { AudioFrame } from '@livekit/rtc-node';
+import { describe, expect, it } from 'vitest';
+import { adoptLocalAudioFrame } from './_frame_identity.js';
+
+/**
+ * Stands in for the `AudioFrame` of a second copy of `@livekit/rtc-node`: identical shape, same
+ * constructor name, unrelated identity. That is exactly what the CJS build produces when the
+ * Cloud backend pulls it in through `createRequire`.
+ */
+class ForeignAudioFrame {
+  constructor(
+    readonly data: Int16Array,
+    readonly sampleRate: number,
+    readonly channels: number,
+    readonly samplesPerChannel: number,
+    private readonly _userdata: Record<string, unknown> = {},
+  ) {}
+
+  get userdata(): Record<string, unknown> {
+    return this._userdata;
+  }
+}
+
+describe('adoptLocalAudioFrame', () => {
+  it('adopts a frame built by another copy of rtc-node', () => {
+    const samples = new Int16Array([1, -2, 3, -4]);
+    const foreign = new ForeignAudioFrame(samples, 16000, 1, 4, { source: 'krisp' });
+
+    // The premise of the bug: this lookalike fails the identity check every consumer relies on.
+    expect(foreign instanceof AudioFrame).toBe(false);
+
+    const adopted = adoptLocalAudioFrame(foreign as unknown as AudioFrame);
+
+    expect(adopted instanceof AudioFrame).toBe(true);
+    expect(adopted.sampleRate).toBe(16000);
+    expect(adopted.channels).toBe(1);
+    expect(adopted.samplesPerChannel).toBe(4);
+    expect(adopted.userdata).toEqual({ source: 'krisp' });
+    // Adopting must not copy the audio: this runs on every frame of every session.
+    expect(adopted.data).toBe(samples);
+  });
+
+  it('returns a local frame untouched', () => {
+    const local = new AudioFrame(new Int16Array([5, 6]), 48000, 1, 2);
+
+    expect(adoptLocalAudioFrame(local)).toBe(local);
+  });
+});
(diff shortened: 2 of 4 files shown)


toubatbrian:
## Live verification with cue-cli

Driven end to end against LiveKit Cloud: real Krisp voice isolation in the input chain, real Deepgram STT, real Inworld TTS, adaptive interruption. cue-cli synthesizes the user side as actual audio, so the barge-in is a genuine overlap. Same scripted scenario throughout — ask for a long story, wait for `AS_SPEAKING`, let it run 6s, then barge in with "Sorry, stop, stop. I changed my mind. What is two plus two?"

`origin/main` (1be8bef5) versus this branch, three runs each. Frame counts come from a probe at the two ends of `ParticipantAudioOutput.captureFrame` — every frame handed in, and every frame that reached `audioSource.captureFrame`.

| | `origin/main` | this branch |
| -- | -- | -- |
| overlap verdict | `false`,`false` / `false` / `false` | `true` / `true` / `true` |
| `pause()` on the sink | **0** / 0 / 0 | 25 / 102 / 57 |
| `clearBuffer()` on the sink | **0** / 0 / 0 | 2 / 4 / 2 |
| audio put on the wire | 549 fr / **53.8s** — 584 fr / **57.3s** — 579 fr / **58.3s** | 103 fr / 12.5s — 182 fr / 20.0s — 103 fr / 13.1s |

Three independent readings of the same failure, 3/3 each way:

1. **The verdict.** Every overlap on `main` comes back `backchannel`. Never once on this branch.
2. **The sink is never even told.** `pause()` and `clearBuffer()` are called **zero** times on `main`. The interruption does not merely arrive late or get overruled — it does not happen at any layer.
3. **What the user hears.** `main` talks straight through the barge-in for **54 to 58 seconds**, the whole story, while the user is speaking over it. With the adoption in place the agent stops after 12–20s.

That third row is the bug report, reproduced and measured.

<details>
<summary>Scenario and harness</summary>

```yaml
turns:
  - say: 'Tell me a very long story about a lighthouse keeper and the sea.'
    until: 'agent_state_changed(.new_state="AS_SPEAKING")'
  - until: '[6s]'
  - say: 'Sorry, stop, stop. I changed my mind. What is two plus two?'
    until: 'agent_state_changed(.new_state="AS_LISTENING") or agent_state_changed(.new_state="AS_IDLE")'
  - until: '[10s]'
```

Two harness details worth stating, since both silently invalidate this kind of run:

- Every stale auto-joining worker on the machine has to be gone. A worker started without `agentName` joins *every* room, answers the cue user, and the run reports a clean pass measured against the wrong build. Two ~14h-old workers were doing exactly that on the first attempt.
- Each run gets a **unique** agent name. A LiveKit `AgentDispatch` outlives the room it was created for and is delivered to the next worker answering to that name, which lands a second job in the process and interleaves two sessions' frames into one probe file. Runs where that happened are detectable (two `job request` lines 2ms apart) and were discarded and redone.
</details>

toubatbrian:
## Live verification: reproduced end to end, and this branch flips it

The defect reproduces in a real room and this fix inverts the outcome, 3 runs to 3 runs, with no overlap between the two groups.

**Method.** `cue-cli` drives a real LiveKit room against the real agent stack — Deepgram nova-3 STT, Inworld TTS, `openai/gpt-4.1-mini`, adaptive interruption, and Krisp voice isolation enabled through the Cloud backend. The script asks for a long story, waits 6 s into the monologue, then barges in with "Sorry, stop, stop. I changed my mind. What is two plus two?". Nothing about the classifier or the Krisp backend is mocked. Three runs on `origin/main`, three on `brian/krisp-frame-identity`, same scenario and same harness.

**Result.** The session's `overlapping_speech` verdicts:

| | run 1 | run 2 | run 3 |
| -- | -- | -- | -- |
| `origin/main` | `isInterruption: false`, `false` | `isInterruption: false` | `isInterruption: false` |
| this branch | `isInterruption: true` | `isInterruption: true` | `isInterruption: true` |

Every barge-in on `main` was ruled a backchannel and the agent talked straight through it. Every barge-in on this branch was ruled a genuine interruption and the agent stopped and answered the new question. One `main` run produced two overlap events; both were `false`.

**Independent corroboration.** Frames handed to the published audio track, counted at `ParticipantAudioOutput`, without reference to the verdict at all: 549 / 584 / 579 on `main` versus 103 / 182 / 103 here. That is 4.4× more audio on `main` comparing run averages (3.0–5.7× across individual pairings), because the agent there is never interrupted and runs the monologue to completion. Same conclusion arriving through a channel that does not read the classifier's output.

**What this does and does not establish.** It is the end-to-end reproduction of the reported symptom on the real Krisp path, and it is the only PR in this stack whose defect I was able to reproduce as a user-visible behaviour change in a live session. It says nothing about the other four PRs; each of those is measured separately in its own thread.

The instrumentation used for the frame counts is throwaway — an `appendFileSync` probe injected into `ParticipantAudioOutput` for the duration of the measurement. It is not committed and was not pushed to any branch; all five branch tips still match `origin`.

chenghao-mou (approved):
lgtm. Though we probably want to remove them after the https://github.com/livekit/node-sdks/pull/700?

toubatbrian:
@chenghao-mou yes, adding this just to make ppl unblocked on older node version since rtc-node is a peer dependency, and might not be upgraded to latest version in time for customer.

PR #2131, opened 2026-07-27
fix(interruption): keep the overlap armed when agent speech restarts mid-interrupt

toubatbrian:
Bottom of a four-PR stack: #2131 → #2132 → #2136 → #2142. Base: `main`.

## Problem

A barge-in can be silently swallowed, because two paths in `interruption_stream.ts` disarm a live overlap:

- **Transport failover** — a retry recreates the stream and replays `agent-speech-started`, wiping the overlap state the old stream was holding.
- **Multi-segment agent turn** — `agent-speech-started` fires once per TTS playout with no idempotence guard, while the matching end fires once per turn. A tool-calling turn emits two starts with no end between them, and the second resets an overlap that is still open.

Either way the user is talking into a classifier that has stopped listening, and only a VAD start-of-speech re-arms it — which will not fire for speech already under way.

## Fix

Skip the overlap reset when a start sentinel arrives while an overlap is already open, and carry overlap state across a failover.

## Testing

Three new cases in `interruption_pipeline.test.ts` fail on `main` and pass here; the existing "resets overlap state on a new agent turn" case is the control and passes on both sides. Also fixes an unhandled `APIConnectionError` rejection `main` emits during failover. Carried here too: the pipeline test harness and the `OverlappingSpeechEvent` re-export, salvaged from the closed #2116.

## Python parity

Same bug — `InterruptionStreamBase._forward_data` (`inference/interruption.py:617-622`) resets unconditionally, start and end sentinels sharing one `case`; the port back splits that case. One metrics-only gap the other way: JS hardcodes `overlapCount = 1` on the replacement stream and loses the audio prefix from its first post-failover window. Verdicts are unaffected.

Written while investigating a live report of agent audio running behind the transcript after a barge-in. That report is now traced elsewhere: the gateway emits multiple `done` messages per `session.flush` on some TTS paths and the client returns on the first, truncating the reply and handing a still-streaming socket back to the pool. Fixes are in flight separately (#2144, #2146).

This PR does not address it. A swallowed barge-in is a different observable, and the report still reproduces with the whole stack applied. Merge on the evidence above and nothing more.

Closes #2117.

### DIFF
--- a/.changeset/adaptive-interruption-overlap-state.md
+++ b/.changeset/adaptive-interruption-overlap-state.md
@@ -1,0 +1,22 @@
+---
+'@livekit/agents': patch
+---
+
+fix(interruption): keep adaptive interruption armed when agent speech restarts mid-overlap
+
+`overlapSpeechStarted` is the gate that lets a user's overlapping audio reach the interruption
+model, and only a VAD start-of-speech raises it. Two events cleared it while the user was still
+talking, and because VAD never re-announces speech that is already under way, nothing could re-arm
+it for the rest of the agent turn: every remaining frame was dropped, no inference request was
+made, and the agent talked straight through the interruption.
+
+- A transport failover rebuilt `InterruptionStreamBase` from scratch and replayed only
+  `agent-speech-started`, which marks the agent as speaking but leaves the overlap disarmed. The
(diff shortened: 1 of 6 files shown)


toubatbrian:

PR #2132, opened 2026-07-27
fix(room_io): stop a finished interruption from dropping the next reply's audio

toubatbrian:
Second of four: #2131 → #2132 → #2136 → #2142. Base: #2131 — review only the top commit; the diff against `main` includes #2131.

## Problem

`ParticipantAudioOutput`'s `interruptedFuture` is resolved by `clearBuffer()` but only reset in `waitForPlayout()`. Between those two points the signal is stale-but-armed, and every frame that parks at the pause gate in that window is discarded — including frames belonging to the *next* reply, which has nothing to do with the interruption that set the flag.

Reachable on defaults: with `resumeFalseInterruption: true` and `falseInterruptionTimeout: 2000`, the reply after a barge-in loses almost all of its audio while the session still reports it as fully spoken.

## Fix

Scope the gate to the segment being captured. `clearBuffer()` bumps a monotonic count, the first frame after a flush boundary snapshots it, and a frame bails only when the count has advanced past its own segment's snapshot. A finished interruption can no longer reach into the next reply.

## Testing

`false_interruption_audio_loss.test.ts` runs a real `AgentSession` against a real `ParticipantAudioOutput` and counts frames at the `AudioSource` boundary: the reply after a barge-in delivers 2 of 20 frames on `main`, 20 of 20 here. Plus two `_output.test.ts` unit tests that also fail on `main`.

## Python parity

Cannot happen in Python — there is no pause gate on `capture_frame`, and both `clear_buffer` and `flush` bail on `if not self._pushed_duration`, so a stale signal is self-healing after at most one frame. JS broke that symmetry by leaving `clearBuffer()` unguarded while `flush()` stayed guarded.

Written while investigating a live report of agent audio running behind the transcript after a barge-in. That report is now traced elsewhere: the gateway emits multiple `done` messages per `session.flush` on some TTS paths and the client returns on the first, truncating the reply and handing a still-streaming socket back to the pool. Fixes are in flight separately (#2144, #2146).

#2124, which this closes, is a distinct symptom with its own reproduction — a whole reply of which the user hears the first fraction of a second. The report above still reproduces with the whole stack applied, so do not merge this as a fix for it.

Closes #2124.

### DIFF
--- a/.changeset/stale-interruption-pause-gate.md
+++ b/.changeset/stale-interruption-pause-gate.md
@@ -1,0 +1,27 @@
+---
+'@livekit/agents': patch
+---
+
+Scope the audio pause gate to the segment being captured
+
+`ParticipantAudioOutput.clearBuffer()` resolves an `interruptedFuture` that frames parked at
+the pause gate consult to decide whether to bail. Nothing reset that signal until the _next_
+segment's `flush()`, which only runs after that segment's frames have all been captured — so
+for the whole of the following reply the signal still described an interruption that was
+already over. If the output was paused mid-reply during that window (an ordinary
+false-interruption pause, on by default), every remaining frame bailed at the gate and never
+reached the wire. The session still reported the reply as fully spoken and committed it to the
+chat context, so the loss was silent: in a reproduction with a real `AgentSession` and a real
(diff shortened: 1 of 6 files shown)


toubatbrian:

PR #2133, opened 2026-07-27
[barge-in] fix(room_io): drop an interrupted reply's TTS backlog instead of playing it

toubatbrian:
Part of the barge-in stack. Base: #2132 — review only the top commit; the diff against `main` includes #2131 and #2132.

> [!IMPORTANT]
> **Depends on #2136 and is inert without it.** Merged in stack order the end state is correct, but this should not be the stack's last landed change — on its own the hoisted check is unreachable on the live barge-in path.

## Problem

`captureFrame` compared a segment's interrupt snapshot only *inside* the branch that handles a closed pause gate, so the check was skipped entirely once the gate reopened.

`cancelSpeechPause` reopens the sink to admit the next reply as soon as the handle is interrupted, but the interrupted reply's `forwardAudio` loop keeps running until its abort signal fires an event loop turn later — and real TTS hands it seconds of audio ahead of realtime to drain in the meantime. Those frames find the gate already open, take the unchecked path, and reach the wire while the next reply's transcript is already streaming.

## Fix

Hoist the interrupt check so it runs on every frame, not only the ones that parked at the gate.

## Testing

`_output_interrupted_segment.test.ts` fails on `main` and passes here, with a control case proving the next reply still plays normally after the interrupted one flushes.

## Scope — please don't over-credit this

Measured live with `cue-cli` as a 2×2 against #2136, three runs per cell, counting frames dropped at the sink during a real barge-in: this fix's own marginal contribution is **one additional ~20ms tail frame** on top of what #2136 already stops. Without #2136 nothing is dropped at all.

It is a genuine defect and worth fixing, but it does not explain the multi-second audio/transcript desync report it was written while chasing. That investigation is still open, and I would not want this merged under that banner.

## Note on `recorder_io.test.ts`

The hoisted check made one recorder test fail; the test was wrong, not the fix. It called `clearBuffer()` without a `flush()`, leaving `segmentOpen` true and the snapshot stale. `forwardAudio` always flushes in its `finally`, so the test now does the same.

### DIFF
--- a/.changeset/interrupted-reply-backlog-on-the-wire.md
+++ b/.changeset/interrupted-reply-backlog-on-the-wire.md
@@ -1,0 +1,13 @@
+---
+'@livekit/agents': patch
+---
+
+Stop an interrupted reply's TTS backlog from reaching the wire
+
+`captureFrame` only compared a segment's interrupt snapshot while the pause gate was closed, so
+the check was skipped entirely once the gate reopened. `cancelSpeechPause` un-gates the sink to
+admit the next reply as soon as the handle is interrupted, but the interrupted reply's
+`forwardAudio` loop keeps running until its abort signal fires an event loop turn later — and real
+TTS providers hand it several seconds of audio ahead of realtime to drain in the meantime. Those
+frames took the open-gate path straight to the wire, so the user kept hearing the barged-over
+speech resume while the next reply's transcript was already streaming.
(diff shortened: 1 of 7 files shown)


toubatbrian:
Folded into #2132. The hoisted `interruptCount` check and `segmentInterruptCount` snapshot are two halves of the same mechanism, and the hoist is inert without the snapshot — keeping them apart made a 3-line logic change look like a comment-only PR. Same code and tests, now in one commit there.

PR #2136, opened 2026-07-27
fix(voice): clear the audio sink before un-gating it for the next speech

toubatbrian:
Third of four: livekit/agents-js#2131 → livekit/agents-js#2132 → livekit/agents-js#2136 → livekit/agents-js#2142. Review only the top commit; the diff against `main` includes livekit/agents-js#2131 and livekit/agents-js#2132. (The base branch is livekit/agents-js#2133's, which is closed and now identical to livekit/agents-js#2132's head.)

> **Replaces** livekit/agents-js#2134, which GitHub closed as merged in error after a bad force-push made its base branch contain its head. Nothing was merged into `main`; the commit is identical (`dc8e1da4`).

## Problem

`cancelSpeechPause()` reopens the sink's pause gate to admit the next speech. The frames parked at that gate belong to the speech it has just interrupted, and they are released first — so audio the user has already barged in over goes back on the wire.

Measured, that is **one 20ms frame**. A correctness fix at a frame boundary, not a fix for an audible regression.

## Fix

Clear the sink before reopening the gate, and only when that same call is what interrupted the paused speech. An `interrupt: false` cancel still resumes without clearing.

## Testing

`confirmed_interruption_pause_and_commit.test.ts` drives a real `AgentSession` against a real `ParticipantAudioOutput` and counts frames at the `AudioSource` boundary; only the model verdict is synthesized. After a confirmed barge-in followed by an STT final transcript, `main` leaks 1 frame and this stack leaks 0.

## Python parity

Parity, not divergence. `_cancel_speech_pause` (`agent_activity.py:4230`) awaits `_wait_for_generation()` before `resume()`, so Python's interrupted reply task has already run its own `clear_buffer()` by the time the gate reopens. JS wraps that await in `waitIfNotInterrupted`, which returns immediately by design (livekit/agents-js#1124), so JS must signal the sink explicitly to land in the same state.

An earlier version also had `onInterruption` call `cancelSpeechPause()`. @chenghao-mou [pointed out](<https://github.com/livekit/agents-js/pull/2126#discussion_r3653386785>) that leaving the speech parked is by design, and Python agrees, so that half was dropped.

## Known limit — read this before merging

On the **no-transcript** path this stack is **identical to** `main`. If nothing arrives inside `falseInterruptionTimeout` (default 2s), all 38 remaining frames of the reply go back on the wire and the session reports `agent_false_interruption(resumed: true)`. That is Python's contract as well, and it is the dead-air trade-off being defended — merging means accepting it.

References [#2122](<https://github.com/livekit/agents-js/issues/2122>); no longer closes it. [#2122](<https://github.com/livekit/agents-js/issues/2122>)'s motivating report — "when I interrupt after asking for a long story, it does pause, but then it continues the audio from where it's been cut off" — is now attributed elsewhere: the gateway emits multiple `done` messages per `session.flush` on some TTS paths, the client returns on the first, and the next reply's pooled socket then serves the previous reply's remaining audio — which sounds exactly like speech resuming where it was cut. Fixes are in flight separately ([#2144](<https://github.com/livekit/agents-js/issues/2144>), livekit/agents-js#2146). That report is **not** addressed here.

What this PR demonstrates is the one-frame leak above. Merge it on that.

Refs [#2122](<https://github.com/livekit/agents-js/issues/2122>).

### DIFF
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -4614,12 +4614,14 @@ export class AgentActivity implements RecognitionHooks {
       return;
     }
 
+    let interruptedPausedSpeech = false;
     if (
       interrupt &&
       !this.pausedSpeech.handle.interrupted &&
       this.pausedSpeech.handle.allowInterruptions
     ) {
       this.pausedSpeech.handle.interrupt();
+      interruptedPausedSpeech = true;
       // ensure the generation is done — but only if a generation
       // was actually started. Must be raced against interrupt: an interrupted
       // paused speech may never mark its generation done, and an un-raced
@@ -4634,6 +4636,15 @@ export class AgentActivity implements RecognitionHooks {
 
     const interruptionOptions = this.agentSession.sessionOptions.turnHandling.interruption;
     if (interruptionOptions.resumeFalseInterruption && this.agentSession.output.audio) {
+      // Frames of the speech just interrupted are parked at the sink's pause gate. Opening the
+      // gate is only meant to admit the *next* speech, but the parked frames are released first
+      // and audio the user has already barged in over reaches the wire. Python blocks here until
+      // the generation finishes, which gets it the same ordering; the await above is raced
+      // against the interrupt (#1124) and so returns immediately, before the interrupted reply
+      // task has run its own clearBuffer(). Signal the interruption first instead.
+      if (interruptedPausedSpeech) {
+        this.agentSession.output.audio.clearBuffer();
+      }
       this.agentSession.output.audio.resume();
     }
   }
--- a/.changeset/clear-sink-before-resume.md
+++ b/.changeset/clear-sink-before-resume.md
@@ -1,0 +1,12 @@
+---
+'@livekit/agents': patch
+---
+
+Signal the interruption before un-gating the audio sink
+
+`cancelSpeechPause()` reopened the sink's pause gate to admit the next speech, but the frames
+parked at that gate belong to the speech it had just interrupted, and they are released first — so
+audio the user has already barged in over reached the wire. Python gets the ordering for free by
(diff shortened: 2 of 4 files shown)


agents/src/voice/agent_activity.ts
```diff
@@ -4634,6 +4636,15 @@ export class AgentActivity implements RecognitionHooks {
...
     const interruptionOptions = this.agentSession.sessionOptions.turnHandling.interruption;
     if (interruptionOptions.resumeFalseInterruption && this.agentSession.output.audio) {
+      // Frames of the speech just interrupted are parked at the sink's pause gate. Opening the
+      // gate is only meant to admit the *next* speech, but the parked frames are released first
+      // and audio the user has already barged in over reaches the wire. Python blocks here until
+      // the generation finishes, which gets it the same ordering; the await above is raced
+      // against the interrupt (#1124) and so returns immediately, before the interrupted reply
+      // task has run its own clearBuffer(). Signal the interruption first instead.
+      if (interruptedPausedSpeech) {
+        this.agentSession.output.audio.clearBuffer();
```
chenghao-mou, line 4646: nit: should we rewrite this as

```
const pausedHandle = this.pausedSpeech.handle;

if (interrupt && !pausedHandle.interrupted && pausedHandle.allowInterruptions) {
  pausedHandle.interrupt();

  if (pausedHandle._hasGenerations) {
    await pausedHandle.waitIfNotInterrupted([pausedHandle._waitForGeneration()]);
  }
}

const shouldClearPausedSpeech = interrupt && pausedHandle.interrupted;
this.pausedSpeech = undefined;

if (shouldClearPausedSpeech) {
  this.agentSession.output.audio.clearBuffer();
}
this.agentSession.output.audio.resume();
```

so we clear when the paused speech handle was interrupted (previously or right before)?

toubatbrian:

PR #2141, opened 2026-07-28
fix(transcription): stop the synchronizer reporting the previous reply's transcript after an interrupt

toubatbrian:
## Problem

`SegmentSynchronizer` can get permanently stuck one segment behind. After an interrupt, every later reply is reported, committed to the chat context, and shown to the user with the **previous** reply's text while a different reply is the one actually being spoken; late in a session the reported text degrades further, to empty.

This is a text-path defect — the audio is correct throughout, only the transcript is wrong. It is **not** the production report of agent audio falling behind the transcript (see Scope).

## Cause

On a barge-in a reply's `playback_finished` usually arrives before its text-forwarding task flushes. The finish rotates the segment, so the late `flush()` ends text input on the *new* impl instead. The next reply's first text chunk sees that impl marked `textInputEnded` and queues it in `_pendingRotatedSegments`, assuming its playback finish must still be in flight — but that impl never carried audio and is owed no finish, so nothing settles it. From then on each real finish is consumed settling the segment queued a turn earlier, which re-queues the current one and makes the off-by-one self-sustaining.

## Fix

Add `SegmentSynchronizerImpl.owesPlaybackFinished` (carried audio downstream, not yet marked finished) and only enqueue a rotated-out segment when it is true. That is exactly the precondition the queue's own comment asserts, so the genuine interrupt-plus-fast-next-reply race it was built for still enqueues.

## Testing

Two regression tests, red before / green after. Voice suite green at 566 tests. The three existing `_pendingRotatedSegments` tests pass unchanged — notably the one covering the case the queue was introduced for.

Written while investigating a live report of agent audio running behind the transcript after a barge-in. That report is now traced elsewhere: the gateway emits multiple `done` messages per `session.flush` on some TTS paths and the client returns on the first, truncating the reply and handing a still-streaming socket back to the pool. Fixes are in flight separately (#2144, #2146). The two defects share a vocabulary and are opposites: this one misreports text against correct audio, that one plays wrong audio against a correct transcript.

Also independent of the interruption stack (#2131 → #2142) — reproduces on `main` regardless of which of those is checked out. Python's `_SyncedAudioOutput.on_playback_finished` only ever reads the current impl and so cannot report another segment's text; this is a JS-only divergence.

### DIFF
--- a/.changeset/fix-stale-synchronized-transcript.md
+++ b/.changeset/fix-stale-synchronized-transcript.md
@@ -1,0 +1,21 @@
+---
+'@livekit/agents': patch
+---
+
+fix(voice): stop the agent transcript from running one reply behind after an interruption
+
+Once a user interrupted the agent, the transcript could permanently desynchronize from the
+audio: every later reply was reported, committed to the chat context, and rendered to the
+user with the _previous_ reply's text while the new reply was the one actually being spoken.
+The lag was one turn and it never recovered for the remainder of the session — asked for a
+story and then for the weather, the agent would speak the weather answer while the transcript
+still showed the story. Late in a session the reported text degraded further, to empty.
+
+The trigger is an interrupted reply whose playback-finished event arrives before its text
+forwarding flushes, which is the ordinary ordering on a barge-in. The flush then lands on the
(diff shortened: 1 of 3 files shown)


toubatbrian:

PR #2142, opened 2026-07-28
fix(voice): release the audio pause a finished speech was still holding

toubatbrian:
Top of a four-PR stack: #2131 → #2132 → #2136 → #2142. Base: #2136 — review only the top commit; the diff against `main` includes #2131 through #2136.

## Problem

When user audio activity pauses the agent mid-reply, the activity records which speech the pause belongs to, so the false-interruption timer can resume that same speech if the overlap turns out to be a backchannel. That record is never cleared when the speech itself finishes, so it outlives its owner.

Two consequences, both measured directly on `main`: the audio sink stays gated indefinitely after a cleanly finished reply, and the next VAD blip to arm the resume timer acts on a speech that is already over — emitting a spurious `agent_false_interruption` for a completed reply.

## Fix

Port the block Python already has. The scheduling loop drops the paused-speech record, cancels the false-interruption timer, and resumes the audio output as soon as the paused speech's generation completes.

## Testing

`paused_speech_completion.test.ts` fails on `main` (`pausedSpeech` still points at the finished handle) and passes here. Voice suite green at 565 tests.

## Python parity

This is a missing port, not a design difference. Python has three `audio_output.resume()` sites and JS had two; the absent one is `_scheduling_task` at `agent_activity.py:1630-1637`, whose own comment reads "clear paused speech after generation done". The JS loop went straight from awaiting the generation to clearing `_currentSpeech`. The added block is that Python block, in the same position.

Found while investigating a live report of stale audio replaying after a barge-in. That report has since been traced, and **not to this**: the inference gateway emits multiple `done` messages per `session.flush` on some TTS paths, the client returns on the first, and the next reply then reads the previous reply's audio off a socket that was handed back to the pool mid-synthesis. Switching only the TTS provider makes the symptom vanish with no code change. Fixes are in flight separately (#2144, #2146).

The reported symptom still reproduces with this whole stack applied. This is a real defect in the pause lifecycle on its own terms, with its own red/green test, and it should be read and merged as that and nothing more.

### DIFF
--- a/.changeset/release-pause-when-paused-speech-completes.md
+++ b/.changeset/release-pause-when-paused-speech-completes.md
@@ -1,0 +1,18 @@
+---
+'@livekit/agents': patch
+---
+
+fix(voice): release the audio pause a finished speech was still holding
+
+When the user's audio activity pauses the agent mid-reply, the activity records which speech
+the pause belongs to so the false-interruption timer can resume that same speech if the
+overlap turns out to be a backchannel. That record was never cleared when the speech itself
+finished, so it outlived its owner: the audio sink stayed gated, and the next thing to read
+the record — a later VAD blip arming the resume timer — acted on a speech that was already
+over. That produced a spurious `agent_false_interruption` event for a completed reply and left
+the sink paused, which any subsequent resume then un-gates without first signalling an
+interruption.
+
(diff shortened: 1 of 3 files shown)


toubatbrian:

PR #2144, opened 2026-07-28
fix(inference): treat a dropped TTS gateway session as a failed attempt

toubatbrian:
## Problem

When the inference gateway ends a TTS session with `session.closed` part-way through a reply, the JS client treats it as a successful completion. One mistake, two consequences.

The rest of the reply is thrown away. `resourceCleanup()` sets `closing`, which kills the input task, so the remaining text is discarded with no error, no warning and no retry. The frame `run()` holds back so it can mark it final is dropped too, so the segment just stops.

And the socket goes back into the `ConnectionPool` mid-synthesis, so the next reply can pick it up and read the previous reply's outstanding audio as its own.

## Fix

Flush the audio the dropped session did produce, mark that segment's last frame final, and reject with a retryable `APIStatusError` so the socket is evicted and the retry finishes the reply. Socket reuse is additionally gated on having observed `done`.

`END_OF_STREAM` is deliberately not emitted — it is the stream terminator that `ttsNode` breaks on, and the stream is not done. The cancelled attempt's input task now reads with its abort signal, so it stops waiting instead of consuming and dropping the chunk of text the retry needs.

## Testing

`tts_session_closed.test.ts` fails on `main` with `expected 6400 to be 8000` (100ms of synthesized audio never delivered) and `expected '' to contain 'He climbed the stairs and lit the lamp.'` (the rest of the reply never resynthesized). `tts_pool_reuse.test.ts` fails on `main` with `expected Set{ 1000 } to deeply equal Set{ 2000 }` — reply 2 speaking only reply 1's audio. Green here; `inference` + `voice` pass at 808 tests.

## Python parity

`inference/tts.py` has no `session.closed` branch at all: the dropped session surfaces as a read timeout or a closed socket, the exception leaves `_run`, and the pool's context manager evicts the connection. Failing the attempt is Python's behaviour. Resolving it was the divergence.

This was opened while root-causing a live report of a reply speaking the *previous* reply's audio after a barge-in, and originally attributed that incident to `session.closed`. **That attribution was wrong** and the description has been corrected.

The real cause of that incident is a separate defect one step earlier: the gateway emits multiple `done` events per `session.flush` on some provider paths, and the client returns on the first one. A direct probe against the gateway records 6 `done`s under a single session id for one flush, the first covering 19.6% of the audio, and switching only the TTS provider to one that emits a single `done` makes the symptom disappear with no code change. Python has that defect identically.

So this PR does **not** fix that report, and it should not be merged as if it did. It stands on its own: treating a dropped session as success truncates the reply and drops the socket back in the pool, and both are wrong regardless of what caused the incident that surfaced them.

### DIFF
--- a/agents/src/inference/tts.ts
+++ b/agents/src/inference/tts.ts
@@ -581,6 +581,14 @@ export class SynthesizeStream<TModel extends TTSModels> extends BaseSynthesizeSt
   protected async run(): Promise<void> {
     let closing = false;
     let lastFrame: AudioFrame | undefined;
+    // Only a `done` from the gateway proves the session owes us no more audio, and a socket
+    // recycled before that hands the leftover audio to whichever SynthesizeStream picks it
+    // up next. `session.closed` is the exit that reaches the pool: it returns from this run
+    // normally, so nothing else evicts the socket. The remaining non-`done` exits — a closed
+    // event channel, a swallowed abort — are only ever reached after `onClose` / `onAbort`
+    // has already removed the socket, so gating reuse on `done` is what keeps reuse tied to
+    // the one event that proves the session is drained rather than to each exit remembering.
+    let sessionDrained = false;
     // Timestamps are delivered in their own WS message; buffer them and attach
     // to the next audio frame that we forward to the output emitter. This
     // mirrors the semantics of `output_emitter.push_timed_transcript` on the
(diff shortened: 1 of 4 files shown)


agents/src/inference/tts.ts
```diff
@@ -631,13 +639,21 @@ export class SynthesizeStream<TModel extends TTSModels> extends BaseSynthesizeSt
     };
 
     const createInputTask = async (signal: AbortSignal) => {
-      for await (const data of this.input) {
-        if (signal.aborted || closing) break;
-        if (data === SynthesizeStream.FLUSH_SENTINEL) {
+      while (!signal.aborted && !closing) {
```
davidzhao, line 642: why would the gateway send `session.closed` in the middle of the session?

davidzhao (approved):

PR #2145, opened 2026-07-28
fix(inference): finish the reply when the TTS gateway drops a session

toubatbrian:
Stacked on #2144.

## Problem

When the inference gateway ends a TTS session with `session.closed` part-way through a
reply, the JS client treats it as a successful completion. The frame `run()` holds back so
it can mark it final is never emitted, and no segment is ever terminated — downstream gets
a segment that just stops. Worse, `resourceCleanup()` sets
`closing`, which kills the input task, so the rest of the reply's text is discarded with
no error, no warning and no retry. In the trace this came from, a ~9000-character reply
had only 2017 characters submitted before the drop; the user never heard the rest.

## Fix

Flush the buffered audio and mark the dropped segment's last frame final, then reject the
attempt with a retryable `APIStatusError` so the socket is evicted and the retry finishes
the reply. `END_OF_STREAM` is deliberately not emitted: it is the stream terminator that
`ttsNode` breaks on, and the stream is not done. The cancelled attempt's input task now
reads with its abort signal so it stops waiting instead of consuming — and dropping — the
chunk of text that belongs to the retry.

## Testing

`agents/src/inference/tts_session_closed.test.ts` fails on the parent branch with
`expected 6400 to be 8000` (100 ms of synthesized audio never delivered) and
`expected '' to contain 'He climbed the stairs and lit the lamp.'` (the rest of the reply
never resynthesized); both pass here. #2144's pool-reuse test is updated because its
premise — that a dropped session completes without a retry — no longer holds. It stays
green, but from here on the socket is evicted by this rejection rather than by #2144's
`sessionDrained` check, which leaves that check with no reachable route; see the note on
#2144.

## Python parity

`inference/tts.py` has no `session.closed` branch: the dropped session surfaces as a read
timeout or a closed socket, an exception leaves `_run`, and the pool's context manager
evicts the connection. Failing the attempt is that behaviour; resolving it was the
divergence.

### DIFF
--- a/.changeset/tidy-eels-repeat.md
+++ b/.changeset/tidy-eels-repeat.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+fix(inference): finish the reply when the TTS gateway drops a session mid-synthesis
--- a/.changeset/tts-pool-drained-session-only.md
+++ b/.changeset/tts-pool-drained-session-only.md
@@ -1,0 +1,20 @@
+---
+'@livekit/agents': patch
+---
+
+fix(inference): only reuse a pooled TTS gateway socket once the session is drained
+
+`inference.TTS` keeps its gateway websockets in a `ConnectionPool` and reuses them across
+replies. A run only proves the gateway owes it no more audio when it receives `done`, but
+several other ways out of `SynthesizeStream.run` also returned normally — a `session.closed`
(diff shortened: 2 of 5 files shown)


toubatbrian:
Squashed into #2144. The two changes are one cause with two consequences on the same `session.closed` branch, and split across two PRs the first one's eviction has no reachable route once this lands — a reviewer would read the same branch twice for no gain.

PR #2146, opened 2026-07-28
fix(inference): treat a TTS gateway `done` as a generation boundary

toubatbrian:
Stacked on #2144, which gates socket reuse on having seen a `done`; this PR changes what a
`done` means. Review only this commit.

## Problem

Several gateway providers answer one `session.flush` with more than one generation, splitting at
roughly 1kB of text or 35s of audio and sending a `done` after each. A probe of
`inworld/inworld-tts-2` got 6 `done`s for one flush, the first covering 39.4s of the 201.0s the
session went on to produce.

Stopping at the first `done` cut every reply longer than one generation off after 20-80% of its
audio while the transcript committed all of it, and released the socket into the
`ConnectionPool` mid-synthesis, so the next reply spoke the previous one's leftover audio. The
session is FIFO and never reset, so each reply queued behind the last one's unsynthesized text
and the lag compounded.

## Fix

`done` is a generation boundary: keep reading, and end the flush only once the session has gone
quiet. A terminal event arriving during that wait is judged on what it proves —
`session.closed` still ends the reply but evicts the socket (`session.create` is only sent when
a socket opens, so a pooled socket would keep the closed session), and an `error` fails the
attempt, because a `done` never promised the reply was complete.

## Testing

A fake gateway splitting one flush into three `done`-terminated generations under one session
id: on `main` the reply speaks 1 of 3 and the next reply speaks the first reply's second
generation. End to end, the barge-in scenario goes from 6 of 10 replies never spoken and the
transcript 26.8s ahead of the audio to 0ms of lag and 1-2 never spoken, in 3 of 3 runs.

## Known limit

The wait is capped by how much synthesized audio is still unplayed, so it hides behind playout
and costs no turn latency. A provider streaming at ~1x realtime has nothing buffered, so the
wait collapses to zero and truncation persists there. Pinned by a test rather than lengthened:
the providers that do split a flush run far faster than realtime. `agent-gateway#1105` fixes the
Inworld path server-side.

## Python parity

`inference/tts.py` has the identical defect — `end_input(); break` on the first `done` — and
needs the same change with the grace in seconds. Filed separately.

### DIFF
--- a/.changeset/tts-multi-generation-flush.md
+++ b/.changeset/tts-multi-generation-flush.md
@@ -1,0 +1,24 @@
+---
+'@livekit/agents': patch
+---
+
+fix(inference): don't stop TTS synthesis at the first gateway `done`
+
+Several inference gateway providers answer one `session.flush` with more than one generation,
+splitting at roughly 1kB of text or 35s of audio and sending a `done` after each. A probe of
+`inworld/inworld-tts-2` saw 6 `done` events for a single flush, the first covering 39.4s of
+the 201.0s the session went on to produce.
+
+The client treated the first `done` as the end of synthesis, so a reply longer than one
+generation was cut off after 20-80% of its audio while the transcript committed all of it. It
+also released the websocket back into the `ConnectionPool` while the gateway was still
+streaming, so the next reply picked up that socket and spoke the previous reply's leftover
+audio. Because the gateway session is FIFO and never reset, each reply then queued behind the
+last one's unsynthesized text and the lag compounded — in the trace this came from, replies
+fell up to 113s behind and 9 of 11 were never audible.
+
+`done` is now treated as a generation boundary. The client keeps reading and only ends the
+flush once the session has stayed quiet, which both delivers the whole reply and means a
+socket is only ever recycled after it has been observed to be idle. The wait for silence is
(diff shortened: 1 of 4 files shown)


toubatbrian:
Closing: the multi-`done` behaviour is a gateway defect, not a protocol feature, and is being fixed at the source in agent-gateway#1105.

The gateway's own design notes treat `done` as the end of one turn/generation, one per flush, across all seven providers. Inworld splitting a single flush at ~1kB / 35s and emitting a `done` per fragment is that contract being broken. Since the gateway is LiveKit-operated rather than self-hosted, once #1105 ships no client sees multiple `done`s and the drain state machine here becomes dead code.

Against that, the cost is real: it diverges from Python, which breaks on the first `done`, and it carries a drain timeout, a reusability flag and reworked error semantics that already drew three P1 review findings. By its own documented limitation the protection also collapses when audio arrives at playback speed.

#2144 is unaffected and still wanted — a session dropped mid-synthesis being reported as a completed reply is an independent defect, and fixing it restores Python parity rather than breaking it.

Reopen if it turns out multiple `done`s per flush are an intended part of the client-facing gateway contract; that would make #1105 the wrong fix and this the right one.

PR #2153, opened 2026-07-28
test(inference): pin pooled TTS socket reuse across a drained session

toubatbrian:
Stacked on #2146.

## Problem

#2146 made a run end on a quiet timeout rather than on the `done`, and #2144 gates returning the socket to `ConnectionPool` on having reached that end. Those two rules meet in one place, so the pool either reuses its socket on every reply or on none — and nothing pinned which. A suspected collapse to a fresh gateway handshake per reply (~157ms on the voice path) was indistinguishable from correct behaviour by reading the code.

## Fix

Test only, no behaviour change. Counts gateway handshakes over a conversation instead of asserting on audio alone.

## Testing

On this branch: 5 clean multi-generation replies dial **once**; a session dropped mid-synthesis dials **twice**, so the eviction that fixes the misattribution bug is unaffected. Deleting `sessionDrained = true` from the quiet-timeout exit turns the first case into **5 dials for 5 replies**, so the test fails closed on exactly the regression it is meant to catch.

Cross-checked against the production gateway rather than only the fake one: `inworld/inworld-tts-2` answers one `session.flush` with 3 `done` events and then goes quiet with no trailing event, and over 5 replies both `main` and this branch dial once — reuse is intact on the stack, not near-zero.

Stacked PRs get no CI, so run locally: `pnpm build`, `pnpm test agents` (115 files, 1508 passed), `pnpm format:check`, `pnpm lint` all pass.

### DIFF
--- a/agents/src/inference/tts_pool_drain_reuse.test.ts
+++ b/agents/src/inference/tts_pool_drain_reuse.test.ts
@@ -1,0 +1,199 @@
+// SPDX-FileCopyrightText: 2026 LiveKit, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+import type { AddressInfo } from 'node:net';
+import { afterEach, describe, expect, it } from 'vitest';
+import { WebSocketServer } from 'ws';
+import type { WebSocket as WsSocket } from 'ws';
+import { initializeLogger } from '../log.js';
+import { TTS } from './tts.js';
+
+/**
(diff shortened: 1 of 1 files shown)


toubatbrian:
Closing with #2146. This pins the drain semantics that #2146 introduces — without that PR there is no quiet-timeout end for a run, so there is nothing here to pin. Pool reuse on its own stays covered by `tts_pool_reuse.test.ts` on #2144.

PR #2162, opened 2026-07-29
feat(google): upgrade @google/genai to 2.x and seed live history with real roles

toubatbrian:
Third in a stack: #2163 → #2165 → this PR. Review the base branch diff only; merge bottom-up.

## Problem

`@google/genai` was pinned to 1.52.0, which predates `historyConfig.initialHistoryInClientContent` (added in 2.12.0). Without that flag the 3.x live models close the socket with 1007 on any prefill containing `model` turns, so an initial chat context had to be flattened into `user` turns with speaker labels (`"Tutor: ..."`, `"Student: ..."`) — which confuses the model.

Separately, Gemini withholds the final input transcript until its reply has finished generating, so every reply rendered ahead of the question it answered in the Cloud transcript.

## Fix

Upgrade to `@google/genai` ^2.13.0 and send `historyConfig.initialHistoryInClientContent` for models that cannot take a mid-session chat context update, so the prefill keeps its real `user`/`model` roles. The flag is spliced into the setup frame because the SDK's converter still drops it (googleapis/js-genai#1520).

The [`HistoryConfig` reference](https://ai.google.dev/api/live#HistoryConfig) also specifies that the seeded history never triggers a model call and that the conversation then starts through `realtimeInput`. A context ending on a user question is therefore seeded as the completed exchanges, followed by that question as realtime text — left inside the history the server files it away and answers nothing. Separately, the 2.x SDK now drains its message queue before `connect()` resolves, which made the `onmessage` callback read `session` inside its temporal dead zone; it reads a holder assigned after the await instead.

Finally, report the generation's creation time as `turnStartedAt` (the field added in #2165) so the user message is placed where the turn happened.

## Testing

22 hermetic realtime tests pass, including one that fails if the question stays in the history. Verified live against `gemini-3.1-flash-live-preview` with cue-cli and in Agent Playground: seeding four real-role turns answers from the history without re-introducing itself, and the uploaded transcript now places each user turn 6-828ms ahead of its reply. On `main` the same run closes with 1007.

## Python parity

Python sends the prefill with `turnComplete: false` and leaves a trailing user turn in the history, so a seeded question goes unanswered there too; this diverges deliberately.

### DIFF
--- a/plugins/google/src/realtime/realtime_api.ts
+++ b/plugins/google/src/realtime/realtime_api.ts
@@ -1100,7 +1100,7 @@ export class RealtimeSession extends llm.RealtimeSession {
               // realtimeInput. A context ending on a user question therefore has
               // to leave the history and arrive as realtime text, or the server
               // silently files it away and answers nothing.
-              const history = [...turns] as types.Content[];
+              const history = [...(turns as types.Content[])];
               const question =
                 history[history.length - 1]?.role === 'user'
                   ? textOf(history[history.length - 1]!)
--- a/plugins/google/src/realtime/realtime_api.ts
+++ b/plugins/google/src/realtime/realtime_api.ts
@@ -144,6 +144,15 @@ interface ToolCallStatus {
   createdAt: number;
 }
 
+/** Empty unless the turn is pure text, so non-text turns stay where they are. */
+function textOf(turn: types.Content): string {
+  const parts = turn.parts ?? [];
+  if (parts.length === 0 || parts.some((part) => !part.text)) {
+    return '';
+  }
+  return parts.map((part) => part.text).join('');
+}
+
 /**
  * Google Realtime Model for real-time voice conversations with Gemini models
  */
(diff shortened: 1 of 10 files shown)


plugins/google/src/realtime/realtime_api.ts
```diff
@@ -473,6 +487,12 @@ export class RealtimeSession extends llm.RealtimeSession {
   private toolResponseCallIds = new WeakMap<types.FunctionResponse, string>();
   private generationPendingTurnComplete?: ResponseGeneration;
 
+  /**
+   * Whether the setup frame declares the leading `clientContent` as history.
+   * When it does, that prefill has to close the history phase (see below).
+   */
+  #seedsHistoryViaConfig = false;
```
davidzhao: nit: the variable name is a bit misleading.. the code has it so that regardless of the value of this, it would always send in history the same way. it's not via config, but it's seeded after the session started.

davidzhao (approved):

plugins/google/src/realtime/realtime_api.ts
toubatbrian: Fair — the prefill goes out the same way either way, after the session starts, so the name described a mechanism that doesn't exist. Renamed to `#prefillReadAsHistory`, since what the config actually changes is whether the server reads that leading content as history instead of as a turn to answer.

PR #2168, opened 2026-07-30
feat(workflows): support deterministic warm-transfer hangup notices

### LINKED ISSUE #2167: WarmTransferTask can repeat an interrupted briefing before the caller-hangup notice
## Describe the bug

When the caller hangs up during a warm-transfer consultation after the human agent has answered but before the merge, `WarmTransferTask` correctly interrupts the active briefing. It then calls `generateReply()` with `callerHangupInstruction`.

Because that generation still has the consultation persona, briefing context, and the agent's interrupted turn, the model can finish or repeat the briefing before complying with the hangup instruction.

In a reproduced call, the human agent heard:

- Initial briefing interrupted at: `"Carrier is unknown. Want me to"`
- Hangup reply: `"Carrier is unknown. Want me to connect or repeat?\n\nThe caller has already left the line, so I am ending the call now."`

The second reply offers an impossible action after the caller has already disconnected. It also makes the human agent wait through repeated context before learning why the call is ending.

This is not an interruption-mechanics issue: `session.interrupt()` does stop the active playout. The problem is using contextual generation for a terminal workflow notification. `toolChoice: 'none'` prevents tool calls, but it does not constrain the generated text to only the hangup notice.

### Expected behavior

Once the caller leaves:

1. Interrupt any active briefing.
2. Play exactly one short, deterministic notice, for example: “The caller disconnected before I could connect you. Hanging up now.”
3. Await that notice with a bounded timeout.
4. Shut down the consultation session.

The notice must not repeat, complete, or summarize the briefing.

The existing silent-cancellation behavior should remain unchanged when the transfer destination has not answered yet.

### Why a deterministic path is needed

Caller abandonment is a terminal state transition with one known meaning, not a conversational turn that benefits from LLM reasoning.

A deterministic notification provides:

- Exact semantics: it cannot offer to connect or repeat after the caller is gone.
- Bounded teardown latency without another LLM inference.
- A path that can avoid both LLM and TTS availability by using prerecorded audio.
- Direct regression testing: interrupt → play one notice → await playout → shutdown.
- Application-controlled wording, language, and voice.

A stronger `callerHangupInstruction` cannot guarantee this because the model still receives the interrupted consultation context.

### Proposed API direction

It would be useful for `WarmTransferTask` to accept a deterministic notice with required text and optional caller-provided audio. For example:

```ts
interface CallerHangupNotice {
  /** Exact text used for transcript output and TTS when no audio is supplied. */
  text: string;

  /** Optional prerecorded audio, created only if the caller actually hangs up. */
  audio?: () => ReadableStream<AudioFrame>;
}

interface WarmTransferTaskOptions {
  callerHangupNotice?: CallerHangupNotice;
}
```

The task would retain ownership of interruption and teardown:

```ts
const handle = session.say(notice.text, {
  audio: notice.audio?.(),
  allowInterruptions: false,
  addToChatCtx: false,
});

await waitUntilAborted(handle.waitForPlayout(), AbortSignal.timeout(10_000));
session.shutdown();
```

This supports both:

- Fixed text through the configured TTS when only `text` is provided.
- Prerecorded playback, while preserving exact text for transcripts, when `audio` is also provided.

The API shape above is illustrative. The important capability is a non-LLM notification path. Existing `callerHangupInstruction` / `generateReply()` behavior could remain the default when no deterministic notice is configured, preserving backward compatibility.

## Relevant log output

```text
playout completed with interrupt
message: "Carrier is unknown. Want me to"

playout completed without interruption
message: "Carrier is unknown. Want me to connect or repeat?

The caller has already left the line, so I am ending the call now."
```

## Describe your environment

- `@livekit/agents`: 1.5.5
- `@livekit/rtc-node`: 0.13.31
- `livekit-server-sdk`: 2.15.3
- Node.js: 22.x
- OS: macOS
- Workflow: inbound SIP caller, `createWarmTransferTask`, outbound SIP warm transfer

The current upstream `main` implementation still uses the same `session.interrupt()` → `session.generateReply()` path:
https://github.com/livekit/agents-js/blob/d4edb58c1f7dc1636f516ef7d379da3a473d97b4/agents/src/workflows/warm_transfer.ts

## Minimal reproducible example

1. Start a warm transfer with consultation instructions that produce a multi-sentence briefing.
2. Have the transfer destination answer.
3. Let the transfer agent begin briefing the destination.
4. Disconnect the original caller mid-sentence, before the calls are merged.
5. Observe that the first briefing is interrupted, but the generated caller-hangup reply may continue or repeat the interrupted briefing before announcing that the caller left.

This requires a SIP trunk and two phone legs; it is not reproducible in the text-only playground.

## Additional information

This behavior was introduced as part of the answered-human notification flow in #2008. That PR initially used a fixed `say()` message, then changed to contextual `generateReply()`. The reproduced behavior demonstrates why consultation context is a liability for this terminal notification.

I can open a PR implementing the deterministic notice path once there is agreement on the preferred API shape.

dtran26:
## Description

Adds an optional deterministic caller-hangup notice for warm transfers. When the caller disconnects after the human agent answers but before the merge, applications can provide exact text and, optionally, prerecorded audio for the notice played before teardown.

This fixes the confusing case where the interrupted consult briefing can be repeated or completed before the generated hangup message. `session.interrupt()` stops active playout, but `generateReply()` still has the consult persona, conversation history, and truncated assistant turn. A caller-hangup notice is a finite-state notification, so the deterministic path uses `session.say()` instead of asking the model to compose it.

Text-only notices use configured TTS. Prerecorded notices accept the existing `AudioSourceType` (`string`, `BuiltinAudioClip`, or `AsyncIterable<AudioFrame>`), so file paths receive the same 48 kHz mono decoding behavior as other framework audio and rate mismatches are resampled downstream. If `say()` cannot be scheduled—for example, a realtime-model session has no TTS—the task logs the error and falls back to `generateReply()` using `callerHangupInstruction` when provided, or the built-in instruction otherwise.

The two options are layers: `callerHangupNotice` is the deterministic primary behavior, while `callerHangupInstruction` controls generated speech when no notice is configured or when the notice cannot be played. Existing applications that only use `callerHangupInstruction` retain their current behavior.

Fixes #2167.

## Changes Made

- Add `callerHangupNotice` with exact `text` and optional `AudioSourceType` audio.
- Resolve file paths and builtin clips through `audioFramesFromFile`; convert async iterables with `ReadableStream.from()`.
- Use one bounded, unref'd timeout signal for audio decoding and playout, and clean up the stream when `say()` throws.
- Fall back to `generateReply()` using a caller-supplied instruction or the built-in default when deterministic speech cannot be scheduled.
- Preserve the normal `say()` conversation-item behavior for transcript consumers.
- Validate empty notice text.
- Test the full `interrupt → notice → await playout → shutdown` lifecycle, customized fallback, timeout, and supported audio forms.

## Pre-Review Checklist

- [x] Build passes (`pnpm build`)
- [x] AI-generated code was reviewed and tested
- [x] The change is scoped to warm-transfer caller abandonment
- [x] Public API behavior is documented
- [ ] Video demo — requires a live two-leg SIP transfer; issue #2167 includes the observed runtime transcript and timing

## Testing

- [x] `pnpm vitest run agents/src/workflows/warm_transfer.test.ts` — 16 passed
- [x] `pnpm test agents/src`
- [x] `pnpm build` — 40 packages built
- [x] `pnpm --filter @livekit/agents typecheck`
- [x] `pnpm --filter @livekit/agents throws:check`
- [x] `pnpm -w lint:fix`
- [x] `pnpm -w format:write`
- [x] `pnpm format:check`

## Additional Notes

The public API shape remains open to maintainer input. In particular: should v1 include the optional `audio: AudioSourceType`, or ship `text` only and add prerecorded audio later? Python parity and naming are also open questions before stabilizing the public option.

Design tradeoffs called out for review:

- The generated-reply fallback favors giving the human agent an audible explanation over silently dropping the call. It logs the underlying `say()` error at `warn`, and applications can supply `callerHangupInstruction` to control the degraded path. The catch is intentionally broader than the no-TTS case because narrowing it would couple this workflow to a private error-message string; an explicit fallback opt-out could be added if exact-wording/compliance use cases require it.
- `AudioConfig` is intentionally not accepted. Its volume scaling is implemented in `BackgroundAudioPlayer`'s mixer, while `session.say({ audio })` uses `performAudioForwarding`, which has no volume stage. Supporting `AudioConfig` here would require a separate per-frame gain implementation.
- The 10-second notice timeout remains an internal teardown safety bound, not a public tuning option, and its timer is unref'd so it cannot keep a worker process alive during shutdown.
- `session.interrupt()` remains the pre-existing synchronous call; the replacement notice is scheduled through the normal speech queue while the interrupted speech unwinds.

The repository-wide `pnpm test` also runs credentialed examples/plugins. Locally it reports unrelated failures from missing `OPENAI_API_KEY` / `CEREBRAS_API_KEY`, one Silero VAD assertion, and two pre-existing unhandled errors in the survey example; the complete `agents/src` suite passes.

`pnpm --filter @livekit/agents api:check` reaches the existing API Extractor limitation at `agents/dist/index.d.ts:10` (`export * as` is not supported). The package build and typecheck pass.

### DIFF
--- a/.changeset/deterministic-warm-transfer-hangup-notice.md
+++ b/.changeset/deterministic-warm-transfer-hangup-notice.md
@@ -2,4 +2,4 @@
 '@livekit/agents': patch
 ---
 
-feat(workflows): support deterministic caller-hangup notices during warm transfers
+Support deterministic caller-hangup notices during warm transfers.
--- a/.changeset/deterministic-warm-transfer-hangup-notice.md
+++ b/.changeset/deterministic-warm-transfer-hangup-notice.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+feat(workflows): support deterministic caller-hangup notices during warm transfers
--- a/agents/src/workflows/index.ts
+++ b/agents/src/workflows/index.ts
@@ -10,6 +10,7 @@ export {
 export {
   WarmTransferTask,
   createWarmTransferTask,
+  type CallerHangupNotice,
   type WarmTransferResult,
   type WarmTransferTaskOptions,
 } from './warm_transfer.js';
--- a/agents/src/workflows/warm_transfer.test.ts
+++ b/agents/src/workflows/warm_transfer.test.ts
@@ -57,17 +57,6 @@ describe('createWarmTransferTask', () => {
     ).toThrow(/must not be empty/);
   });
 
-  it('rejects both caller hangup options together', () => {
-    expect(() =>
-      createWarmTransferTask({
-        sipCallTo: '+15551234567',
-        sipTrunkId: 'ST_dummy',
-        callerHangupNotice: { text: 'The caller disconnected. Hanging up now.' },
-        callerHangupInstruction: 'Generate a short notice.',
-      }),
-    ).toThrow(/cannot both be set/);
-  });
-
   it('rejects an empty deterministic notice', () => {
     expect(() =>
       createWarmTransferTask({
@@ -205,7 +194,7 @@ describe('startCallerHangupSpeech', () => {
     await expect(stream?.getReader().read()).resolves.toEqual({ value: frame, done: false });
   });
 
-  it('cancels decoded audio and falls back when say throws', () => {
+  it('uses a supplied fallback instruction when the deterministic notice cannot be played', () => {
     const { session, say, generateReply, generatedHandle } = createSession();
     const error = new Error('AgentSession is closing, cannot use say()');
     const cancel = vi.fn();
--- a/agents/src/workflows/warm_transfer.test.ts
+++ b/agents/src/workflows/warm_transfer.test.ts
@@ -2,14 +2,33 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import type { AudioFrame } from '@livekit/rtc-node';
-import { describe, expect, it, vi } from 'vitest';
+import { ReadableStream } from 'node:stream/web';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import type * as AudioModule from '../audio.js';
+import { log } from '../log.js';
 import type { AgentSession } from '../voice/agent_session.js';
+import { BuiltinAudioClip } from '../voice/background_audio.js';
 import { SpeechHandle } from '../voice/speech_handle.js';
 import {
-  createCallerHangupSpeech,
   createWarmTransferTask,
+  notifyHumanAgentOfHangup,
   resolveHumanAgentRoomName,
+  startCallerHangupSpeech,
 } from './warm_transfer.js';
+
+const { audioFramesFromFileMock } = vi.hoisted(() => ({
+  audioFramesFromFileMock: vi.fn(),
+}));
+
+vi.mock('../audio.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof AudioModule>();
+  return { ...actual, audioFramesFromFile: audioFramesFromFileMock };
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+  audioFramesFromFileMock.mockReset();
+});
 
 describe('resolveHumanAgentRoomName', () => {
   it('defaults to `<callerRoom>-human-agent` when no override is given', () => {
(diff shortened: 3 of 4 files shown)


dtran26:
Applied the review follow-up in `abf077f5`, with the layered fallback correction in `932cc80d`:

- changed prerecorded audio to the existing `AudioSourceType` and added file/builtin decoding;
- shared a bounded abort signal across decoding and playout, with stream cleanup on scheduling failure;
- added `generateReply()` fallback for realtime/no-TTS sessions, using `callerHangupInstruction` when supplied;
- restored default conversation-item recording;
- kept empty-text validation while allowing notice and fallback instructions to be configured together;
- expanded coverage to the full interrupt → notice → playout → shutdown lifecycle (16 focused tests).

The agents suite, full build, typecheck, throws check, lint, and formatting pass. The PR remains open to maintainer input on whether optional audio should be included in v1 or follow a text-only first release.

dtran26:
Addressed the shutdown-timer follow-up in `903a8d1e` and the layered fallback correction in `932cc80d`:

- unref the 10-second timeout so the notice path cannot keep a worker alive during teardown;
- use a named internal timeout constant;
- document generated-reply fallback on `callerHangupNotice` rather than its `audio` field;
- allow `callerHangupInstruction` to customize the fallback when a notice cannot be played;
- document the intentional 48 kHz mono decode defaults and downstream resampling.

The PR body now records the fallback, `AudioConfig`, timeout, synchronous-interrupt, and Python-parity tradeoffs. Focused tests (16), typecheck, throws check, lint, and formatting remain green.

dtran26:
@toubatbrian ptal when you get a chance, thanks!

PR #2171, opened 2026-07-30
fix(voice): reset tool call createdAt to tool execution start

u9g:
A `FunctionCall` is constructed while the model stream is parsed, before the speech is authorized, so its `createdAt` sorts it ahead of the assistant message that requested it — in chat context, in `RunResult`, and in the uploaded transcript. Reset it when execution begins, as Python does in `_tool_execution_started_cb` (livekit/agents `voice/agent_activity.py`).

No new test: the reset lives in the reply paths' `onToolExecutionStarted` closures, which the existing harnesses mock rather than drive; Python has no dedicated test either. Full agents suite: 1505 passing.

### DIFF
--- a/.changeset/tame-clocks-agree.md
+++ b/.changeset/tame-clocks-agree.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+stamp a tool call's `createdAt` when its execution begins rather than when it was parsed off the model stream, so it no longer sorts ahead of the assistant message that requested it
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -3080,7 +3080,11 @@ export class AgentActivity implements RecognitionHooks {
     //TODO(AJS-272): before executing tools, make sure we generated all the text
     // (this ensure everything is kept ordered)
 
+    // items are ordered by `createdAt`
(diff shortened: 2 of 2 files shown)


toubatbrian:
Agreed, this is the better change — thanks for pushing back. I've closed #2163 and rebuilt my side on top of your approach.

Two things I confirmed while checking it, in case they're useful:

Python's uploader doesn't sort at all — it walks `chat_history.items` and stamps `item.created_at` (`telemetry/traces.py:578-595`) — so source-stamping is the whole strategy there, and this is the missing piece for tool calls.

A live pipeline session then disproved the sort key I'd added in #2163. On the same 28-item session, `metrics.startedSpeakingAt` put a user turn two slots too early, ahead of a tool output and ahead of the reply it followed: that item's `createdAt` was `1785439626231` while its `startedSpeakingAt` was `1785439602375`, a 23.9s gap, because the onset was captured while a slow tool was still running and never re-armed for the utterance that actually committed. Ordering on `createdAt` gets it right.

One dependency worth flagging: #2165 now sorts by `createdAt`, which means a stale tool-call timestamp would become *visible* rather than being hidden by the uploader's monotonic clamp. So this PR should land first — I've marked that on mine.

toubatbrian (approved):

PR #2183, opened 2026-07-31
plugins/phonic: add configsForTools for per-tool behavior overrides

arunwpm-work:
## What

Adds `configsForTools` to the Phonic realtime model — per-tool behavior overrides forwarded to Phonic's tool config. One entry per tool (keyed by `name`); each may set `require_speech_before_tool_call`, `forbid_speech_after_tool_call`, and `forbid_tool_call_after_speech`. Omitted fields fall back to the plugin defaults; tools with no entry are unchanged. `wait_for_speech_before_tool_call` (on) and `allow_tool_chaining` (off) remain fixed by the plugin.

This generalizes the existing `forbidSpeechAfterToolCall: string[]` option into a single per-tool config object, and adds `forbid_tool_call_after_speech` (drop a tool call if the agent already spoke that turn). See the plugin README for the field reference. Changeset included.

### DIFF
--- a/.changeset/phonic-configs-for-tools.md
+++ b/.changeset/phonic-configs-for-tools.md
@@ -2,4 +2,4 @@
 '@livekit/agents-plugin-phonic': patch
 ---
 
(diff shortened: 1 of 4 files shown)


tinalenguyen (approved):
thank you!!

PR #2225, opened 2026-08-05
fix(voice): align session.say transcripts with TTS

### LINKED ISSUE #2067: session.say() transcription is never TTS-aligned — scripted turns drift then flush in one chunk at playback end
**Version:** @livekit/agents 1.5.2 (say path unchanged in 1.5.3 dist)

**Setup:** TTS with aligned transcripts enabled (`cartesia` + `add_timestamps: true`; `useTtsAlignedTranscript` default true). LLM-generated turns caption in sync (aligned mode working). Scripted turns via `session.say(text)` do not.

**Actual:** `say()` captions stream at the estimated fixed pace (`STANDARD_SPEECH_RATE` 3.83 hyphens/s), fall progressively behind actual speech, and when playout finishes the `TranscriptionSynchronizer` flushes the entire remaining text in one visible chunk. On a ~15s scripted greeting this reads as several seconds of trailing captions, then a wall of text at the end.

**Why:** `AgentActivity.ttsTask` (the say pipeline, `agent_activity.js` ~L1753) tees the raw text straight into `transcriptionNode` and never consults `ttsGenData.timedTextsFut` — even though `performTTSInference` runs in that same path and the TTS frames DO carry `USERDATA_TIMED_TRANSCRIPT`. The aligned swap exists only in the generation path (~L2075).

**Expected:** say() turns use the same aligned timed-transcript input as generations when the TTS advertises `alignedTranscript` — scripted openers/goodbyes are exactly the turns where text is known upfront and users notice desync most.

**Context:** we hit this on a production voice-screening product where the scripted opener is the first thing every candidate sees; the mismatch reads as "captions broken" even though the rest of the call is aligned.

swayamg20:
## Summary

- use TTS-provided timed transcripts for `session.say()` when aligned transcripts are enabled and supported
- preserve raw scripted text when alignment is disabled, unsupported, bypassed by caller-provided audio, or audio output is disabled
- add regression coverage for the aligned path and each direct fallback path

Fixes #2067

## Testing

- `pnpm test agents/src/voice/agent_activity_say_aligned_transcript.test.ts agents/src/voice/agent_activity_close_commit.test.ts agents/src/voice/agent_activity_interrupted_commit.test.ts agents/src/voice/generation_tts_timeout.test.ts --silent` (20 passed)
- `pnpm test agents/src --exclude agents/src/voice/remote_session.test.ts --silent` (1,561 passed, 5 skipped; the excluded file has an unrelated current-main event assertion mismatch)
- `pnpm build:agents`
- `pnpm format:check`
- targeted ESLint for the changed voice files
- `pnpm exec changeset status --since=origin/main`

### DIFF
--- a/.changeset/fair-voices-align.md
+++ b/.changeset/fair-voices-align.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Use TTS-aligned transcripts for speech created with `session.say()`.
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -2698,25 +2698,25 @@ export class AgentActivity implements RecognitionHooks {
         .catch(() => this.logger.debug('firstFrameFut cancelled before first frame'));
     }
 
-    const trNode = await this.agent.transcriptionNode(transcriptionInput, {});
-    let textOut: _TextOut | null = null;
-    if (trNode) {
-      const [textForwardTask, _textOut] = performTextForwarding(
-        trNode,
-        replyAbortController,
-        transcriptionOutput,
-      );
-      textOut = _textOut;
-      tasks.push(textForwardTask);
-    }
-
-    if (!audioOutput && textOut) {
-      textOut.firstTextFut.await
-        .then(() => onFirstFrame(null))
-        .catch(() => this.logger.debug('firstTextFut cancelled before first frame'));
-    }
-
     try {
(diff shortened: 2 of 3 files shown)


albertocubeddu:
Any update?

swayamg20:
Thanks for checking in! The fix is ready, and this PR is currently awaiting maintainer review. I’ve checked it against the latest `main`, and it still applies cleanly. I’ll share any updates here.

albertocubeddu:
Applied on my staging and now production works perfectly!

PR #2254, opened 2026-08-10
fix(vad): copy speech buffer when emitting frames instead of aliasing it

pairteamdave:
## Summary

`copySpeechBuffer()` in both the inference VAD and the silero plugin builds the `START_OF_SPEECH` / `END_OF_SPEECH` event frame with `subarray()`, which returns a **view** into the VAD's internal speech buffer rather than a copy.

Immediately after `END_OF_SPEECH` is emitted, `resetWriteCursor()` slides the last `prefixPaddingDuration` of samples to the head of that same buffer to seed the next segment's pre-roll. Because this runs synchronously in the pump loop — before any consumer can observe the event — it overwrites the first samples of the frame that was just emitted.

The audible symptom: any non-streaming STT wrapped in `StreamAdapter` receives the utterance with its own tail prepended, and transcribes duplicated speech. We hit this in production with ElevenLabs `scribe_v1`: the user says "Hello?" and the final transcript is "Hello? Hello?" (the tail of a short utterance is the whole utterance). Longer utterances get their last ~500ms duplicated onto the front, e.g. "Take your time. Jeez" → "Jeez. Take your time. Jeez".

The Python implementation copies the buffer at this point (`bytes.tobytes()`), so this is a porting bug specific to agents-js.

## Fix

Use `slice()` (copy) instead of `subarray()` (view) in `copySpeechBuffer()`, in `agents/src/inference/vad.ts` and `plugins/silero/src/vad.ts`.

## Test

Adds a regression test that drives the real inference-VAD pump with a scripted fake native VAD and a monotonic ramp signal, and asserts the `END_OF_SPEECH` frame still holds the pushed samples after the pump has performed the pre-roll slide. Fails before the fix with the frame's head holding the segment's tail:

```
AssertionError: expected [ 6144, 6145, 6146, 6147 ] to deeply equal [ +0, 1, 2, 3 ]
```

### DIFF
--- a/.changeset/eighty-planets-argue.md
+++ b/.changeset/eighty-planets-argue.md
@@ -1,0 +1,6 @@
+---
+'@livekit/agents': patch
+'@livekit/agents-plugin-silero': patch
+---
+
+fix(vad): copy the speech buffer when emitting START_OF_SPEECH/END_OF_SPEECH frames instead of aliasing it — the pre-roll slide that runs right after END_OF_SPEECH was overwriting the head of the just-emitted frame with the segment's tail, so batch STT (via StreamAdapter) received the end of the utterance prepended to it and transcribed duplicated speech (e.g. "Hello?" → "Hello? Hello?")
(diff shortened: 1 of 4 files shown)


pairteamdave:
Closing — will handle internally.

PR #2295, opened 2026-08-17
feat: add cancellation to warm transfers

### LINKED ISSUE #2288: WarmTransferTask: add cooperative cancellation and a teardown-complete lifecycle
### Feature request

`WarmTransferTask` represents a long-running workflow, but it does not currently expose a supported cancellation or deadline API.

Applications commonly need to stop a warm transfer when:

- the original caller disconnects,
- the consult deadline expires,
- an application is shutting down or draining, or
- a user explicitly cancels the transfer.

Without a cancellation API, consumers have to race `task.run()` against application signals and use deletion of the human-agent room as an indirect stop mechanism. Room deletion is a resource operation rather than a reliable workflow-control signal: the room may already be absent, a participant move may be in flight, and the application cannot tell when caller I/O, hold audio, the human-agent session, and SDK-owned cleanup have actually finished.

The current caller-disconnect path also launches `notifyHumanAgentOfHangup(session)` without awaiting it and then settles the task, so task settlement is not necessarily a teardown-complete boundary:

https://github.com/livekit/agents-js/blob/b4d35f0306c0b39c8b9639d0c484ae4eaf7b75f5/agents/src/workflows/warm_transfer.ts

### Proposed API

Accept an `AbortSignal` either when creating or running the task. For example:

```ts
const controller = new AbortController();

const task = createWarmTransferTask({
  ...options,
  signal: controller.signal,
});

const result = await task.run();
await task.closed;
```

or:

```ts
const result = await task.run({
  signal: AbortSignal.timeout(120_000),
});
```

The exact API shape is illustrative. `AbortSignal` would integrate naturally with timers, worker shutdown, and other Web APIs. There is already a related note in the codebase about moving cancellation propagation toward `AbortSignal`:

https://github.com/livekit/agents-js/blob/b4d35f0306c0b39c8b9639d0c484ae4eaf7b75f5/agents/src/_exceptions.ts

### Suggested lifecycle contract

- Aborting cooperatively stops dialing, ringing, or consulting.
- A participant move that has already committed wins over a concurrent abort; cancellation must not turn an already-bridged transfer into a failed transfer.
- The cancellation reason remains observable, such as `timeout`, `caller_disconnected`, or `application_abort`.
- Callers have a teardown-complete boundary after caller hold audio has stopped, caller I/O has been restored, the human-agent session has shut down, and SDK-owned room cleanup has finished.
- Any human-agent terminal notification is completed before that boundary.
- Existing behavior remains unchanged when no signal is provided.

If changing when `run()` settles would be disruptive, a separate `closed` or `waitUntilClosed()` promise could provide the stronger lifecycle boundary while preserving the current result timing.

### Why this belongs in the SDK

The SDK owns the transfer state machine, participant merge, caller I/O state, hold audio, human-agent session, and its room lifecycle. It is the only layer that can arbitrate cancellation against an in-flight merge and report when teardown is truly complete. Applications should provide the cancellation policy or deadline without having to reproduce those internal lifecycle rules.

### Environment

- Node.js 22.x
- `@livekit/agents` 1.6.3
- Inbound SIP caller with an outbound SIP warm transfer

chenghao-mou:
Warm transfers need a supported cancellation input. Add constructor-level `abortSignal`. Aborting stops waiting for a pending SIP request or ends an active consultation, and `run()` rejects with the signal reason. The SIP API request itself is not cancellable.

A successful participant move wins a concurrent abort.

Addresses AJS-482
Fixes #2288

<details>

**Model:** GPT-5.6

> Can you take a look at https://github.com/livekit/agents-js/issues/2288 and the comments and create a draft PR for this based on the conclusion?
>
> can you compare yours and this PR: dtran26:feat/warm-transfer-abort-signal
>
> https://github.com/livekit/agents-js/pull/2292/changes
>
> okay, port the fixes, and add the original author as a co-author for credit.
>
> can you run a subagent with $thermo-nuclear-code-quality-review ?
>
> yeah, I asked only for an abort signal, now we are tracking states with 100+ lines of code.

</details>

### DIFF
--- a/.changeset/warm-transfer-cooperative-cancellation.md
+++ b/.changeset/warm-transfer-cooperative-cancellation.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Add an optional `abortSignal` to `WarmTransferTask`. Aborting stops waiting for a pending dial or ends an active consultation, and `run()` rejects with the signal reason.
--- a/agents/etc/agents.api.md
+++ b/agents/etc/agents.api.md
@@ -8770,6 +8770,7 @@ class WarmTransferTask extends AgentTask<WarmTransferResult> {
 //
 // @public (undocumented)
 interface WarmTransferTaskOptions {
+    abortSignal?: AbortSignal;
     // (undocumented)
     allowInterruptions?: boolean;
     callerHangupInstruction?: string | null;
--- a/agents/src/workflows/warm_transfer.test.ts
+++ b/agents/src/workflows/warm_transfer.test.ts
@@ -273,4 +273,25 @@ describe('createWarmTransferTask', () => {
 
     expect(complete).toHaveBeenCalledWith(reason);
   });
+
+  it('observes aborts after a participant move fails', async () => {
+    const controller = new AbortController();
+    const reason = new Error('application shutdown');
+    mockDial();
+    vi.spyOn(RoomServiceClient.prototype, 'moveParticipant').mockRejectedValue(
+      new Error('move failed'),
+    );
+    const { complete, create } = setupTransfer(controller.signal);
+
+    const options = create.mock.calls[0]![0];
+    await options.onEnter!({} as never);
+    const connect = (options.tools as FunctionTool[]).find(
+      (entry) => entry.name === 'connect_to_caller',
+    )!;
+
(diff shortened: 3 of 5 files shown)


dtran26:
thank you @chenghao-mou 🙏

theomonnom (approved):

davidzhao (approved):

PR #2296, opened 2026-08-17
fix(voice): preserve pre-set input.audio and output.transcription in RoomIO

1egoman:
## Summary

`RoomIO` preserves a pre-set `session.output.audio` since #2103 (`externalAudioOutput`). This PR applies the same check to the other two IO ports, which are still silently replaced today:

- a pre-set `session.input.audio` was replaced by `ParticipantAudioInputStream`
- a pre-set `session.output.transcription` was replaced by RoomIO's transcription outputs

This matches the python implementation, where `agent_session.start()` disables the corresponding RoomIO port (`room_options.audio_input = False` / `room_options.text_output = False`) whenever the user already set one.

The two remaining `"... is already set, ignoring.."` warnings were misleading — nothing was ignored, the pre-set IO was replaced. They are reworded to the same "preserving and using the existing ..." wording #2103 introduced for audio output.

## Background

Found while investigating an avatar (Beyond Presence) session with no lip sync on agents 1.4.x: before #2103, RoomIO replaced the avatar plugin's `DataStreamAudioOutput`, so the agent published its own audio track while the avatar received zero bytes over `lk.audio_stream` — with no error anywhere, because `ParticipantAudioOutput` self-reports playback events. #2103 fixed the audio-output port; this PR completes the pattern for input and transcription, which have the same silent-replacement behavior.

## Changes

- `agents/src/voice/room_io/room_io.ts`
  - skip creating `ParticipantAudioInputStream` when `session.input.audio` is already set
  - skip creating the room transcription outputs and the `TranscriptionSynchronizer` when `session.output.transcription` is already set (python parity: a pre-set text output disables RoomIO's text output entirely)
- `agents/src/voice/agent_session.ts` — reword the two warnings to reflect the preserve behavior
- `agents/src/voice/room_io/room_io.test.ts` — two new tests mirroring the existing `RoomIO external audio output` suite

## Testing

- `vitest run agents/src/voice/room_io/room_io.test.ts` — 17 passed (15 existing + 2 new)
- full `agents/src` suite — 2138 passed, 5 skipped
- `pnpm build`, `pnpm lint`, `pnpm format:check` pass (`pnpm api:check` fails identically on pristine `main` — pre-existing `export * as` limitation in api-extractor, unrelated)
- end-to-end: a real `AgentSession` with `session.output.audio = new DataStreamAudioOutput(...)` set **before** `session.start()`, against a local avatar worker running the released python `DataStreamAudioReceiver` — audio segments arrive, `lk.playback_finished` round-trips, and transcription sync stays enabled

## Notes

Follow-up candidate (not in this PR, larger change): port python's `output.replace_audio_tail()` and use it in the `DataStreamAudioOutput`-based avatar plugins, so an avatar started **after** `session.start()` composes with the transcript synchronizer instead of replacing it (today that order disables transcription sync with a warning).

### DIFF
--- a/.changeset/witty-pandas-refuse.md
+++ b/.changeset/witty-pandas-refuse.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+RoomIO now preserves a pre-set `session.input.audio` and `session.output.transcription` instead of silently replacing them, matching the existing behavior for `session.output.audio` and the python implementation.
--- a/agents/src/voice/agent_session.ts
+++ b/agents/src/voice/agent_session.ts
@@ -873,7 +873,7 @@ export class AgentSession<
       // Check for existing input/output configuration and warn if needed
       if (this.input.audio && inputOptions?.audioEnabled !== false) {
         this.logger.warn(
-          'RoomIO audio input is enabled but input.audio is already set, ignoring..',
+          'RoomIO audio input is enabled; preserving and using the existing input.audio',
         );
       }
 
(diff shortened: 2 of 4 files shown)


davidzhao (approved):
lgtm!

PR #2306, opened 2026-08-19
fix(interruption): bind inference responses to the overlap they were requested for

### LINKED ISSUE #2119: Adaptive interruption: responses are not bound to the overlap they were requested for, so a late bargein can falsely interrupt a later overlap
### Summary

`bargein_detected` / `inference_done` responses are matched to "whatever overlap is open right now" rather than to the overlap their request was cut for. A response that arrives after its own overlap ended, but while a *later* overlap is open, is accepted and attributed to that later overlap — which can emit a false `isInterruption: true` and cut off the agent for a user utterance the model never scored that way.

This is **pre-existing on `main`** and is not introduced by #2116 / #2117; it was raised by automated review on both of those PRs, so filing it here to track it separately.

### Mechanism

`handleMessage()` gates only on the *current* global overlap flag:

```ts
case MSG_INTERRUPTION_DETECTED: {
  const createdAt = message.created_at;
  const overlapSpeechStartedAt = state.overlapSpeechStartedAt;
  if (state.overlapSpeechStarted && overlapSpeechStartedAt !== undefined) {
    const existing = state.cache.get(createdAt);
    ...
    const entry = state.cache.setOrUpdate(createdAt, () => new InterruptionCacheEntry({ createdAt }), { ... });
```

Two things make this insufficient:

- Nothing ties `created_at` back to the overlap that produced it. The audio transformer clears the cache at `overlap-speech-started`, so a stale response finds `existing === undefined` — but `setOrUpdate()` then *recreates* an entry from scratch, so the clear is not a filter.
- `detectionDelayInS` and the emitted event's `overlapStartedAt` are computed from the **new** overlap's `overlapSpeechStartedAt`, so the resulting event looks like a well-formed verdict for the current overlap.

Ordinary network latency is enough to hit this — no reconnect required:

1. Overlap A is open; a slice is sent at `T`.
2. The user pauses; `overlap-speech-ended` fires; `overlapSpeechStarted` → `false`.
3. The user speaks again; `overlap-speech-started` fires; `overlapSpeechStarted` → `true` (overlap B), cache cleared.
4. The response for the slice sent at `T` arrives. `state.overlapSpeechStarted` is `true`, so it is accepted — and if it is `bargein_detected`, overlap B is reported as an interruption on the strength of overlap A's audio.

Since the gap between two overlaps in one agent turn is often only a few hundred ms (the `keeps making requests across repeated short overlaps in one agent turn` test in `interruption_pipeline.test.ts` exercises exactly that pattern), this is reachable with normal gateway latency.

### Relationship to #2116 / #2117

`handleMessage()` is byte-identical to `main` across both PRs; the only change in that file is the removal of the send-time gate in `transform()`. The send gate never protected the response path — it only decided which slices got *sent*, and the misattribution above happens for slices that were sent perfectly legitimately inside their own overlap.

#2116 does make one *new* variant reachable (a slice parked on an in-flight reconnect being sent after its overlap closed), but that is much rarer than plain response latency, and it is a strictly smaller instance of the same underlying gap.

The per-overlap request accounting added in #2116 introduces `InterruptionAudioSlice.overlapGeneration`, which is the natural hook for the fix: stamp the generation onto the `InterruptionCacheEntry` in `sendAudioData()` and reject responses in `handleMessage()` whose request belongs to a generation that is no longer open.

Note that this needs a store that survives `cache.clear()` (or the generation has to be recorded outside the bounded cache), since a stale response's entry is exactly the one that got cleared — and simply requiring `existing !== undefined` would be wrong, because the `BoundedCache` holds only 10 entries and can legitimately evict a live request during a long overlap, which would suppress real interruptions.

### Origin

Reported by Codex review on #2116 and #2117 ("Keep late responses bound to their original overlap" / "Reject responses belonging to a completed overlap"). Verified against the code; not reproduced at runtime.

mastermanas805:
The issue notes this wasn't reproduced at runtime, so this starts with a failing test: overlap A sends a request, A ends, B opens, and A's `bargein_detected` lands and is reported as an interruption of B. Ordinary response latency is enough to hit it.

`InterruptionAudioSlice.overlapGeneration` from #2116 isn't on `main` — #2116 and #2117 were both closed unmerged. So this adds an `overlapGeneration` that bumps per overlap and never resets, stamps it onto each request in `sendAudioData()`, and rejects responses whose generation has since closed. It's deliberately separate from the existing `overlapCount`, which restarts each agent turn for the audio-prefix shift — reusing that would let turn N's generation collide with turn N+1's.

`inference_done` needed the same guard. It emits no event, but it writes into `cache`, and a stale write poisons the entry `overlap-speech-ended` later pops for its verdict.

The check fails open: an unrecorded generation counts as current, since treating an evicted request as stale would suppress a real interruption — worse than the misattribution being fixed. Same reason the ledger sits outside `cache`, which is cleared at every overlap boundary.

Regression covers the late response, the cross-turn case, the fail-open path, and a control that still interrupts normally.

Fixes #2119

### DIFF
--- a/agents/src/inference/interruption/interruption_stream.ts
+++ b/agents/src/inference/interruption/interruption_stream.ts
@@ -168,6 +168,12 @@ export class InterruptionStreamBase {
     let overlapSpeechStarted = false;
     let overlapCount = 0;
     const cache = new BoundedCache<number, InterruptionCacheEntry>(10);
+    // Which overlap each in-flight request was cut for, keyed like `cache` by the request's
+    // `createdAt`. Deliberately kept outside `cache`: that one is cleared at every overlap
+    // boundary, which is exactly when a late response still needs to be identifiable. Bounded
+    // well above the ~10 requests/s one overlap produces, so a live request is not evicted
+    // mid-overlap.
+    const requestGenerations = new BoundedCache<number, { generation: number }>(256);
     const inferenceS16Data = new Int16Array(
       Math.ceil(this.options.maxAudioDurationInS * this.options.sampleRate),
     ).fill(0);
@@ -178,6 +184,7 @@ export class InterruptionStreamBase {
       overlapSpeechStartedAt: this.overlapSpeechStartedAt,
       cache,
       overlapCount,
+      requestGenerations,
     });
     const setState = (partial: { overlapSpeechStarted?: boolean }) => {
       if (partial.overlapSpeechStarted !== undefined) {
@@ -245,6 +252,7 @@ export class InterruptionStreamBase {
             startIdx = 0;
             this.numRequests = 0;
             cache.clear();
+            requestGenerations.clear();
           } else if (chunk.type === 'agent-speech-ended') {
             this.logger.debug('agent speech ended');
             agentSpeechStarted = false;
@@ -255,6 +263,7 @@ export class InterruptionStreamBase {
             startIdx = 0;
             this.numRequests = 0;
             cache.clear();
+            requestGenerations.clear();
           } else if (chunk.type === 'overlap-speech-started' && agentSpeechStarted) {
             this.overlapSpeechStartedAt = chunk.startedAt;
             this.userSpeakingSpan = chunk.userSpeakingSpan;
--- a/agents/src/inference/interruption/interruption_stream.ts
+++ b/agents/src/inference/interruption/interruption_stream.ts
@@ -167,12 +167,16 @@ export class InterruptionStreamBase {
     let accumulatedSamples = 0;
     let overlapSpeechStarted = false;
     let overlapCount = 0;
+    // Identifies an overlap for the lifetime of the stream. Distinct from `overlapCount`, which
+    // resets every agent turn and drives the audio-prefix shift: a response that outlives its
+    // turn must still be distinguishable, and per-turn counts collide across turns.
+    let overlapGeneration = 0;
     const cache = new BoundedCache<number, InterruptionCacheEntry>(10);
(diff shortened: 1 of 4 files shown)


mastermanas805:
Both review points addressed in b078075.

The cross-turn one was real: clearing the ledger at the agent-speech boundary sent stale responses straight down the fail-open path, and `overlapCount` restarting at 0 each turn meant keeping the entries wouldn't have helped either. There's now a separate `overlapGeneration` that never resets, and a regression test that fails without it. Changeset added.

agents/src/inference/interruption/interruption_stream.ts
```diff
@@ -167,7 +167,17 @@ export class InterruptionStreamBase {
     let accumulatedSamples = 0;
     let overlapSpeechStarted = false;
     let overlapCount = 0;
+    // Identifies an overlap for the lifetime of the stream. Distinct from `overlapCount`, which
+    // resets every agent turn and drives the audio-prefix shift: a response that outlives its
+    // turn must still be distinguishable, and per-turn counts collide across turns.
+    let overlapGeneration = 0;
```
chenghao-mou, line 173: we've fixed this with a simpler implementation: https://github.com/livekit/agents/pull/6957, can we use `cache.updateValue()` and ignore missing entries in both handlers, then remove the generation ledger and fail-open test?

For an interruption verdict that, for some reason, failed to interrupt the agent speech, we are going to fall back to VAD interruption path in another PR, because running inference with the same/similar audio context will have the same prediction anyway.

chenghao-mou (commented):
thanks for the PR! I have left one comment.

PR #2309, opened 2026-08-20
fix: restore Changesets with js-yaml 4

chenghao-mou:
The js-yaml security update broke the local Changesets CLI. Use its compatible YAML reader and resolve every affected js-yaml 3.x and 4.x path to patched version 4.3.1.

Validated with a frozen install, no js-yaml audit findings, Changesets status, and an isolated Changesets version run.

Addresses AGT-3345

<details>

**Model:** GPT-5.6

> > Separate issue: the current js-yaml override breaks the unshimmed local Changesets CLI. Since the release workflow invokes that CLI, this should be fixed before release.
>
> what do you mean?
>
> okay, can you create a PR to fix that?
>
> wouldn't 2309 re-introduce the security issue?
>
> yeah, please do that

</details>

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -23,8 +23,8 @@ overrides:
   protobufjs@<7.6.5: ^7.6.5
   ws@>=8.0.0 <8.21.0: ^8.21.0
   read-yaml-file@<2.0.0: ^2.1.0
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
thanks for fixing it!

PR #2319, opened 2026-08-21
fix(voice): preserve held STT turn boundaries

Thytu:
## Description

Preserve an accepted STT turn boundary when agent speech starts between the final transcript and its `END_OF_SPEECH` sentinel.

The final transcript is accepted before agent speech, but the following sentinel is buffered while the agent is speaking. During the later flush, the sentinel has no transcript alternatives, so it cannot set `shouldFlush` and is discarded. The user turn remains uncommitted and can be merged with later speech.

## Changes Made

- Re-emit a held `END_OF_SPEECH` when STT has accepted transcript text for an uncommitted turn.
- Keep the exception scoped to STT turn detection and buffers that actually contain the pending boundary.
- Add a regression covering the exact final → agent speech → end sentinel → agent speech end ordering.

## Pre-Review Checklist

- [x] **Build passes**: Full monorepo build and typecheck pass locally
- [x] **AI-generated code reviewed**: Comments and test setup were reviewed and kept to the state-machine invariant
- [x] **Changes explained**: The failed transition and bounded exception are documented above
- [x] **Scope appropriate**: Only held STT turn-boundary handling and its regression are changed
- [x] **Video demo**: Not applicable; this is an internal event-ordering race covered by a deterministic regression

## Testing

- [x] Regression fails on `main` with `userTurnCommitted === false`
- [x] Regression passes with the fix
- [x] `pnpm test agents/src` — 136 files passed; 2,248 tests passed; 5 skipped
- [x] `pnpm build`
- [x] `pnpm typecheck`
- [x] `pnpm format:check`
- [x] `pnpm -w lint:fix`

## Additional Notes

The unscoped test command also runs provider/example integration tests. Locally those reported unrelated missing `OPENAI_API_KEY`/`CEREBRAS_API_KEY`, Hugging Face download timeouts, and one FishAudio assertion. The complete `agents/src` suite passes.

### DIFF
--- a/agents/src/voice/audio_recognition.ts
+++ b/agents/src/voice/audio_recognition.ts
@@ -1026,8 +1026,20 @@ export class AudioRecognition {
       }
     }
 
+    // A final can be accepted just before agent speech; its held END_OF_SPEECH must still
+    // close that pending STT turn even though the sentinel has no transcript alternatives.
+    const hasPendingAcceptedTurnBoundary =
+      this.turnDetectionMode === 'stt' &&
+      !this.userTurnCommitted &&
+      this.audioTranscript.length > 0 &&
+      emitFromIndex !== null &&
+      this.transcriptBuffer
+        .slice(emitFromIndex)
+        .some((event) => event.type === SpeechEventType.END_OF_SPEECH);
     const eventsToEmit =
-      emitFromIndex !== null && shouldFlush ? this.transcriptBuffer.slice(emitFromIndex) : [];
+      emitFromIndex !== null && (shouldFlush || hasPendingAcceptedTurnBoundary)
+        ? this.transcriptBuffer.slice(emitFromIndex)
+        : [];
 
     // Snapshot the ignore-until before resetting so the added-delay diagnostic below mirrors
     // the value the holding decision was made against.
--- a/.changeset/quiet-turns-close.md
+++ b/.changeset/quiet-turns-close.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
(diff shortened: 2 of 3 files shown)


agents/src/voice/audio_recognition.ts
```diff
@@ -1026,8 +1026,20 @@ export class AudioRecognition {
...
+    // A final can be accepted just before agent speech; its held END_OF_SPEECH must still
+    // close that pending STT turn even though the sentinel has no transcript alternatives.
+    const hasPendingAcceptedTurnBoundary =
+      this.turnDetectionMode === 'stt' &&
+      !this.userTurnCommitted &&
+      this.audioTranscript.length > 0 &&
+      emitFromIndex !== null &&
+      this.transcriptBuffer
+        .slice(emitFromIndex)
+        .some((event) => event.type === SpeechEventType.END_OF_SPEECH);
```
chenghao-mou, line 1038: Could we handle this when the EOS arrives? If it would be the first buffered event and it closes an accepted, uncommitted STT turn, we can process it immediately:

```diff
 private shouldHoldSttEvent(ev: SpeechEvent): boolean {
   if (!this.isInterruptionEnabled) {
     return false;
   }

   if (this.isAgentSpeaking) {
-    return true;
+    const closesPendingSttTurn =
+      ev.type === SpeechEventType.END_OF_SPEECH &&
+      this.turnDetectionMode === 'stt' &&
+      !this.userTurnCommitted &&
+      this.audioTranscript.length > 0 &&
+      this.transcriptBuffer.length === 0;
+
+    return !closesPendingSttTurn;
   }
```

chenghao-mou (commented):
Thanks for the PR! Left one comment

PR #2330, opened 2026-08-24
fix(openai): reject TTS bodies that aren't raw PCM

### LINKED ISSUE #2320: openai TTS hardcodes response_format: 'pcm' and never checks the response, so a compatible endpoint returning anything else is played as raw samples
### Describe the bug

`plugins/openai/src/tts.ts` asks for `response_format: 'pcm'` and then treats whatever comes back as raw samples, without decoding it or looking at the response `Content-Type`:

```ts
response_format: 'pcm',
...
const buffer = await this.stream.then((r) => r.arrayBuffer());
const audioByteStream = new AudioByteStream(OPENAI_TTS_SAMPLE_RATE, OPENAI_TTS_CHANNELS);
const frames = audioByteStream.write(buffer);
```

`response_format` is also hardcoded, so it can't be changed through `TTSOptions` or `updateOptions()`.

Against api.openai.com this is fine you ask for `pcm`, you get `pcm`. But `baseURL` is a public option in `TTSOptions`, and `get provider()` reads the host off `this.#client.baseURL`, so pointing the plugin at an OpenAI-compatible server looks like an intended use. If that server doesn't support `pcm`, or ignores `response_format` and answers with its own default, the bytes are written into the audio buffer as 16-bit samples.

Serving a 9644-byte WAV (44-byte RIFF header + 9600 bytes of 24 kHz mono PCM) emits 9600 bytes whose **first four bytes are the ASCII characters `RIFF`**. The container header is played as audio; only a trailing partial frame is dropped.

For a WAV that happens to match the expected sample rate, the result is a short click before otherwise-correct audio, which is easy to miss. For a compressed format mp3, opus, aac, flac none of the body is PCM, so the whole utterance is noise. Either way there is no error, no warning and no retry.

livekit/agents#6930 fixed the equivalent problem on the Python side by choosing the parser from the response `Content-Type`.

### Relevant log output

```
(none the failure is silent; no error, warning or retry is emitted)
```

### Describe your environment

```
  System:
    OS: Windows 11 10.0.26200
    CPU: (14) x64 Intel(R) Core(TM) Ultra 5 235U
  Binaries:
    Node: 24.19.0
    pnpm: 11.13.1
  npmPackages:
    @livekit/agents-plugin-openai: 1.7.0 (verified against main @ 31d66e4)
```

Run as a vitest test inside the repo:

```ts
import { OpenAI } from 'openai';
import { describe, expect, it } from 'vitest';
import { TTS } from './tts.js';

const pcm = Buffer.alloc(9600);
const wav = (() => {
  const h = Buffer.alloc(44);
  h.write('RIFF', 0); h.writeUInt32LE(36 + pcm.length, 4); h.write('WAVE', 8);
  h.write('fmt ', 12); h.writeUInt32LE(16, 16); h.writeUInt16LE(1, 20); h.writeUInt16LE(1, 22);
  h.writeUInt32LE(24000, 24); h.writeUInt32LE(48000, 28); h.writeUInt16LE(2, 32); h.writeUInt16LE(16, 34);
  h.write('data', 36); h.writeUInt32LE(pcm.length, 40);
  return Buffer.concat([h, pcm]);
})();

it('does not play a wav container as if it were pcm', async () => {
  const client = new OpenAI({
    apiKey: 'test',
    baseURL: 'https://compatible.example.com/v1',
    maxRetries: 0,
    fetch: async () =>
      new Response(new Uint8Array(wav), { status: 200, headers: { 'content-type': 'audio/wav' } }),
  });
  const tts = new TTS({ client, model: 'kokoro' });

  const chunks: Buffer[] = [];
  for await (const ev of tts.synthesize('hello')) {
    chunks.push(Buffer.from(ev.frame.data.buffer, ev.frame.data.byteOffset, ev.frame.data.byteLength));
  }
  const out = Buffer.concat(chunks);

  expect(out.subarray(0, 4).toString('latin1')).not.toBe('RIFF'); // currently fails
});
```

### Additional information

A fix could mirror what landed in Python: read `Content-Type` off the response and decode accordingly, falling back to the requested format when the server doesn't declare something usable. Exposing `response_format` through `TTSOptions` would also help, since some compatible backends don't offer `pcm` at all.

Worth noting that `plugins/openai/src/tts.test.ts` currently requires a live `OPENAI_API_KEY` and skips without one, so nothing exercises this path in CI. A hermetic test like the one above would cover it without network access.

I'm happy to open a PR if you'd like I wrote the equivalent change for the Python plugin, so the shape would be familiar.

abidullahdev:
Closes #2320.

The plugin asked for `pcm` and then wrote whatever came back into an `AudioByteStream` as 16-bit samples, with no decoder and no look at the response `Content-Type`. Against api.openai.com that's fine. Against an OpenAI-compatible server that doesn't support `pcm`, or ignores `response_format` and answers with its own default, the body is played as audio: a container header becomes a click, and a compressed body becomes noise for the whole utterance. No error, no warning, no retry.

### What this does

Rejects bodies the plugin cannot play, and lets `response_format` be set:

- A `Content-Type` in a small deny set (`audio/mpeg`, `audio/wav`, `audio/ogg`, …) now throws `APIError(..., { retryable: false })` instead of being emitted as samples.
- `responseFormat` is exposed on `TTSOptions`, still defaulting to `pcm`.

The deny set is deliberately a denylist rather than a PCM allowlist. Unknown types  `application/octet-stream` in particular  keep flowing, because a compatible server may return raw PCM without labelling it precisely, and I didn't want to break setups that work today.

### Why this doesn't mirror the Python fix

livekit/agents#6930 fixed the analogous bug by selecting a decoder from the response `Content-Type`. That isn't available here: `agents/src/audio.ts` only exposes file-based ffmpeg helpers (`audioFramesFromFile`), so there's no way to decode a response buffer without building a streaming decoder first. So JS turns silent corruption into a clear error rather than decoding  the behaviour still differs from Python, and closing that gap properly would mean adding a decoder. Happy to look at that separately if it's wanted.

`APIError` with `retryable: false` follows the existing precedent in `plugins/elevenlabs/src/tts.ts`, which does the same content-type check for the same reason, and keeps the failure recognisable to `instanceof APIError` call sites such as `tts/fallback_adapter.ts`.

### Tests

Added to `plugins/openai/src/tts.test.ts`, using a stubbed `fetch` so they run in CI without network or an API key the existing suite in that file requires `OPENAI_API_KEY` and skips without one, which is why this path had no coverage.

Three of the five fail on `main`. Serving a 9644-byte WAV (44-byte RIFF header + 9600 bytes of 24 kHz mono PCM) currently emits 9600 bytes beginning with the ASCII characters `RIFF`.

Also regenerated `etc/agents-plugin-openai.api.md`, since the new public option and type weren't in the report.

### DIFF
--- a/.changeset/openai-tts-reject-undecodable-audio.md
+++ b/.changeset/openai-tts-reject-undecodable-audio.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-openai': patch
+---
+
+fix(openai): reject TTS bodies that aren't raw PCM instead of playing them as samples, and allow response_format to be set
--- a/plugins/openai/etc/agents-plugin-openai.api.md
+++ b/plugins/openai/etc/agents-plugin-openai.api.md
@@ -1897,14 +1897,17 @@ export interface TTSOptions {
     instructions?: string;
     // (undocumented)
     model: TTSModels | string;
-    // Warning: (ae-forgotten-export) The symbol "TTSResponseFormat" needs to be exported by the entry point index.d.ts
-    // Warning: (ae-unresolved-link) The @link reference could not be resolved: The package "@livekit/agents-plugin-openai" does not have an export "UNPLAYABLE_CONTENT_TYPES"
     responseFormat?: TTSResponseFormat;
     // (undocumented)
     speed: number;
     // (undocumented)
     voice: TTSVoices;
 }
+
+// Warning: (ae-missing-release-tag) "TTSResponseFormat" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+//
+// @public
+export type TTSResponseFormat = NonNullable<OpenAI.Audio.SpeechCreateParams['response_format']>;
 
 // Warning: (ae-missing-release-tag) "TTSVoices" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
 //
--- a/plugins/openai/etc/agents-plugin-openai.api.md
+++ b/plugins/openai/etc/agents-plugin-openai.api.md
@@ -1897,6 +1897,9 @@ export interface TTSOptions {
     instructions?: string;
     // (undocumented)
     model: TTSModels | string;
+    // Warning: (ae-forgotten-export) The symbol "TTSResponseFormat" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-unresolved-link) The @link reference could not be resolved: The package "@livekit/agents-plugin-openai" does not have an export "UNPLAYABLE_CONTENT_TYPES"
+    responseFormat?: TTSResponseFormat;
     // (undocumented)
     speed: number;
     // (undocumented)
--- a/plugins/openai/src/index.ts
+++ b/plugins/openai/src/index.ts
@@ -9,7 +9,7 @@ export * from './tools.js';
 export * as realtime from './realtime/index.js';
 export * as responses from './responses/index.js';
 export { STT, type STTOptions } from './stt.js';
-export { ChunkedStream, TTS, type TTSOptions } from './tts.js';
+export { ChunkedStream, TTS, type TTSOptions, type TTSResponseFormat } from './tts.js';
 export * as ws from './ws/index.js';
 
 class OpenAIPlugin extends Plugin {
--- a/plugins/openai/src/tts.test.ts
+++ b/plugins/openai/src/tts.test.ts
@@ -2,9 +2,10 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import { tts } from '@livekit/agents-plugins-test';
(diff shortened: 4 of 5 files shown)


abidullahdev:
Fixed in the follow-up commit `TTSResponseFormat` is now re-exported from `index.ts` and the API report regenerated.

That also cleared a second warning I'd introduced: the report had an `ae-unresolved-link` for a `{@link UNPLAYABLE_CONTENT_TYPES}` that I'd already removed from the source, so the committed report had been generated from a stale build. Both warnings my change added are gone now.

I left `TTSResponseFormat` without a release tag, since `ae-missing-release-tag` appears 151 times in that report and `TTS`/`TTSOptions` carry it too tagging only the new type seemed more inconsistent than matching the file.

PR #2360, opened 2026-08-28
fix(workflows): re-attach the human agent room close listener when a warm transfer merge fails

### LINKED ISSUE #2358: `WarmTransferTask`: a failed participant move permanently detaches `onHumanAgentRoomClose`, so a human agent hangup no longer completes the task
### Summary

`mergeCalls` detaches `onHumanAgentRoomClose` before it awaits `moveParticipant`. If the move throws, the catch in `connect_to_caller` re-attaches the *abort* listener but not that one — so for the rest of the call, the human agent hanging up no longer completes the task.

Your own tests show the asymmetry. [`warm_transfer.test.ts:448`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.test.ts#L448) and [`:476`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.test.ts#L476) both mock `moveParticipant` to reject, then assert the abort path still settles the task. The move-failure case was considered; one of the two exits was restored.

### The sequence

- [`:584`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.ts#L584) removes the abort listener
- [`:434`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.ts#L434), inside `mergeCalls`, removes `onHumanAgentRoomClose`
- [`:446`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.ts#L446) awaits the move
- [`:588-595`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.ts#L588-L595) catches, re-attaches only the abort listener, rethrows

Detaching before the move is right for the success case — the old room closing afterwards is expected, not a failure. It's the failure path that loses an exit it needed.

### Why `closeOnDisconnect` doesn't cover it

The obvious answer is that a human agent hangup closes the consultation session, `deleteRoomOnClose` deletes the room, and `Disconnected` fires `onHumanAgentRoomClose`. That's true — and it's exactly the chain the detach disarms.

It also can't win the race, because the ordering is fixed rather than chancy. `closeImplInner` awaits `activity.drain()` ([`agent_session.ts:1927`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/voice/agent_session.ts#L1927)) and only emits `Close` afterwards ([`:1975`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/voice/agent_session.ts#L1975)). The in-flight tool call is one of the pending tasks that drain waits on. So the room delete is always ordered after the tool body that already ran the `off()`.

A hangup with more lead time is handled correctly — the chain completes and the `task.done` guard at [`:578`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.ts#L578) rejects the late `connect_to_caller`. The window is the tool's own execution.

### What the caller experiences

`setIoEnabled(false)` ([`:669`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.ts#L669)) disables the agent's audio in, audio out and transcription for the consultation, and it's only re-enabled in `onExit`, which runs on task completion. So on this path the agent never resumes listening or speaking, and default hold audio keeps looping. The caller's own mic is untouched — they can talk, nobody is listening.

### What isn't broken

Worth stating, since it narrows the fix: the room and session cleanup still happen, so there's no room or WebSocket leak. Only task completion is lost. An application-supplied `abortSignal` also still ends it — the option doc at [`:48-56`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.ts#L48-L56) offers it for "a consult deadline", and it's re-attached at `:593`. So this is bounded for anyone passing one, and unbounded for anyone who isn't.

An LLM retry isn't a way out either: retrying `connect_to_caller` re-enters `mergeCalls`, which fails the same way or throws `'calls are not ready to merge'` at [`:430`](https://github.com/livekit/agents-js/blob/10cc5f4fe9626fae80174157790c992e9be53c77/agents/src/workflows/warm_transfer.ts#L430) — neither calls `setResult`. And in the case that matters, the human agent is gone, so there's no turn to drive a retry.

### Suggested fix

Re-attach `onHumanAgentRoomClose` in the catch, next to the abort listener. I'd avoid moving the `off()` to after the move instead — that opens a window where a `Disconnected` between the move returning and the detach would report a successful transfer as failed, which the current placement prevents.

Happy to send a PR if you'd take one.

### Caveats

Read from `main` at `10cc5f4`, not reproduced against a live call. I'm assuming `moveParticipant` rejects once the participant is gone rather than succeeding as a no-op — if it no-ops, the window is narrower than I think, though the tests above suggest a rejecting move is a case you already expect.

ketan0095:
## Description

Fixes #2358.

`mergeCalls` detaches `onHumanAgentRoomClose` at `warm_transfer.ts:434` before awaiting `moveParticipant`. That is correct for the success case — moving the human agent out tears their room down, and that close is expected teardown rather than a failure.

When the move throws, the `catch` in `connect_to_caller` re-attaches the abort listener but not this one. From then on the human agent hanging up calls no `setResult`, `task.done` stays `false`, and the caller is left on hold with the agent's IO still disabled (`setIoEnabled(true)` only runs in `onExit`, which needs the task to complete).

This re-attaches the listener alongside the abort listener.

## Changes Made

- `warm_transfer.ts`: re-attach `onHumanAgentRoomClose` in `connect_to_caller`'s catch, next to the existing abort listener re-attach.
- `warm_transfer.test.ts`: two tests — one that the human agent hanging up after a failed move completes the task, and one that a room close *during* an in-flight move does not (which is what the `off()` at `:434` exists to prevent, and was previously untested).
- Changeset.

## Pre-Review Checklist
- [ ] **Video demo**: not applicable — the trigger is a `moveParticipant` rejection, which isn't reachable from the playground.

## Testing
- [x] All tests pass

`pnpm vitest run` in `agents/`: 2282 passed, exit 0. `pnpm build` (tsup + `tsc --declaration`) exit 0. `pnpm lint` 0 errors.

Both tests were checked against mutations rather than just run:

| Mutation | Result |
|---|---|
| Remove the re-attach (the bug) | fails |
| Remove the `off()` at `:434` | fails |
| Remove both | fails |
| Re-attach to `callerRoom` / wrong handler / wrong event | fails |
| Drop the reason from the `ToolError` message | fails |

## Additional Notes

Placement — the re-attach sits *after* the `if (abortSignal?.aborted)` early return, so an aborted signal doesn't re-attach. That branch calls `setResult`, which nulls `humanAgentRoom`, so there'd be nothing to attach to. The `?.` is load-bearing for a different reason: `cancelForCallerHangup` can null `humanAgentRoom` between the `off()` and the catch if the caller hangs up mid-move.

Considered and not done — moving the re-attach inside `mergeCalls`'s own try/catch so the pair is local and symmetric. The whole file still passes that way, but it also re-attaches on the abort path where this deliberately doesn't, so it didn't seem clearly better. Happy to switch if you prefer it. I also considered dropping the `off()` entirely in favour of a `mergeSucceeded` flag on the handler, but that's isomorphic to the attach/detach and adds another mutable field to the closure.

One case worth naming: if `moveParticipant` rejects but the server did perform the move (a timeout on a successful request), the re-attached listener converts the eventual room close into a `room closed:` result instead of success. Still better than the caller being stranded, and the local participant stays in the human agent room until session shutdown so it won't fire promptly.

### DIFF
--- a/.changeset/warm-transfer-reattach-room-close.md
+++ b/.changeset/warm-transfer-reattach-room-close.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Re-attach the human agent room close listener when a warm transfer merge fails. A human agent who hangs up after the failed move now completes the task instead of leaving the caller on hold.
--- a/agents/src/workflows/warm_transfer.test.ts
+++ b/agents/src/workflows/warm_transfer.test.ts
@@ -2,12 +2,12 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import { SIPParticipantInfo } from '@livekit/protocol';
-import { ParticipantKind, Room, RoomEvent } from '@livekit/rtc-node';
+import { DisconnectReason, ParticipantKind, Room, RoomEvent } from '@livekit/rtc-node';
 import { AccessToken, RoomServiceClient, SipClient } from 'livekit-server-sdk';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import * as job from '../job.js';
 import type { JobContext } from '../job.js';
-import { ChatContext, type FunctionTool } from '../llm/index.js';
+import { ChatContext, type FunctionTool, ToolError } from '../llm/index.js';
 import { AgentTask } from '../voice/agent.js';
 import { AgentSession } from '../voice/agent_session.js';
 import { BackgroundAudioPlayer } from '../voice/background_audio.js';
@@ -492,6 +492,71 @@ describe('createWarmTransferTask', () => {
     controller.abort(reason);
 
     expect(complete).toHaveBeenCalledWith(reason);
+  });
+
+  it('completes when the human agent hangs up after a participant move fails', async () => {
+    const controller = new AbortController();
+    mockDial();
+    let humanAgentRoom: Room | undefined;
+    vi.spyOn(AgentSession.prototype, 'start').mockImplementation(async (startOptions) => {
+      humanAgentRoom = startOptions.room;
+    });
+    vi.spyOn(RoomServiceClient.prototype, 'moveParticipant').mockRejectedValue(
+      new Error('move failed'),
+    );
+    const { complete, create } = setupTransfer(controller.signal);
+
+    const options = create.mock.calls[0]![0];
+    await options.onEnter!({} as never);
+    const connect = (options.tools as FunctionTool[]).find(
+      (entry) => entry.name === 'connect_to_caller',
+    )!;
+
+    await expect(connect.execute({}, {} as never)).rejects.toThrow('move failed');
+    expect(complete).not.toHaveBeenCalled();
+
+    humanAgentRoom!.emit(RoomEvent.Disconnected, DisconnectReason.CLIENT_INITIATED);
+
+    expect(complete).toHaveBeenCalledOnce();
+    expect(complete).toHaveBeenCalledWith(
+      new ToolError(`room closed: ${DisconnectReason.CLIENT_INITIATED}`),
+    );
+  });
+
+  it('ignores the human agent room closing while the participant move is in flight', async () => {
+    const controller = new AbortController();
+    mockDial();
+    let humanAgentRoom: Room | undefined;
+    vi.spyOn(AgentSession.prototype, 'start').mockImplementation(async (startOptions) => {
+      humanAgentRoom = startOptions.room;
+    });
+    let finishMove!: () => void;
+    const move = new Promise<void>((resolve) => {
+      finishMove = resolve;
+    });
+    const moveParticipant = vi
+      .spyOn(RoomServiceClient.prototype, 'moveParticipant')
+      .mockReturnValue(move as never);
+    const { complete, create } = setupTransfer(controller.signal);
+
+    const options = create.mock.calls[0]![0];
+    await options.onEnter!({} as never);
+    const connect = (options.tools as FunctionTool[]).find(
+      (entry) => entry.name === 'connect_to_caller',
+    )!;
+    const merging = connect.execute({}, {} as never);
+    await vi.waitFor(() => expect(moveParticipant).toHaveBeenCalled());
+
+    // Moving the human agent out tears their room down; that is the expected
+    // end of a successful merge, not a failure.
+    humanAgentRoom!.emit(RoomEvent.Disconnected, DisconnectReason.CLIENT_INITIATED);
+    expect(complete).not.toHaveBeenCalled();
+
+    finishMove();
+    await merging;
+
+    expect(complete).toHaveBeenCalledOnce();
+    expect(complete).toHaveBeenCalledWith({ humanAgentIdentity: 'human-agent-sip' });
   });
 
   it('starts greeting speech only after the outbound SIP call answers', async () => {
--- a/agents/src/workflows/warm_transfer.ts
+++ b/agents/src/workflows/warm_transfer.ts
@@ -590,6 +590,10 @@ export function createWarmTransferTask({
             setResult(asError(abortSignal.reason ?? new Error('warm transfer aborted')));
             return;
           }
+          // The move failed, so the human agent is still in their own room: a
+          // close is a failure again, not the post-move teardown `mergeCalls`
+          // detached for.
+          humanAgentRoom?.on(RoomEvent.Disconnected, onHumanAgentRoomClose);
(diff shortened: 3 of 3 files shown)


dtran26:
We hit #2358 in production on 1.7.1 (Cloud telephony, high-volume warm transfers) and have server-side room events for two calls that show a second failure shape this PR's catch-based re-attach doesn't cover: **the tool being cancelled rather than throwing, with the room-close event firing before any re-attach could run.**

Observed sequence (consult room `RM_BLrF6HnJjUVh`, timestamps UTC; second call identical):

1. `15:51:24` — `human-agent-sip` answers and joins the consult room.
2. Rep speaks; the transfer agent delivers the briefing; `connect_to_caller` (aliased) is executing — `mergeCalls` has already detached `onHumanAgentRoomClose`.
3. `15:52:00` — the rep hangs up: `human-agent-sip` leaves `CLIENT_INITIATED`, at the exact end of the briefing's playout. (Structural, not bad luck: the merge only executes once playout ends, which is also the natural moment for an unconvinced human to bail.)
4. The consult session's `closeOnDisconnect` teardown force-interrupts and **aborts the in-flight tool task** — the `catch` in `connect_to_caller` never runs (the tool span ends with no output and no error attributes; filed the observability side as #2374).
5. `15:52:01` — session close deletes the consult room (`deleteRoomOnClose`); `RoomEvent.Disconnected` fires — but the listener was detached in step 2, and even with this PR, a re-attach inside `catch` would happen (if at all) *after* this event has already fired.
6. Nothing settles the task. The caller stayed on dead hold ~4.5 more minutes until our application-level consult cap aborted the transfer.

So the fix needs to survive both orderings: (a) the tool aborted instead of throwing, and (b) the close event arriving before any re-attach. The `mergeSucceeded`-flag alternative mentioned in the PR body gets closest, with one refinement: keep the listener attached throughout, track the in-flight move promise **outside** the cancellable tool task, and on room close while a move is in flight, await the move's settlement — ignore the close if the move succeeded (expected teardown), complete the task with the room-closed error otherwise (including when the move was cancelled/never settles because its owning task was aborted). Equivalently: any re-attach must be followed by a room/connection state check that completes the task if the room is already gone.

Happy to share the full pair of room-event sequences and worker traces if useful — between them they cover rep-hangup-before-merge, rep-hangup-during-merge, and the successful-merge control case.

dtran26:
Follow-up: #2361 (merged) does cover the shape I described above. The `!room.isConnected` state check after re-attach handles the close-event-fired-first ordering, and in the cancelled-tool case the underlying `execute` keeps running past the task abort, so when `moveParticipant` settles the `catch` still runs and the state check completes the task. The only residual gap I can see is a `moveParticipant` call that never settles at all, which is narrow enough to leave alone. Thanks — we'll pick this up in the release that carries it.

ketan0095:
Closing in favour of #2361 — thanks for picking it up so quickly.

Your version is the better one. I'd considered the `isConnected` check and left it out to keep the diff minimal, which was the wrong call: it's what makes the close-fired-during-the-move ordering work, and my catch-only re-attach would have missed it. The parameterised before/after test is a nicer shape than my two separate ones, too.

@dtran26 — thanks for the production room events, genuinely useful. I'd filed this off a static read of `main` and flagged that I hadn't reproduced it, so having two real calls confirm the shape within a day was the best possible outcome.

One thing #2361 doesn't carry over, if it's wanted: both of its cases fail the move, so the *success* path is still unpinned — a `Disconnected` arriving while the move is in flight, where the move then succeeds, must complete with `{ humanAgentIdentity }` and not `room closed`. That's the case the `off()` at `mergeCalls` exists for, and as far as I can see nothing in the suite covers it today, so deleting that line still passes. Happy to send it as a ~20-line test-only follow-up if you'd take one — otherwise I'll leave it.

PR #2361, opened 2026-08-28
fix: handle room disconnect after failed warm transfer

chenghao-mou:
A failed participant move could detach the consultation room's close observer and leave the transfer waiting forever. Restore observation after failure, recover a disconnect missed during the move, and warn when the room disconnects before completion. Successful moves still treat source-room teardown as expected.

Addresses AGT-3390
Related to #2358
Supersedes #2360

<details>

**Model:** GPT-5.6

> investigate agt3390
>
> okay, yeah let's fix that. So a failed move caused the hang, right?
>
> could it be too late if the room is already disconnected?
>
> okay, yeah, we should check room state like how we check abortSignal aborted state first. But also, we don't need to do this for success path, right? But why if so?
>
> do we have a warning log about this? it could be helpful to know human agent disconnected, right?

</details>

### DIFF
--- a/.changeset/restore-warm-transfer-room-close.md
+++ b/.changeset/restore-warm-transfer-room-close.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Fix `WarmTransferTask` hanging after a failed participant move when the human agent disconnects.
--- a/agents/src/workflows/warm_transfer.test.ts
+++ b/agents/src/workflows/warm_transfer.test.ts
@@ -2,12 +2,13 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import { SIPParticipantInfo } from '@livekit/protocol';
-import { ParticipantKind, Room, RoomEvent } from '@livekit/rtc-node';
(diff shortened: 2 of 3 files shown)


tinalenguyen (approved):

dtran26:
@chenghao-mou Following up on the "helpful to know human agent disconnected" thread here. I filed livekit/agents#7200 (feature requests are routed to the Python repo per this repo's issue config) asking `WarmTransferTask` to report a typed failure code and keep the underlying evidence: the human-agent participant's `disconnectReason` and `sip.callStatus` at departure, and the `SipCallError` from the dial as `cause`. The contract keeps the listener discipline this PR established around the move: the participant-left listener only records, the existing room-close and dial-rejection handlers stay the only completers, and it detaches before the move and re-attaches on failure.

Node specifics: the same four `setResult` sites at main b11b7b4 (`warm_transfer.ts` L286, L362, L617, L626, L707), and typing the dial cause needs the `livekit-server-sdk` floor raised to 2.17.0 for `SipCallError`. Production evidence in the issue is from `@livekit/agents` 1.8.0.

PR #2372, opened 2026-08-28
fix: cancel speech refused while scheduling is paused

### LINKED ISSUE #2373: say() on a paused activity leaks an unabortable speech task; the next drain deadlocks forever — AgentTask.run() never settles and close() hangs
### Environment

- `@livekit/agents` 1.7.1 (also reproduces on 1.6.4 dist)
- `@livekit/rtc-node` 0.13.34
- Telephony pipeline (SIP), LiveKit Cloud-hosted agents — but the repro below needs no rooms/audio at all

### Summary

If `session.say()` is called while the current activity's scheduling is paused (e.g. an `AgentTask` handoff such as `WarmTransferTask` is in flight), the SDK throws `SchedulingPausedError` **after** it has already registered the speech task. The orphaned task can never complete, and the next `drain()` of that activity waits on it forever. Downstream: `AgentTask.run()` never settles, the owning function tool never resolves (no `FunctionCallOutput`), `resume` back to the previous agent never happens, and `AgentSession.close()` hangs behind the activity lock.

We hit this in production at meaningful scale: a function tool finished during a warm transfer's pause window and spoke its result via `session.say(...)`; after the consult merged successfully, the caller session was permanently wedged and the job only ended via external cleanup. Trace signature: a `drain_agent_activity` span that never ends (its `on_exit` child *does* end), no `resume_agent_activity`, and a close timeout.

### The three interacting defects

**1. `AgentActivity.say()` registers the speech task before scheduling can fail** (`agents/src/voice/agent_activity.ts`):

```js
const task = this.createSpeechTask({ taskFn: ..., ownedSpeechHandle: handle, name: 'AgentActivity.tts_say' });
task.result.finally(() => this.onPipelineReplyDone(stateLease));
this.scheduleSpeech(handle, SpeechHandle.SPEECH_PRIORITY_NORMAL);   // throws SchedulingPausedError when paused
```

When `scheduleSpeech` throws, the created task stays in `speechTasks`. Its `ttsTask` is blocked in `waitIfNotInterrupted([_waitForAuthorization(), ...])`:
- never **authorized** — the handle never entered `speechQueue`, and only `mainTask` authorizes queued handles;
- never **interrupted** — `interrupt()` only walks `speechQueue`/`_currentSpeech`;
- not **abortable** — the authorization wait does not observe the task's own abort controller, so even `cancelAndWait()` cannot reclaim it.

**2. `AgentSession.say()` redirects into a not-yet-started, still-paused `nextActivity`** (`agents/src/voice/agent_session.ts`):

```js
if (activity.schedulingPaused) {
  if (!nextActivity) throw ...;
  return nextActivity.say(text, options);
}
```

During `_updateActivity(previousActivity: 'pause', newActivity: 'start')` the next activity still has its initial `_schedulingPaused = true` (cleared only in `_startSession`), so the redirect guarantees defect 1 fires — and plants the orphan on the **incoming** activity, the one that will be drained when the task unwinds.

**3. The drain has no escape hatch.** `_drainImpl` → `_pauseSchedulingTask([])` → `await this._mainTask.result`, and `mainTask` only exits when `getDrainPendingSpeechTasks()` is empty. One orphaned task ⇒ deadlock, with no timeout anywhere on the path (`mainTask.result`, and later `detachSttPipeline()`'s bare `cancelAndWait()`). The existing `allZombie` check doesn't cover never-queued/never-authorized handles.

### Reproduction (standalone, no rooms/audio)

```bash
mkdir /tmp/lk-repro && cd /tmp/lk-repro
npm pack @livekit/agents@1.7.1 && tar xzf livekit-agents-1.7.1.tgz && cd package
npm install --omit=dev --ignore-scripts && npm install @livekit/rtc-node --ignore-scripts
# save the script below as repro.mjs, then:
node repro.mjs   # exits 2 after "close TIMED OUT"; comment out the say() block and it exits 0 cleanly
```

```js
// repro.mjs — say() during an AgentTask pause window wedges the unwind and close()
import { initializeLogger } from './dist/log.js';
import { Agent, AgentTask } from './dist/voice/agent.js';
import { AgentSession } from './dist/voice/agent_session.js';
import { FakeLLM } from './dist/voice/testing/fake_llm.js';
import { tool, ToolError } from './dist/llm/tool_context.js';
import { FunctionCall } from './dist/llm/chat_context.js';

// FakeLLM hardcodes callId `fake_call_${index}` per response; uniquify across turns
let callSeq = 0;
const origCreate = FunctionCall.create.bind(FunctionCall);
FunctionCall.create = (opts) => origCreate({ ...opts, callId: `${opts.callId}_${callSeq++}` });

initializeLogger({ pretty: true, level: 'warn' });
const log = (...a) => console.log(new Date().toISOString().slice(17, 23), ...a);

let consultTask;
class ConsultTask extends AgentTask {
  constructor() { super({ instructions: 'consult task' }); }
  async onEnter() { log('ConsultTask.onEnter'); }
  async onExit() { log('ConsultTask.onExit'); }
}

const llm = new FakeLLM([
  { input: 'counter', content: 'Give me one moment.', toolCalls: [{ name: 'evaluate_bid', args: {} }] },
  { input: 'agree', content: 'Sure, please hold.', toolCalls: [{ name: 'transferCall', args: {} }] },
  { input: 'transfer done', content: 'You are connected.' },
]);

const agentA = new Agent({
  instructions: 'root agent',
  tools: [
    tool({
      name: 'transferCall', description: 'transfer the call',
      execute: async () => {
        log('transferCall: running ConsultTask');
        consultTask = new ConsultTask();
        const res = await consultTask.run();            // <-- never resolves once the orphan exists
        log('transferCall: task.run resolved:', JSON.stringify(res));
        return 'transfer done';
      },
    }),
    tool({
      name: 'evaluate_bid', description: 'slow tool racing the transfer',
      execute: async () => {
        await new Promise((r) => setTimeout(r, 2000)); // finishes mid-pause
        throw new ToolError('unable right now');
      },
    }),
  ],
});

const session = new AgentSession({ llm });
await session.start({ agent: agentA });

session.generateReply({ userInput: 'counter' });          // starts the slow tool
setTimeout(() => session.generateReply({ userInput: 'agree' }), 50); // starts the transfer task

// THE TRIGGER — a say() while the pause is in flight (our tool wrapper speaks tool results).
// Comment this block out and everything resolves + close() completes.
setTimeout(() => {
  try { session.say('spoken tool result'); }
  catch (e) { log('say() threw:', e.constructor.name, e.message); } // SchedulingPausedError — but the task already leaked
}, 1000);

setTimeout(() => { log('completing task (consult merge analog)'); consultTask.complete({ ok: true }); }, 3000);

setTimeout(async () => {
  const closed = await Promise.race([
    session.close().then(() => true),
    new Promise((r) => setTimeout(() => r(false), 8000)),
  ]);
  log(closed ? 'close completed' : 'close TIMED OUT');
  process.exit(closed ? 0 : 2);
}, 12000);
```

Observed with the say(): `SchedulingPausedError` at ~1s, `task.run()` never resolves after `complete()`, the task activity's drain logs pending `['AgentActivity.tts_say']` forever, no resume, and `close()` times out. Without the say(): clean resume and close. Same behavior against 1.6.4 dist.

### Suggested fixes

1. In `say()` (and any speech creation), don't leave the speech task registered when `scheduleSpeech` throws — schedule first, or cancel/deregister the created task in a catch.
2. `AgentSession.say()` should not redirect into a pre-start paused activity — defer the speech until the activity starts, or throw before creating the task.
3. Defense in depth: make `ttsTask`'s authorization wait observe its abort controller; bound `_pauseSchedulingTask`'s `mainTask.result` wait and `detachSttPipeline`'s `cancelAndWait()` during drain; extend the `allZombie` check to handles that were never queued/authorized.

Happy to provide the full production trace exports (span timelines showing the unended drain) privately if useful.

chenghao-mou:
**Why:** During an `AgentTask` handoff, `session.say()` can target a paused next activity. Its unscheduled speech task then waits forever and blocks later drains and `AgentSession.close()`.

**What:** Keep parity with Python next-activity routing. If the scheduler refuses speech, force-interrupt its handle so the task settles. Task cancellation and STT handoff cleanup add defense in depth. The regression test requires normal `session.close()` resolution.

**Python reference:** livekit/agents#4730, [`_schedule_speech`](https://github.com/livekit/agents/blob/d3e4279ed3c637e08e7462201b475e33b0925fc0/livekit-agents/livekit/agents/voice/agent_activity.py#L1050-L1065)
```python
if self._scheduling_paused and not force:
    speech.interrupt(force=True)
    raise RuntimeError(
        "cannot schedule new speech, the speech scheduling is draining/pausing, the speech will be cancelled"
    )

if self._scheduling_atask and self._scheduling_atask.done():
    logger.warning(
        "attempting to schedule a new SpeechHandle, but the scheduling_task is not running, the speech will be cancelled"
    )
    speech.interrupt(force=True)
    return
```

Addresses AGT-3403

Fixes #2373

<details>

**Model:** Claude Fable 5

> I got a customer report: FYI noticed a regression in worker shutdown / finalization after upgrading the SDK from 1.6.4 to 1.7.1: [...] Are there known 1.7.1 regressions involving: AgentActivity.drain() or AgentSession.close(), AgentTask handoff or inline-task serialization, shutdown of nested AgentSession instances, worker jobs remaining alive during shutdown [...] Could you inspect the task/activity state for these job IDs and tell us what prevented their job processes from responding and closing? Can you investigate?

> okay, can you create the PR?

> did you include python implementation for reference?

</details>

### DIFF
--- a/.changeset/cancel-refused-speech.md
+++ b/.changeset/cancel-refused-speech.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+fix(voice): cancel a speech that `scheduleSpeech` refuses while scheduling is paused, so its speech task does not stay parked and block the next `drain()`/`pause()` and `AgentSession.close()`
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -4856,8 +4856,17 @@ export class AgentActivity implements RecognitionHooks {
   ): void {
     // when force=true, we allow tool responses to bypass scheduling pause
     // This allows for tool responses to be generated before the AgentActivity is finalized
-    if (this.schedulingPaused && !force) {
-      throw new SchedulingPausedError();
+    const schedulingPaused = this.schedulingPaused && !force;
+    if (schedulingPaused || this._mainTask?.done) {
+      this.logger.warn(
+        { speech_id: speechHandle.id, scheduling_paused: schedulingPaused },
+        'attempting to schedule a new SpeechHandle while speech scheduling is paused or stopped; the speech will be cancelled',
+      );
+      speechHandle.interrupt(true);
+      if (schedulingPaused) {
+        throw new SchedulingPausedError();
+      }
+      return;
     }
 
     // Monotonic time to avoid near 0 collisions
--- a/agents/src/voice/agent_activity.ts
+++ b/agents/src/voice/agent_activity.ts
@@ -2280,7 +2280,8 @@ export class AgentActivity implements RecognitionHooks {
       });
     };
 
-    const task = Task.from(wrappedFn, controller, name);
+    const taskController = controller ?? new AbortController();
+    const task = Task.from(wrappedFn, taskController, name);
     _setActivityTaskInfo(task, { speechHandle: ownedSpeechHandle, inlineTask });
 
     this.speechTasks.add(task);
@@ -2289,6 +2290,15 @@ export class AgentActivity implements RecognitionHooks {
     });
 
     if (ownedSpeechHandle) {
+      const interruptOwnedSpeech = () => ownedSpeechHandle.interrupt(true);
+      taskController.signal.addEventListener('abort', interruptOwnedSpeech, { once: true });
+      if (taskController.signal.aborted) {
+        interruptOwnedSpeech();
+      }
+      task.addDoneCallback(() => {
+        taskController.signal.removeEventListener('abort', interruptOwnedSpeech);
+      });
+
       ownedSpeechHandle._tasks.push(task);
       task.addDoneCallback(() => {
         if (ownedSpeechHandle._tasks.every((t) => t.done)) {
--- a/agents/src/voice/agent_activity_speech_lifecycle.test.ts
+++ b/agents/src/voice/agent_activity_speech_lifecycle.test.ts
@@ -2,11 +2,19 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 import { describe, expect, it, vi } from 'vitest';
+import { Future, type Task } from '../utils.js';
 import { AgentActivity, SchedulingPausedError } from './agent_activity.js';
 import { SpeechHandle } from './speech_handle.js';
 
 interface AgentActivitySpeechLifecycle {
   scheduleSpeech(speechHandle: SpeechHandle, priority: number, force?: boolean): void;
+  createSpeechTask(options: {
+    taskFn: (controller: AbortController) => Promise<void>;
+    controller?: AbortController;
+    ownedSpeechHandle?: SpeechHandle;
+    inlineTask?: boolean;
+    name?: string;
+  }): Task<void>;
 }
 
 const activityLifecycle = AgentActivity.prototype as unknown as AgentActivitySpeechLifecycle;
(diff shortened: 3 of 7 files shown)


davidzhao (approved):

dtran26:
Found a reproducible crash in the abort listener added by this PR. Opened #2435 with the reproduction and production stack trace.

In `AgentActivity.createSpeechTask()`, this callback returns a `SpeechHandle`:

```js
const interruptOwnedSpeech = () => ownedSpeechHandle.interrupt(true);
```

Because the handle is thenable, Node's `EventTarget.addCatch` implicitly invokes `waitForPlayout()`. When forced interruption runs inside the handle's owning function tool, that rejects with `SpeechHandleCircularWaitError`. Node rethrows it on the next tick, terminating the job process outside the interruption future's catch handler.

With SDK 1.8.0 / Node 22.23.2, the isolated reproduction exits with code 1. Changing only the callback to return nothing avoids the crash:

```js
const interruptOwnedSpeech = () => {
  ownedSpeechHandle.interrupt(true);
};
```

Could we make the listener explicitly return `void` and add a regression test for aborting a speech task from its owning tool's context? The callback fix is verified in the isolated reproduction; full-call validation is still pending.

PR #2414, opened 2026-09-03
feat(sarvam): add STTRealtime for realtime streaming STT

devbm7:
## Motivation

Sarvam's realtime speech-to-text API (`saaras:v3-realtime`) has been available in the Python SDK since livekit/agents#6562 (`STTRealtime`), but the Node.js plugin only exposes the legacy `STT` class (REST + a non-realtime WebSocket streaming mode). This PR closes that gap.

## Change

Adds `STTRealtime` and `RealtimeSpeechStream` to `plugins/sarvam/src/stt_realtime.ts`, exported alongside the existing `STT`/`SpeechStream` — the legacy class is untouched.

- Single-attempt WebSocket connection: `stream()` forces `connOptions.maxRetry` to `0`, since this endpoint bills per connection and must never silently reconnect on a socket failure.
- Full event mapping from Sarvam's realtime protocol to LiveKit `SpeechEventType`s: `session.begin`, `vad.speech_start`/`vad.speech_end`, `transcript.partial`/`transcript.final`, `session.end`, `config.updated`, `error`, `pong`.
- `endpointing: 'vad'` (default) buffers `transcript.final` until the matching `vad.speech_end` supplies the turn boundary, emitting `END_OF_SPEECH` + `FINAL_TRANSCRIPT` together; `endpointing: 'manual'` emits `FINAL_TRANSCRIPT` immediately and derives turn boundaries from `flush()`/`pushFrame()` instead of server VAD.
- Usage/billing: periodic local `RECOGNITION_USAGE` reporting (5s), reconciled against the server's authoritative `audio_duration_s` on `session.end` to avoid double-billing.
- Live `config.update` support for mid-session option changes (endpointing switches apply only at the next utterance boundary, gated by the server's ack).
- All four wire encodings supported: `linear16`, `linear32`, `mulaw`, `alaw`.

Also adds `plugins/sarvam/src/_utils.ts` (`PeriodicCollector`, matching the existing pattern already duplicated in `deepgram`/`xai`/`elevenlabs`), new types in `models.ts` (`STTRealtimeLanguages` correctly uses `or-IN` for Odia, unlike the legacy `od-IN`), and a README section documenting the no-reconnect/per-connection-billing behavior.

## Usage

```ts
import * as sarvam from '@livekit/agents-plugin-sarvam';

const stt = new sarvam.STTRealtime({
  language: 'en-IN',
  streamType: 'balanced',
  endpointing: 'vad',
});
```

## Validation

- `pnpm --filter @livekit/agents-plugin-sarvam build | lint | api:check` and `prettier --check` all pass; API report regenerated.
- `pnpm vitest run plugins/sarvam` — new `stt_realtime.test.ts` (vad-mode final-transcript gating, manual-mode immediate finals, no-reconnect guarantee, streaming-only contract) plus existing suite, all green.
- Manually verified end-to-end against the live Sarvam realtime API: all four wire encodings connect and complete a full session cleanly; a real speech sample (English) produces accurate partial/final transcripts with correctly gated VAD-mode turn boundaries and usage reconciliation.

## Notes

Companion Python PR: livekit/agents#6562.

### DIFF
--- a/.changeset/sarvam-realtime-stt.md
+++ b/.changeset/sarvam-realtime-stt.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents-plugin-sarvam': minor
+---
+
+Add `STTRealtime`, a Sarvam realtime speech-to-text plugin (`saaras:v3-realtime`) with server-side VAD, partial/final transcript gating, and per-connection usage reporting. Ported from `livekit/agents` (Python) PR #6562. Note: this stream never reconnects after a socket failure — Sarvam bills per connection, so `stream()` forces `connOptions.maxRetry` to `0`.
--- a/plugins/sarvam/README.md
+++ b/plugins/sarvam/README.md
@@ -77,6 +77,28 @@ const stt = new sarvam.STT({
 
 Set the `SARVAM_API_KEY` environment variable or pass `apiKey` directly.
 
+### STT (Realtime)
+
+```typescript
+import * as sarvam from '@livekit/agents-plugin-sarvam';
+
+const stt = new sarvam.STTRealtime({
+  language: 'en-IN',
+  streamType: 'balanced',
+  endpointing: 'vad',
+});
+```
+
+`STTRealtime` connects to Sarvam's realtime API (`saaras:v3-realtime`, not configurable), which
+streams partial transcripts and uses Sarvam's own VAD for turn detection by default. Set
+`endpointing: 'manual'` to delimit turns from your application instead — the plugin then emits
+`START_OF_SPEECH` on the first audio frame of a turn and `END_OF_SPEECH` when you flush the
+stream.
+
+Realtime streams don't reconnect after a socket failure, because Sarvam bills per connection —
+`stream()` forces `connOptions.maxRetry` to `0`. Create a new stream (or restart the session) if
+the connection drops.
+
 ## STT Models
 
 | Model                 | Endpoint                    | Languages   | Modes | Prompt |
--- a/plugins/sarvam/etc/agents-plugin-sarvam.api.md
+++ b/plugins/sarvam/etc/agents-plugin-sarvam.api.md
@@ -30,7 +30,37 @@ export class ChunkedStream extends tts.ChunkedStream {
     protected run(): Promise<void>;
 }
 
+// Warning: (ae-missing-release-tag) "RealtimeEncoding" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+//
+// @public
+export type RealtimeEncoding = 'linear16' | 'linear32' | 'mulaw' | 'alaw';
+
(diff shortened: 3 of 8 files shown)


devbm7:
Addressed all 5 findings from the automated reviews in c0658492:

1. **A-law sign inversion** (Devin + Codex) — fixed. While correcting it, discovered the rest of my hand-derived G.711 encoder also diverged from the reference algorithm, so replaced both `mulaw`/`alaw` encoders with a direct port of the standard ITU-T reference (`linear2ulaw`/`linear2alaw`) instead of patching further. Verified live against the API: all four encodings now produce identical, correct transcripts (previously `alaw` silently distorted audio).
2. **Hang on server-initiated close before `endInput()`** (Devin + Codex) — fixed. The audio pump and message loop now share an `AbortController` scoped to the connection attempt; whichever side settles first aborts the other instead of leaving it blocked on `this.input.next()`.
3. **Streams not removed from `STTRealtime`'s tracking set on natural completion** — fixed with an idempotent `#notifyClosed()` called from both `close()` and `run()`'s `finally`.
4. **Untagged PII in error logs** — fixed. Provider-supplied error text is now tagged `lk.pii.message` per this repo's convention, and no longer embedded in the thrown error's own `message`/`body`.
5. **Non-mono audio silently corrupted** (Codex) — fixed. `#processAudio` now rejects a non-mono `AudioFrame` with a clear error instead of feeding it through the mono-hardcoded framer.

Added regression tests for all 5 (`stt_realtime.test.ts`), including independent-reference round-trip tests for both G.711 encoders.

devbm7:
Addressed the second round of Devin findings in 09b7ea72:

1. **Turn completion precedes final transcript** — real bug, confirmed against `AudioRecognition`: `runEOUDetection` fires on `END_OF_SPEECH`, and in `turnDetectionMode: 'stt'` it uses whatever transcript has been processed so far. Since events from one stream are consumed in order, emitting `END_OF_SPEECH` before the buffered `FINAL_TRANSCRIPT` meant the turn could commit before its own final text arrived. Fixed by committing the buffered final first (`#tryCommitUtterance` already emits `FINAL_TRANSCRIPT` then `END_OF_SPEECH` in the right order) instead of emitting `END_OF_SPEECH` unconditionally.
2. **Live encoding updates corrupt audio** — confirmed: the server's decoder is fixed by the `encoding` query param at connect time, but `updateOptions` had no guard stopping a caller from switching `#opts.encoding` mid-stream, which would send a wire format the server never agreed to. Added `encoding` to the connection-only field list alongside `sampleRate`/`returnTimestamps`/`vadPrefixPaddingMs`.
3. **Final audio tail discarded** — confirmed: `#processAudio`'s loop just `break`s when `endInput()` is called without a preceding `flush()`, leaving up to one chunk (~50ms) of buffered PCM in the framer unsent. Now drains it before sending `end`.

Added 4 regression tests total for these two rounds' findings (13 passing now), and reverified live against the real API that audio/transcription still works end-to-end after the encoding-guard change.

PR #2426, opened 2026-09-05
perf(rime): stream sentence batches and reuse WebSocket connections

chasef07:
## Description

With Rime WebSocket `segment: 'never'`, the current plugin sends its provider flush only after all input text has arrived and closes the socket after each speech stream. This delays synthesis even when tokenized sentences are already available and repeats connection setup on later replies.

Add opt-in `flushSentences` and `reuseWebsocket` options. The first synthesizes available tokenizer sentences sequentially; the second reuses a fully drained connection. Existing HTTP and WebSocket behavior remains unchanged unless enabled.

## Changes Made

- Serialize each sentence's text/flush/done batch, preserving PCM, rebasing synthesis-local timestamps, and recording actual provider context IDs. Tokenizer buffering remains under the caller's control.
- Honor explicit SDK `stream.flush()` boundaries with separate final frames and metrics, including already-queued segments and whitespace-only input. Retain at most one real PCM sample for the final frame so short audio is delivered promptly.
- Exclusively lease sockets to streams, keep at most one completed idle connection for 30 seconds, discard interrupted/failed connections, invalidate reuse after connection-option changes, and release resources through `TTS.close()`.
- Add deterministic WebSocket regressions, document the options, regenerate the API report, and run Rime tests in CI. The regenerated report also removes a stale Arcana union member already absent from the current model source; this PR does not change model selection.

## Pre-Review Checklist

- [x] **Build passes**: Build, lint, typecheck, formatting, throws checks, and the affected/core tests pass locally.
- [x] **AI-generated code reviewed**: Independent standards and behavior reviews completed; identified segment metrics and provider trace-correlation issues fixed.
- [x] **Changes explained**: Public options and lifecycle behavior documented.
- [x] **Scope appropriate**: Rime plugin implementation, tests, documentation/API report, and its CI coverage.
- [ ] **Video demo**: Not applicable to this transport change; automated audio-stream assertions and direct provider validation are described below.

## Testing

- [x] Automated tests added/updated: 19 WebSocket tests cover early PCM through the real default `ttsNode`, baseline behavior, serialized batches, exact PCM conservation, timestamps, explicit and queued SDK segments, metrics, context tracing, reuse/concurrency, cancellation, option changes, timeout, idle expiry, and shutdown.
- [x] All tests pass: `pnpm test agents plugins/rime --maxWorkers 2 --silent` — 153 files passed; 2,497 tests passed and 6 skipped.
- [ ] `restaurant_agent.ts` / `realtime_agent.ts`: Not run; this change is isolated to the Rime plugin and its opt-in WebSocket path.

Local checks used Node 24 and the repository-pinned pnpm 11.13.1: full monorepo `pnpm build`, `pnpm lint`, `pnpm typecheck`, `pnpm format:check`, `pnpm throws:check`, and Rime `api:check`.

An earlier run alongside concurrent build/lint work had two timing failures in unchanged core tests. Both affected files passed on targeted rerun, and the complete final run above passed with two test workers.

The regression fixture was also run against unchanged upstream source: early audio, socket reuse, and default `ttsNode` early delivery failed before the change; the compatibility test passed. The explicit/queued SDK segment metrics failures were reproduced before their fixes.

## Additional Notes

The underlying transport change was exercised against the real Rime Coda `/ws3` endpoint in ten synthetic-text scenarios, including English/Spanish, cancellation/recovery, connection reuse, and shutdown. No credentials, patient data, or private application code are included in this PR.

With a sentence-ready tokenizer and a deliberately paced two-second text-input schedule, English first PCM arrived at 2.27–2.36 seconds on the original path versus 0.31–0.65 seconds on the optimized path. The second optimized reply opened no new connection. These controlled measurements demonstrate removal of client buffering; they are not a production latency distribution or a claim about the default tokenizer's latency.

Rime can coalesce overlapping flushes, so this implementation waits for each matching `done` before submitting the next batch. See the [segmentation contract](https://docs.rime.ai/docs/websockets-segment) and [Coda JSON protocol](https://docs.rime.ai/api-reference/coda/websockets-json).

Live word timestamps sometimes exceeded returned PCM duration on both original and optimized paths. This PR preserves/rebases provider timing; it does not attempt to correct provider alignment accuracy. Sentence prosody and full SIP-call behavior still require listening/call validation.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,7 @@ on:
       - 'agents/**'
       - 'package.json'
       - 'plugins/test/**'
+      - 'plugins/rime/**'
       - 'pnpm-lock.yaml'
       - 'pnpm-workspace.yaml'
       - 'turbo.json'
@@ -48,6 +49,8 @@ jobs:
         run: pnpm build
       - name: Test agents
         run: pnpm test agents --silent
+      - name: Test Rime plugin
+        run: pnpm test plugins/rime --silent
       - name: Test LiveKit inference
         env:
           LIVEKIT_INFERENCE_TESTS: '1'
--- a/plugins/rime/README.md
+++ b/plugins/rime/README.md
@@ -3,6 +3,7 @@ SPDX-FileCopyrightText: 2024 LiveKit, Inc.
 
 SPDX-License-Identifier: Apache-2.0
 -->
+
 # Rime plugin for LiveKit Agents
 
 The Agents Framework is designed for building realtime, programmable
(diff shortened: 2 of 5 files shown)


chasef07:
@naszzz can you look and test this aap and if you want it to be default?

naszzz (approved):

PR #2481, opened 2026-09-11
fix(agents): refuse to start a DuplexModel under a text simulation

u9g:
## Problem

A duplex model such as `GPTLiveModel` hears and speaks only audio, and `DuplexRealtimeAdapter` resolves each reply from the audio the model produces. A text simulation (`lk agent simulate` without `audio`) disables audio I/O, so the first reply never opens and every scenario fails on turn one with "the model did not start speaking when asked" after a 10 second timeout, with nothing pointing at the cause.

## Fix

`AgentActivity` throws at construction when the session is a text simulation and the resolved llm is a `DuplexRealtimeAdapter`:

```
a DuplexModel speaks only through audio, so it cannot run under a text simulation; run `lk agent simulate audio` instead
```

No new capability flag: every duplex model is audio-only by definition, so the type is the signal.

## Tests

`duplex_adapter.test.ts` gains a test that `AgentSession.start` rejects with that message when `_textOnly` is true. Includes a patch changeset for `@livekit/agents`.

Python counterpart: https://github.com/livekit/agents/pull/7236.

### DIFF
--- a/.changeset/duplex-text-simulation-error.md
+++ b/.changeset/duplex-text-simulation-error.md
@@ -1,0 +1,5 @@
+---
+'@livekit/agents': patch
+---
+
+Fail fast with a clear error when a DuplexModel such as GPT-Live is started under a text simulation, which has no audio, instead of timing out on the first reply.
--- a/agents/src/llm/duplex_adapter.test.ts
+++ b/agents/src/llm/duplex_adapter.test.ts
@@ -817,6 +817,17 @@ describe('duplex requested replies', () => {
 });
 
 describe('duplex model integration', () => {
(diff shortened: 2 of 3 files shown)


u9g:
Before and after under `lk agent simulate` (text mode) with a GPT-Live agent, recorded against the Python counterpart of this change since the behavior is identical: the released SDK times out on turn one with "the model did not start speaking when asked" and no hint why; the fix refuses at session start and names `lk agent simulate audio` as the mode to use.

https://github.com/user-attachments/assets/40135444-610f-414d-be5f-70d2ba894e7f

chenghao-mou (approved):

PR #2485, opened 2026-09-11
ipc: add room_id to the job logger and bind it as the process logger

andrewnitu:


agents/src/ipc/job_proc_lazy_main.ts
```diff
@@ -285,7 +286,11 @@ const startJob = (
             throw new Error('job task already running');
           }
 
-          logger = logger.child({ jobID: msg.value.runningJob.job.id });
+          logger = logger.child({
+            jobID: msg.value.runningJob.job.id,
+            room_id: msg.value.runningJob.job.room?.sid,
+          });
+          setLoggerState(logger, loggerOptions()!);
```
andrewnitu, line 293: this line is needed to update the global logger, i.e. whenever log() is called.

theomonnom (approved):

PR #2500, opened 2026-09-15
feat(telemetry): one agent_turn span per speech handle

davidzhao:
Port of livekit/agents#7143. Stacked on #2499.

## Description

A reply that calls a tool runs two generations (LLM steps) in two tasks. They were two `agent_turn` spans linked only by `lk.parent_generation_id`, so one response rendered as two turns. A speech handle is now exactly one `agent_turn`.

- The speech handle owns the span. The first reply task (pipeline, realtime, or `say`) opens `agent_turn` under `agent_session`; the follow-up generation after a tool call continues the open span instead of opening a second one. It ends with the speech in `SpeechHandle._markDone`, recording the speech's error (redaction-aware) and the `gen_ai.invoke_agent.duration` metric for the whole turn.
- Each generation is a `generation` event on the span with `lk.generation_id` and `lk.parent_generation_id`; the span's `lk.generation_id` names the latest generation and the new `lk.generation_count` how many there were. `lk.speech_id` is set at creation.
- Every step's `llm_node`, `function_tool`, `tts_node`, `realtime_inference` and `agent_speaking` nest under the one turn.
- A preemptive generation discarded for a successor answering the same user turn hands its open `agent_turn` to that successor (`preemptive_generation_discarded` event; `lk.speech_id` follows the speech that answered). An attempt cancelled with no successor still ends as its own turn. The queue-wait and interruption helpers tolerate a span already ended with the speech.

```
agent_turn  (lk.speech_id, lk.generation_count=2; events: generation ×2)
├─ llm_node          generation 1: the tool call
├─ function_tool
├─ llm_node          generation 2: the reply
├─ tts_node
└─ agent_speaking
```

## Changes Made
- `voice/speech_handle.ts`: span ownership, `_generationId` / `_parentGenerationId`, `_takeAgentTurn` / `_continueAgentTurn`, end in `_markDone`.
- `voice/agent_activity.ts`: `withAgentTurn` and `continueDiscardedTurn` wrap the pipeline, realtime and `say` tasks; per-step span creation removed.
- `telemetry/otel_metrics.ts`: `recordInvokeAgentDuration` (`gen_ai.invoke_agent.duration`, unit `s`, job attribution); `telemetry/trace_types.ts`: `lk.generation_count`, the metric name.

## Adaptations from the Python source
- `say` previously had no `agent_turn` span in JS; it now gets one like Python's TTS task.
- `lk.generation_id` / `lk.parent_generation_id` were never stamped on `agent_turn` in JS before (the constants existed unused); they are now, as `<speech_id>_<step>`.

## Testing
- New `voice/agent_turn_span.test.ts` (6 tests): a tool-calling reply is one `agent_turn` with two `generation` events and both `llm_node`s, the tool, `tts_node` and `agent_speaking` inside it; a plain reply is one generation; the discarded preemptive hand-off; an LLM failure fails the turn; the duration metric when sampled out.
- Full `agents` suite green; build, typecheck, lint, API report updated.

**Review follow-ups**
- A realtime tool reply, which the framework runs on a new speech handle (Python runs it on the same one as its next step), now continues the tool call's `agent_turn` instead of opening a second one: `continueToolReplyTurn` hands the open span over, the turn keeps the tool call's speech id, and the reply's generation is numbered after and parented to the tool call's (`lk.generation_id` / `lk.parent_generation_id` continue the original handle's numbering). The exported trace matches Python's: one turn, two generations, parent link intact.
- A speech whose owned task fails now ends its turn with the error and reports it through `exception()`: the first non-cancellation rejection of an owned task is stored on the handle before it is marked done. The pipeline also stores a genuine LLM node failure the way Python's `_on_llm_task_done` does, since nothing awaits that task's rejection.
- `lk.generation_count` counts the generation events on the turn, carried across handoffs, instead of the current handle's own step count: a turn continued from a discarded preemptive attempt or a tool call reports every generation it shows. Python's `_continue_agent_turn` needs the same carry.


agents/src/voice/agent_activity.ts
```diff
@@ -4226,7 +4348,6 @@ export class AgentActivity implements RecognitionHooks {
     inferenceSpan: Span;
   }): Promise<void> {
     const { speechHandle } = stateLease;
-    speechHandle._agentTurnContext = otelContext.active();
 
     span.setAttribute(traceTypes.ATTR_SPEECH_ID, speechHandle.id);
```
chenghao-mou, line 4352: should we use the parent speech (or root parent) id here?

agents/src/voice/speech_handle.ts
```diff
@@ -538,6 +594,91 @@ export class SpeechHandle {
...
+    if (span === undefined) return;
+    // the duration metric does not depend on the span being sampled in
+    if (this._agentTurnStartedAt !== undefined && this._agentTurnAgentName !== undefined) {
+      recordInvokeAgentDuration(
+        (performance.now() - this._agentTurnStartedAt) / 1000,
+        this._agentTurnAgentName,
+      );
+    }
+    if (!span.isRecording()) return;
+    if (error instanceof Error) {
```
chenghao-mou, line 676: nit: we should normalize error first in case it is a string:

```
const failure = error instanceof Error ? error : new Error(String(error));
```

chenghao-mou (approved):

PR #2510, opened 2026-09-16
fix(workflows): forward Twilio CallToken during warm transfers

piyush-gambhir:
A warm transfer creates a new outbound call, so human recipients normally see the business number rather than the inbound customer's number. This adds per-call caller-ID preservation through Twilio's CallToken, while keeping the business number as the default and fallback. It lets applications use the SDK's consultation and cleanup flow instead of replacing Twilio call creation to preserve customer identity.

## API and behavior

- `twilioFromNumber` stays the agent/business Twilio number or verified caller ID.
- Optional `originalCallerNumber` carries the original inbound webhook's `From`.
- A nonempty `twilioCallToken` opts into preservation and requires the matching original caller number. No separate enable flag or fallback-number option is needed.
- Without a token (including an empty token), dial from the business number even if the original caller number is supplied.
- If the preservation request is explicitly rejected with **HTTP 400 / Twilio error 21210**, retry once from the business number without CallToken. This error means the original caller ID was not accepted; it was also the response observed for the invalid-token live control.
- Do not retry transport timeouts, unrelated rejection codes, authentication/rate-limit errors, or failures after a call was created. A failed fallback propagates normally. This avoids duplicate dialing after an ambiguous outcome.

Keep the number/token pair in server-side state keyed by the incoming `CallSid`, obtained from a validated Twilio webhook. Tokens are not included in connector requests, TwiML, prompts, or participant attributes. The token is not permission to choose arbitrary caller IDs.

This can help recipients identify the customer and support caller-ID-based lookup/callback records; the PR does not implement CRM matching. Caller-ID presentation remains subject to Twilio acceptance and downstream carrier behavior. The existing Calls API is used, with no conference or LiveKit server API change.

## Twilio references

- [Incoming TwiML request parameters](https://www.twilio.com/docs/voice/twiml#request-parameters): `From`, `CallSid`, and `CallToken`.
- [Calls API: create a call](https://www.twilio.com/docs/voice/api/call-resource#create-a-call): the endpoint used for the transfer leg.
- [Trusted calling / forwarding](https://www.twilio.com/docs/voice/trusted-calling-with-shakenstir): matching the inbound token and caller number.
- [Error 21210: unverified From](https://www.twilio.com/docs/api/errors/21210): the narrowly scoped fallback condition.
- [Conference CallToken changelog](https://www.twilio.com/en-us/changelog/calltoken-support-for-conference--participants-api) and [create a conference participant](https://www.twilio.com/docs/voice/api/conference-participant-resource#create-a-participant): related support for conference flows; this PR does not require a conference.

## Prerequisite

This PR targets `main` and depends on #2402, which introduces the JavaScript connector-transfer helper. The branch includes its two prerequisite commits with original authorship and merges current `main`. Once #2402 merges, the branch can be rebased to leave only this feature. The public API report and examples are updated.

Python counterpart: https://github.com/livekit/agents/pull/7309.

## Validation

- 40 focused workflow tests passed, covering token forwarding/omission, original-caller validation, single fallback, failed fallback, unrelated errors, network ambiguity, and cancellation of the fallback call.
- Typecheck, SDK build, API Extractor check, formatting, and repository lint passed (existing lint warnings remain).
- Error responses expose status/code without logging provider response bodies that could contain caller data or tokens.

## Live QA — 2026-09-16, current fallback behavior

Ran both local SDK workers against UAT LiveKit Cloud and real Twilio calls. Two temporary phone numbers were used: a customer/receiver in a temporary subaccount and a business number in the SDK account. This makes the customer's number unverified in the dialing account. Each SDK passed all four cases below (8/8 total):

| Input | Observed result |
| --- | --- |
| Valid token + original caller | Transfer completed; receiver saw the customer caller ID. |
| No token, no original caller | Transfer completed; receiver saw the business caller ID. |
| Original caller supplied, no token | Transfer completed; receiver saw the business caller ID. |
| Invalid token + original caller | Preservation rejected; fallback completed with the business caller ID. |

Every successful case had one receiver webhook and published audio tracks for the customer and transferred human in the caller room. The invalid-token Python attempt also recorded the SDK's explicit caller-ID fallback warning. Unit tests verify both request bodies, the exact single-retry limit, and failure exclusions.

One Python harness attempt raced room initialization before dialing. After adding an explicit room connection in the harness, that case passed using the same temporary numbers; interrupted harness attempts are excluded from the final matrix. No SDK behavior was changed to work around that harness issue.

Independent API reads confirmed both purchased numbers released (404), the temporary subaccount closed, 38 associated call legs checked with zero active, and zero QA rooms remaining. Local QA workers and the webhook tunnel were stopped.

Earlier live QA on the pre-fallback revision also covered supervisor decline, no answer, busy/rejected calls, and caller hangup during ringing in both SDKs. Those lifecycle cases are historical validation, not a fresh full-matrix run of this revision. Current regression tests cover cancellation of an unanswered fallback call.

Limits: automated endpoints verify caller-ID signaling and published audio tracks, not physical handset display, subjective two-way audio quality, or every carrier. Telnyx is not included in these results.


piyush-gambhir:
Superseded by livekit/agents-js#2540, which carries the same change as linear commits on top of main (the two commits from livekit/agents-js#2402 with original authorship, then the CallToken change).

PR #2539, opened 2026-09-19
fix(openai): listen for ws 'error' while the Realtime socket connects (unhandled error crashes the worker)

yaniv-peretz:
## Summary

`RealtimeSession.#connect` (plugins/openai `realtime_model.ts`) wraps the OpenAI Realtime `WebSocket` in a promise and listens for `open` and `close` only. `ws` emits **`error` before `close`** whenever the handshake fails — refused connection, TLS failure, a 4xx on the upgrade — and also when the plugin's own connect timeout calls `ws.close()` on a socket that is still `CONNECTING` (`ws` aborts the handshake with *"WebSocket was closed before the connection was established"*).

With no `error` listener, Node's `EventEmitter` throws that event as an **unhandled `error` event**. It fires on `process.nextTick`, so it escapes the connect promise entirely and becomes an uncaught exception: the **whole worker process exits**, taking every other live session on that worker down with it, instead of just failing the one session whose connect failed.

We hit this in a deployment where egress to `api.openai.com` was intermittently blocked: the first session to time out killed the worker. We have carried this listener as a `pnpm patch` since July; this PR upstreams it.

## Proof chain (from source)

1. `realtime_model.ts` `#connect`: on timeout → `ws.close()`; listeners registered: `open`, `close`.
2. `ws@8` `WebSocket.close()`: `readyState === CONNECTING` → `abortHandshake(this, this._req, 'WebSocket was closed before the connection was established')`.
3. `abortHandshake` → `process.nextTick(emitErrorAndClose, …)` → `websocket.emit('error', err)` **then** `emitClose()`.
4. No `'error'` listener → `EventEmitter` throws (`ERR_UNHANDLED_ERROR`) on `nextTick` → uncaught → process exit.

## Fix

- Add `ws.once('error', …)` to the connect promise; it rejects with the socket error (a handshake failure now surfaces as the real error instead of a generic *"connection closed"*).
- Mark the timeout branch as settled (`waiting = false`) **before** it calls `ws.close()`, so the timeout stays the reported cause and the `error` the close triggers is swallowed rather than double-rejecting.

14 lines, no behaviour change on the happy path.

## Test

New `RealtimeSession connect failures` case: a `node:net` server accepts the TCP connection but never answers the HTTP upgrade, so the socket sits in `CONNECTING` until a 50 ms `connOptions.timeoutMs` fires. `generateReply()` must reject with *"Realtime session closed"* and the file must finish with no unhandled error.

- **With the fix:** 53/53 pass.
- **Without the fix** (reverting only `realtime_model.ts`): vitest reports
  ```
  ⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯
  Error: WebSocket was closed before the connection was established
  ```
  and exits non-zero — the exact crash, in CI form.

Changeset included (`@livekit/agents-plugin-openai`: patch).


admin-bot-pulse-dev:
The gist of this PR is that a connection is blocked (for example local WIFI in my case) but could be any other TLS failure ext in prod.

there is no emmited path -> this PR should reject the connect (before the timeout) and present the log.
I specifically patch for this in my Dockerfile - though not critical but it helps for debugging.

```ts
      ws.once('error', (err) => {
        if (!waiting) return;
        waiting = false;
        clearTimeout(timeout);
        reject(err instanceof Error ? err : new Error(String(err)));
      });
```

yaniv-peretz:
Two notes for reviewers after checking the repo for prior art:

- There is no existing PR or issue for this crash; the closest is #2379, which introduced `waitForWebSocketOpen` in agents core (with `error` / `unexpected-response` handling) and migrated Deepgram, xAI and MiniMax to it — the OpenAI Realtime connect path was not migrated and still has only `open`/`close` listeners. I kept this PR minimal (add the listener in place, keep the existing timeout semantics), but I'm happy to switch `#connect` to `waitForWebSocketOpen` + the timeout if you prefer that shape.
- #2532 moves `realtime_model.ts` into agents core; the hunk is 14 lines in `#connect` and applies cleanly on top of the move, so I can rebase onto whichever lands first.

PR #2541, opened 2026-09-20
feat(telemetry): nest event loop stalls under the blocked span and sample their stacks

davidzhao:
Stacked on #2501. Not a port: the Python monitor gets these from `sys._current_frames()` and the blocked task's context, which Node has no equivalent for; this layer reaches the same observable result with Node mechanisms.

## Description

A cloud export from the JS build showed every `event_loop_blocked` span parented to `agent_session` with no stack, which made the stalls hard to act on. Two changes, both matching what the Python monitor reports:

**A stall nests under the span that was blocked.** Every framework span is created and ended on the main thread, so a span created before a stall and still open (or ended when the blocking call returned) was current during it. `BlockedSpanTracker`, a span processor installed wherever the framework installs its own processors, keeps creation and end times; the monitor parents a stall under the innermost such span, then the session root, then the job root. A slow tool lands under `function_tool`, a slow RPC under `rpc_handler`, a slow hook under `on_user_turn_completed`. Creation time is used rather than the span's start time, so a back-dated `eou_wait` cannot claim a stall that predates it.

**`lk.blocking.stack` is populated.** The watchdog worker thread attaches an inspector session to the main thread and, when the heartbeat is a warn threshold late, posts `Debugger.pause`: V8 honours it inside a spinning script, hands back full symbolized frames (optimized code included), and resumes about a millisecond later. A second sample at 10x the threshold, like Python's late sample. The format matches the Python monitor: `# loop thread sampled Nms into the stall` then the innermost 20 frames; Node internals dropped except innermost; frames outer than the framework's job runner cut. The warn log names the innermost frame of the agent's own code.

Sampling is **adaptive by default**: off until the process reports its first code-caused stall, then on for that process, so a healthy process never enables the debugger domain. `LIVEKIT_AGENTS_LOOP_BLOCK_STACKS=always` turns it on from the start, `never` off. Enabling the debugger domain costs one loop gap of about 90 ms with the framework loaded (script enumeration), once per process; in adaptive mode that gap is itself reported as a stall with the note that the sampler was starting, and in `always` mode it happens at monitor start, before any job work. Not enabled while an inspector is attached to the process. The worker process and an idle child sample too (third commit): their stalls only reach the log, and a log with a stack names the blocking code just as a span does, so no stall is reported with a note saying its process does not sample.

**Rejected alternative, not what ships:** V8's sampling profiler was tried first. On Node 24 only the first profile of a process names code optimized before it started, reading its samples requires restarting the profile, and each restart blocked the loop for 25 to 40 ms with the framework loaded. Those figures describe the profiler, not the `Debugger.pause` mechanism above.

**Measured overhead** of the shipped mechanism with the framework loaded (ten interleaved rounds, medians): with the debugger domain enabled a tight compute loop is 1.1% slower and a mixed JSON/regex/promise workload 0.6% slower; idle CPU with the 20 ms heartbeat is unchanged (0.33% vs 0.31% of a core); a pause sample costs about 1 ms and only while a stall is under way. In adaptive mode none of this is paid until the process's first code-caused stall.

**Limits.** A native call that does not check for interrupts (sync file or child-process calls, native addons) keeps the pause waiting until it returns; the sample then shows the caller still in the frame that made the call, which is what the Python watchdog sees once a native call releases the GIL. `lk.blocking.task` stays unset: Node has no task names.

## Verified live

Hotel receptionist example with a `test` RPC that spins for 600 ms, called twice from a client: both `rpc_handler` spans under `agent_session`, both stalls under their `rpc_handler`, the first without a stack (adaptive), the second sampled 121 ms into the stall naming the handler, the SDK's `invokeRpcHandler`, and the interceptor's `interceptIncoming`.

## Testing
- `telemetry/loop_monitor_stacks.test.ts` (13 tests): parenting under an open span, fallbacks, tracker retention, adaptive / always / never, the formatted stack from a real blocked call, restart bounds, env parsing. The existing 28 monitor tests unchanged.
- Full `agents` suite green; build, typecheck, lint, API report updated.

**API report noise (second commit).** Exports in `@livekit/agents` are not release-tagged, so API Extractor wrote an `ae-missing-release-tag` warning for every one of them into `agents.api.md` (854 by this layer), burying the real changes in each review. `agents/api-extractor.json` now silences that one message for this package and the report is regenerated: the diff is only the removed warning lines. Other warning kinds (`ae-forgotten-export`, `ae-unresolved-link`, ...) are untouched.

**Review follow-ups (fourth commit)**
- Parenting: when two in-flight operations of the same kind qualify for a stall (two `function_tool`s of one turn, two `rpc_handler`s), the tracker no longer picks the newer one, which timing cannot justify: the stall lands on their nearest common ancestor among the open spans, or on the session/job root when there is none. Operations of different kinds (a `user_turn` open while an `rpc_handler` runs) still resolve to the newest. An exact answer for sampled stalls is possible as a follow-up: evaluate the active span on the paused main thread during the sample, which reads the blocked operation's async context the way Python reads the blocked task's context variables.
- Sample offsets are measured from the stall's start as the report dates it; the header used to overstate by one heartbeat interval.
- `stacks: always|adaptive` with `watchdog: false` is normalized to `never` (the sampler runs on the watchdog thread), documented, so such a configuration does not note on every stall that sampling starts after the first one.
- An inspector attached after sampling started stops it: the heartbeat checks the inspector URL each tick and has the watchdog disable the debugger domain and disconnect, so a debugger's breakpoints are never resumed by the sampler. A `debugger;` statement in user code is still resumed while no inspector is attached: it only pauses at all because the sampler enabled the domain. Later stalls note that an inspector is attached. Tested against a real `inspector.open()`.


agents/src/telemetry/blocked_span_tracker.ts
```diff
@@ -0,0 +1,159 @@
...
+      }
+    }
+    const leaves = [...candidates.entries()].filter(([id]) => !hasChild.has(id));
+    // ties (same millisecond) go to the later-created span, which the maps yield last
+    let newest = leaves[0]!;
+    for (const leaf of leaves) if (leaf[1].createdAt >= newest[1].createdAt) newest = leaf;
+    const sameKind = leaves.filter(
+      ([, entry]) => spanName(entry.span) === spanName(newest[1].span),
+    );
+    if (sameKind.length <= 1) return newest[1].span;
```
chenghao-mou, line 104: from Codex:

> Differently named operations have the same ambiguity. A tool can await I/O, a newer RPC can start and await I/O, then the tool can resume and block. Both spans remain open, but this selects the idle RPC.

Should we use the common ancestor across all overlapping leaves, regardless of span name?

agents/src/telemetry/blocked_span_tracker.ts
```diff
@@ -0,0 +1,159 @@
...
+    exclude: ReadonlySet<string>,
+    slack = 2,
+  ): Span | undefined {
+    const opened = startedAt + slack;
+    const closed = endedAt - slack;
+    const candidates = new Map<string, { span: Span; createdAt: number }>();
+    const consider = (entry: { span: Span; createdAt: number }) => {
+      if (entry.createdAt > opened) return;
+      if (exclude.has(spanName(entry.span))) return;
+      candidates.set(entry.span.spanContext().spanId, entry);
```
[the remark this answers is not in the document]
chenghao-mou: also from Codex:
> Can we filter candidates by the session/job context’s trace ID before selecting a parent? This processor receives spans from the entire provider. A newer background HTTP span from another trace can win here, moving `event_loop_blocked` out of the agent’s trace.

u9g (approved):
I ran it locally, it looks to work well, except for a problem I found that. Imade. pr for here: https://github.com/livekit/agents-js/pull/2575

agents/src/telemetry/loop_stack_sampler.ts
```diff
@@ -0,0 +1,282 @@
...
+    if (event.params.url) scripts.set(event.params.scriptId, event.params.url);
+  });
+  s.on('Debugger.paused', (event) => {
+    if (pausing) {
+      pausing.resolve(event.params);
+    } else {
+      // a debugger statement in user code, which only pauses because this session enabled
+      // the domain: resume, as it would have been a no-op without it. Breakpoints of a real
+      // debugger never reach here: the monitor disables sampling as soon as one attaches
+      s.post('Debugger.resume', () => undefined);
```
chenghao-mou, line 144: from Codex:

> The monitor detects attachment **on the next heartbeat**, not at the instant of attachment. A debugger can attach and pause before that heartbeat runs. This handler then resumes its pause.

PR #2546, opened 2026-09-21
fix(inference): empty interims do not keep a finalizing STT stream open

davidzhao:
## What

An `interim_transcript` with no text no longer keeps a finalizing LiveKit Inference STT stream open.

## Why

After `endInput()` the client sends `session.finalize` and waits for transcript activity to stop before it sends `session.close`: 3 s after the last final, 30 s otherwise. Every transcript event re-armed that wait.

For `xai/stt-1` the gateway acknowledges the finalize with `session.finalized` and then sends an **empty** `interim_transcript` about once a second, always with the same `start`/`duration`, for as long as the socket is open, and never closes the session itself. Each one re-armed the wait, so the stream never ended and the integration test timed out at 120 s (twice, with the retry: the 264 s in the CI log) while `deepgram/nova-3`, `cartesia/ink-whisper` and `assemblyai/universal-streaming` finish in 55–80 s. Recorded from the gateway with a standalone websocket client, at session rates 16000 and 24000:

```
[ 52258ms] >> session.finalize (after 50.0 s of audio + 2 s silence)
[ 52299ms] << session.finalized
[ 52448ms] << interim_transcript (0 chars, start=50.753s dur=1.21s) ""
[ 53445ms] << interim_transcript (0 chars, start=50.753s dur=1.21s) ""
... one per second, 25 in the 25 s the client kept listening; the transcript itself was complete and correct
```

The other three providers send nothing after their last final. The server-side behaviour is being reported separately with that repro.

Python's inference client is not affected only because it never finishes a stream on its own: it has no finalization wait and relies on the caller closing the stream, and its STT test breaks out of the event loop after the first final.

## The fix

In the socket message handler, an interim with no text does not call `scheduleFinalizationTimeout()`. Finals and preflight transcripts still do, so a provider that flushes a late final is not cut off, and nothing depends on `session.finalized`, which Deepgram and AssemblyAI never send. Empty interims were already dropped from what the stream emits.

## Tests

- `does not let empty interims after finalize hold the stream open` (`inference/stt.test.ts`): replays the recorded sequence over a local socket on real timers, and checks the stream ends within the 3 s window with `session.close` sent while the empty interims keep arriving. It hangs to its timeout against the previous code.
- Live: `LIVEKIT_INFERENCE_TESTS=1 vitest run agents/src/inference/stt.test.ts -t xai` passes both streaming cases in 59.7 s, in line with the other models.


shawnfeldman (approved):
Approving — reviewed the change end to end, no Critical or Important issues.

**Verified**

- **Field path is right.** `transcript` is a top-level string on `interim_transcript`, and the gateway's own schema (`agent-gateway/docs/STT_TTS_WEBSOCKET_API.md`) documents it as `May be empty (silence interims)` — so an empty interim is an expected message shape, not a malformed one, and `(json as { transcript?: string }).transcript` reads the correct field.
- **Consistent with the emit side.** `processTranscript` already drops empty non-final transcripts (`if (!text && eventType !== SpeechEventType.FINAL_TRANSCRIPT) return`). An empty interim was already a non-event for consumers; the shutdown path now agrees with that, rather than introducing a new notion of "empty".
- **Blast radius is the shutdown path only.** `scheduleFinalizationTimeout()` returns early unless `inputEnded`, so nothing changes for a live stream.
- **A silent stream is still bounded.** The timer has a second call site right after `inputEnded = true` / `sendSessionFinalize()` (stt.ts:987), so a stream that emits *only* empty interims after finalize still gets its 30 s `FINALIZATION_TIMEOUT_MS` — this is strictly better than before, not a swap of one hang for another.
- **Late finals still re-arm.** Finals and preflight transcripts are untouched, and the 3 s / 30 s constants are unchanged.
- **Test matches the described failure.** It replays the recorded sequence on real timers and asserts `session.close` was sent, `emptyInterims >= 3` (they really were arriving through the wait) and `endedAfter` in [2.5 s, 5 s) — so it pins the window rather than just asserting termination. It would hang against the previous code.

Devin Review reported no issues; no other review comments on the PR. All checks green.

**Non-blocking**

- `Boolean(transcriptText)` treats a whitespace-only transcript as non-empty, so it still re-arms. That's the conservative direction and matches `processTranscript`'s own falsy check, so it's consistent — just noting it's whitespace-sensitive if a provider ever emits `" "` padding.
- The gateway-side behaviour is the real bug: emitting empty interims is within spec, but `xai/stt-1` never terminating the session after `session.finalized` is not. Worth making sure that report lands against agent-gateway — this fix correctly protects the client either way.

chenghao-mou (approved):

PR #2551, opened 2026-09-21
fix(google): keep declared property order in Live API tool schemas

### LINKED ISSUE #2529: google plugin: tool schemas never send Gemini propertyOrdering, so arguments generate alphabetically
### What happens

`@livekit/agents-plugin-google` builds each tool's Gemini `Schema` from its JSON schema in `dist/utils.js`, copying a fixed set of keys (`type`, `description`, `required`, `properties`, `items`, `enum`, …). It never emits `propertyOrdering`.

The Gemini API orders an object's properties alphabetically unless the schema carries `propertyOrdering`, so a tool's arguments are generated in alphabetical order rather than the order the schema declares.

### Why it matters

Gemini writes a tool call's arguments as JSON, one field at a time, so a field can only be conditioned on the fields already written. When a tool has a field that is meant to be worked out first and a field that depends on it, alphabetical order can invert them.

A minimal example — a tool declared as:

```ts
{
  reasoning: z.string(),   // meant to be written first
  answer: z.enum(['a', 'b']),
}
```

is generated as `answer` first, then `reasoning`, because `answer` sorts before `reasoning`. The field intended to inform the decision is produced after it.

### Expected

Tool declarations carry `propertyOrdering` taken from each object schema's own property order, nested objects included, so Gemini generates the fields in the order the tool declares them.

### Notes

- Google's docs: [property ordering](https://ai.google.dev/gemini-api/docs/structured-output#property-ordering) — "properties listed in `propertyOrdering` are generated first, in the specified order".
- `@google/genai` passes the field through on the Live setup path (the `Schema` type carries `propertyOrdering`, and unknown keys survive `processJsonSchema`), so no SDK change is needed.
- We currently patch the plugin locally: set `propertyOrdering` from the property key order of every object schema that has `properties`, applied recursively. Confirmed present in the declarations sent to a Gemini 3.8 Live session.
- Happy to open a PR for that if it's wanted.

### Version

`@livekit/agents` 1.9.0, `@livekit/agents-plugin-google` 1.9.0, Node 24, Gemini Live (realtime).

JustinScitech:
Fixes https://github.com/livekit/agents-js/issues/2529

## What
Gemini generates a tool call's arguments in alphabetical order unless the schema carries `propertyOrdering`. A Live API tool declared as `{ reasoning, answer }` was generated `answer` first.

This sets `propertyOrdering` from each object schema's own property order (nested objects included) when building the Live API `Schema`.

## Tests

| Test | Command | Result |
|---|---|---|
| `uses legacy parameters on the Live API` (updated: expects `propertyOrdering: ['fields']`) | `pnpm vitest run plugins/google/src/utils.test.ts` | ✅ pass |
| `orders Live API parameters as declared, nested objects included` (new: `['reasoning', 'answer', 'details']`, nested `['zeta', 'alpha']`) | `pnpm vitest run plugins/google/src/utils.test.ts` | ✅ pass |
| Full `utils.test.ts` suite (6 tests) | `pnpm vitest run plugins/google/src/utils.test.ts` | ✅ 6/6 pass |
| Lint / format | `pnpm exec eslint plugins/google/src/utils.ts plugins/google/src/utils.test.ts` / `pnpm exec prettier --check ...` | ✅ clean |


chenghao-mou (approved):
lgtm

Issue #2508, opened 2026-09-16
Unhandled promise rejections from `.finally(...)` on task results, and a lint rule that would catch them


Rehansanjay:
## The pattern

`agent_activity.ts:1616` (also `:2822`, `:3983`) attaches cleanup to a task's result:

```ts
task.result.finally(() => this.onPipelineReplyDone(stateLease));
```

`.finally()` returns a **new** promise that re-rejects with the task's error. The task's own rejection is
handled elsewhere, but nothing handles the derived promise, so a failing speech task produces an unhandled
rejection.

Reproduced with `Task` itself:

```ts
const task = Task.from(async () => {
  throw new Error('speech task failed');
});

task.result.finally(() => {});          // the pattern above
await task.result.catch(() => undefined); // the task's own failure is handled

// -> 'unhandledRejection' fires for the promise returned by .finally()
```

Adding `.catch()` to the derived promise silences it; nothing else changes.

## Why it matters

- In a job process, the only handler is `ipc/job_proc_lazy_main.ts:269`, which logs at `debug`. At normal log
  levels the error is invisible.
- `worker.ts:720`, `:745` and `:963` use the same pattern in the **worker** process, where no handler is
  installed, so Node's default behaviour applies. Those callbacks can also throw on their own
  (they throw "task ... not found in tasks"), which rejects the derived promise even when the
  task succeeded.

The codebase already guards the same shape elsewhere — `agent_activity.ts:2221` and `:3612` use `.catch(...)`
on a task result — so this reads as an oversight rather than a decision.

All current sites: `agent_activity.ts:1616/2822/3983`, `worker.ts:720/745/963`, `proc_pool.ts:168`,
`job.ts:574`, `tts.ts:655/764`, `avatar/datastream_io.ts:205`, `speech_handle.ts:168`. None has a `catch`.
`speech_handle.ts:168` cannot reject from its own future, but a user `done` callback that throws rejects it.

## A rule that catches the whole class

`@typescript-eslint/no-floating-promises` covers exactly this. The repo already lints with type information
(`parserOptions.project` in `.eslintrc.cjs`) and enables `no-misused-promises`, so turning it on needs no new
setup, and `pnpm lint` already runs in CI.

Enabling it today reports **163 findings across 657 files** (`agents/src` 111, tests 49, plugins 3, examples 0),
so it can't be flipped on in one step. A rough split of the non-test findings:

| kind | count | note |
|---|---|---|
[...]

PR #2553, opened 2026-09-21
chore: enable @typescript-eslint/no-floating-promises

JustinScitech:
## Description

Part of #2508. Stacked on #2552; retarget to `main` once that merges.

Enables `@typescript-eslint/no-floating-promises` as an `error` (with `ignoreIIFE: true`) in `.eslintrc.cjs`. The rule catches the class of bug fixed in #2552: a promise created and dropped. The repo already lints with type information, so it needs no setup, and `pnpm lint` runs in CI.

With #2552 applied, the rule reports 254 sites across the repo. This PR makes each one explicit so the rule can be on from day one and every new occurrence fails lint.

| Package | Sites |
|---|---|
| `agents` | 142 |
| `examples` | 62 |
| `plugins` | 50 |

## Changes Made

- **`void` at every existing site.** Behaviour is unchanged; the drop is now deliberate and reviewable. The bulk are `SpeechHandle` returns from `say()` / `generateReply()` / `interrupt()` (the handle is a thenable), fire-and-forget stream `write()` / `close()` calls, and `async` methods that were already called without `await`. The installed typescript-eslint (6.21) predates `allowForKnownSafePromises`, so `SpeechHandle` cannot be whitelisted by type; that would remove most of the `void`s if the toolchain is upgraded later.
- **Two test files use `await` instead**, because the dropped promise was a test bug:
  - `examples/src/testing/agent_task.test.ts`: three `judge()` LLM assertions were never awaited, so their failures were discarded.
  - `plugins/rime/src/websocket.test.ts`: the connection close at the end of a test was not awaited.

### Sites worth a look in a follow-up

Marked `void` here to keep this PR behaviour-neutral, but each looks like it may want a real handler:

- `agents/src/job.ts` `accept()`: drops `#onAccept(...)`; Python awaits it.
- `plugins/livekit/src/turn_detector/base.ts`: `downloadFileToCacheDir(...).then(...)` has no `.catch`, so a download failure is still an unhandled rejection and `languagesFuture` never settles.
- `agents/src/worker.ts` and `ipc/proc_pool.ts`: `void this.#tasks.splice(...)` is flagged only because the array holds promises; a `Set` would read better.

## Pre-Review Checklist
- [x] **Build passes**: `pnpm lint` reports zero `no-floating-promises` findings. The single remaining lint error (`no-misused-promises` in `plugins/openai/src/ws/llm.ts:127`) exists on `main` and is untouched here. `tsc` is clean for `agents` and `examples`.
- [x] **Changes explained**: above
- [x] **Scope appropriate**: rule enablement plus the mechanical acknowledgement it requires; no behaviour changes
- [ ] **Video demo**: not applicable

## Testing
- [x] All tests pass: full `agents` suite, 2728 passed, 5 skipped. The one failure in the parallel run (`agent_task_close.test.ts`) passes in isolation and fails intermittently on `main` too.


JustinScitech:
Closing: enabling the rule requires touching 254 sites across 91 files for no runtime change. Leaving the rule as a follow-up in #2508; #2552 covers the .finally fixes.

JustinScitech:
Reopening per Jason: add the rule. Rebuilt on the rebased #2552 base; same 254 sites, all checks rerun.

JustinScitech:
Reopening per Jason: add the rule. Rebuilt on the rebased #2552 base; same 254 sites, all checks rerun.

